PR #13, opened 2016-09-21
Added "European Tyre and Rim Technical Organisation" and "Tire and Ri…

magnusfeuer:
Clarified reason behind mixed unit type use in wheel attributes.

### DIFF
--- a/spec/Attribute/Chassis.vspec
+++ b/spec/Attribute/Chassis.vspec
@@ -8,19 +8,19 @@
 
 - FrontWheelDiameter:
   type: Uint8
-  description: Diameter of wheels (without tires), in inches.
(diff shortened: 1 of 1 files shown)


rstreif:
As part of the rather extensive additions of signals and attributes originating from W3C discussions I reorganized spec/Attribute/Chassis.vspec and added the branches Axle, Axle.Row1, Axle.Row2. This provides more flexibility for axle, wheel and tire configuration. This scheme is also replicated for the signals where brake signals etc. are added.

This pull request conflicts. However, I have added the changes to the reworked chassis.vspec file.

rstreif:
These changes have already been merged manually as indicated earlier.

PR #23, opened 2017-01-11
Unified all files and artifacts under the same license

rstreif:
Previously the vspec files were licensed CC-BY while code and
generated files were licensed MPLv2. Now all files and artifacts
are licensed under MPLv2.

Signed-off-by: Rudolf J Streif <rstreif@jaguarlandrover.com>


gunnarx:
Great!  Thanks.

PR #31, opened 2017-03-07
Minor changes to description and generated files.

as2902b:
1. Description change.
2. Change in ID and description of files generated from running 'make'.


rstreif:
@as2902b Hi Sanjeev,
You rightfully corrected a type in a signal name. In this case you need to delete the spec/VehicleSignalSpecification.id before running make again otherwise it contains the signal with its correct and incorrect writing. As VSS is stil work in progress this is ok. Once it's released we will probably have to live with incorrect names so that subsquent builds do not renumber existing signals. Thanks for doing the due diligence. I very much appreciate it.

as2902b:
Rudi, I did the following steps. Please verify and let me know. I am still getting used to github workflow.
Revert id file to previous version.
Add regenerated id file after running make again.
Add remaining generated files.
Added a clean target to Makefile (run it by default)
Recommend to run make after changes to spec/ folder in README.md

rstreif:
Sanjeev,
Good idea adding a clean target to the makefile. The only concern is, that the id file is now deleted every time when the other files are created. That is ok as long as we are in development state. However, once we make the first release we do not want to recreate the entire id file but just add to it. If we recreate it then we break backwards compatibility. The generator actually read the id file first before adding new id's and pick up from the highest number.

My suggestion: remove deleting the id file from the clean target but add it to a clean_all target:
clean:
 	rm vss_rel_$$(cat VERSION).json vss_rel_$$(cat VERSION).fidl \
        vss_rel_$$(cat VERSION).vsi vss_rel_$$(cat VERSION).csv

clean_all: clean
       rm spec/VehicleSignalSpecification.id 


Sorry for the fuzz...

as2902b:
Thanks for taking the time. I ll resubmit with the changes.

PR #32, opened 2017-03-08
Resubmitting previous changes.

as2902b:
Have redone the commits.
1. Makefile with clean target that cleans only generated files and DOES NOT touch id file.
2. AllWindow spec file change.
3. ID File deletion (with the exception) that we are now in development mode and have the flexibility to renumber the params. This kind of commit wont be allowed, post VSS release.


rstreif:
Looks good. Thank you, @as2902b

Issue #40, opened 2017-08-14
Spelling error: DriverPostion


erikbosch:
I assume it should be called DriverPosition. 

Furthermore, I am not 100% sure that I understand this variable correctly? Should it be interpreted as seat number from left?

https://github.com/GENIVI/vehicle_signal_specification/blob/222bbc6d1900e2c82dc15138da2266dcdae3c53e/vss_rel_1.0.fidl#L306

gunnarx:
Please provide PRs if possible.  Anyway, easy fix so I pushed this and a few things in a PR (see ref above)

erikbosch:
Thanks for creating the PR Gunnar

rstreif:
Fixed with #55.

PR #53, opened 2018-02-19
Spelling and other fixes (misc)

gunnarx:
Spelling fixes as reported in #40 and other.

Also another unrelated README fix, removing the comment and linking to SI units (I still kept the table for now, it's probably convenient to have).  

Let me know if I should split something out of the PR.   I might update with a few more commits, unless it's merged by then.


gunnarx:
branch was on wrong remote - filing a new PR

Issue #54, opened 2018-02-19
Branch policy?


gunnarx:
@rstreif What's the branch policy?

**develop** is way ahead of **master**, which has not been updated since 2016.  
Not sure what **master** is for now.  I'd suggest simply merging **develop** into **master** (it's a fast forward) and then delete **develop** and keep **master**.  OK?

gunnarx:
Bump.  Maybe we can discuss it today?

rstreif:
Merge `develop` into `master`.

gunnarx:
As agreed, I will do this change.

gunnarx:
* merged develop into master (fast forward) and pushed.  
* Changed settings in GitHub for the repo such that master is now the default branch
* develop has been deleted

PR #62, opened 2018-03-26
Corrected formating error in README.md

klotzbenjamin:


gunnarx:
I fixed your commit and pushed it back so it does not blow away previously merged changes.

gunnarx:
Trivial change, and also mentioned in meeting.  Since the PR is aimed at develop and develop will disappear as soon as I execute #54 (as agreed), I thought it is best to merge this first, to avoid any issues.

PR #63, opened 2018-03-26
clarification of fwd-rwd-awd

klotzbenjamin:


gunnarx:
Did we have time to discuss this?  I might have missed.  Anyway, it's a policy decision here whether to allow spaces in the enums.  If we want to be more programming friendly then spaces should not be there.  If we want to be more abstract then it's in theory OK for a YAML/JSON format.

PR #67, opened 2018-04-19
Add HMI Related Signals

acrofts84:
Including current HMI language and system units for HMI


spec/Signal/Cabin/Infotainment.vspec
```diff
@@ -117,3 +117,48 @@
...
+  description: HMI related signals
+
+- HMI.CurrentLanguage:
+  type: string
+  description: ISO 639-1 standard language code for the current HMI
+  sensor: HMI
+
+- HMI.DateFormat:
+  type: string
+  enum: [ "DD MM YYYY", "MM DD YYYY" ]
```
gunnarx: I find this inadequate.  It doesn't even include YYYY-MM-DD which is in fact an [ISO standard format](https://en.wikipedia.org/wiki/ISO_8601) and  preferred in various parts of Europe at least.  Yes, it is explicitly using hyphens.  

The proposal here with spaces does not seem not clarify the actual display format, which would depend on the characters used in between digits.   For example hyphens, or slashes:  DD/MM/YYYY.  Also it could be YYYY or only YY, and so on?

I'm not sure if it makes sense to include this in a standard signal database - it could be seen as an example, but if the intention is to make it usable for all cars then I am sure there are many more alternatives than those given.
Even if it is an example it should be as accurate as possible of course.

spec/Signal/Cabin/Infotainment.vspec
```diff
@@ -117,3 +117,48 @@
...
+- HMI.TimeFormat:
+  type: string
+  enum: [ "12HR", "24HR" ]
+  description: Time format used in the current HMI
+  sensor: HMI
+  actuator: HMI
+
+- HMI.DistanceUnit:
+  type: string
+  enum: [ "Miles", "Kilometers" ]
```
gunnarx: Or use standard abbreviations instead?  (The README even speaks about standard units, so that can be reused)

acrofts84:
Thanks @gunnarx for the feedback. The main motivation behind adding these signals is to ensure the UI is consistent across multiple views throughout the vehicle. Some of these units are more likely to be changed in the vehicle cluster, than in the main HMI, therefore exposing these through the standard would be beneficial.

Regarding date format, YYYY MM DD clearly needed to be added, as well as YY for each format. I personally don't see the delimiters as too important - if an implementation wishes to implement a selection of characters for this it could be added as a separate field, but ultimately ensuring an application UI renders the fields in the correct order is paramount. 

I have also added FuelEconomyUnits and EVEconomyUnits.

rstreif:
@acrofts84 , @gunnarx,

I am ok with the proposed changes after the modification. @G-Sriram?

PR #72, opened 2018-06-28
Vssnative

UlfBj:
Support for rbanch and element node types.


UlfBj:
To build, follow instructions in c_native/README.md

UlfBj:
Closed to be replaced by version removing need for hardcoded path to so file.

UlfBj:
As this is ok, just missing one feature, -i would like to have this reopened.

PR #80, opened 2018-11-27
Use single spec file for windows

### LINKED ISSUE #60: AllWindows.vspec not used
The spec file [Signal/Cabin/AllWindows.vspec](https://github.com/GENIVI/vehicle_signal_specification/blob/develop/spec/Signal/Cabin/AllWindows.vspec) is never included.

As a result, the only window signals available are the ones defined outside this file.

danielwilms:
* rename allwindows to singlewindow
* Take in use single spec file
* Remove curtain, as it's called shade
* Status of windows as in door "IsOpen" "IsClosed"

Fixes #60

### DIFF
--- a/spec/Signal/Cabin/SingleDoor.vspec
+++ b/spec/Signal/Cabin/SingleDoor.vspec
@@ -28,20 +28,7 @@
   type: branch
   description: Door window status
 
-- Window.Position:
-  type: Uint8
-  min: 0
-  max: 100
-  unit: percent
-  description: Window position. 0 = Fully closed 100 = Fully opened.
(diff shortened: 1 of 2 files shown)


gunnarx:
- Reason for removing position between 0 to 100% open?
- Can a window be "locked"?

danielwilms:
@gunnarx:
1. It is used in the window vspec file and not in the door itself anymore. the file was there already, but never included.
1. Good catch. That was an earlier copy & paste error in the original file, I haven't noticed. will fix that one.

gunnarx:
> It is used in the window vspec file

Yes you're right.  I found that now.

PR #92, opened 2018-12-11
Added Recorder branch/Removed signals, attributes from root

### LINKED ISSUE #84: Top nodes in VehicleSignalSpecification.vspec
Under the root node Vehicle there is now, besides the Media node and the Private node, also also all the nodes like DriveTrain, Cabin, Body, ADAS, Chassis, and OBD. 
I think these should be pushed one level down, under a branch node that may have the name Car (from VIWI), or something else. The nodes mentioned above belong to the same logical domain, and should reside under a domain name node for these.  It will be more comparable to the Media domain node, and further new domain nodes that might be added later.

danielwilms:
As the vehicle node got the new root of the tree in
one of the previous commits, the root node now had
signals and attributes attached to it directly.

This has been seen as an issue, check discussion #84 
Resolved by:
* adding Recorder node
* moving attributes to vehicleidentification branch
* moving signals to recorder branch

Fixes #84


UlfBj:
I believe the following lines should be removed from the VehicleSignalSpecification.id file.
Vehicle.RoofLoad: 1212
Vehicle.Speed: 1200
Vehicle.TravelledDistance: 1201
Vehicle.TripMeterReading: 1202

Given the changes of this PR I think that the file Vehicle.vspec should be renamed to VehicleIdentification.vspec.

magnusfeuer:
@danielwilms  - Can you rebase the PR so that we don't have conflicts?

gunnarx:
@magnusfeuer @danielwilms :

I rebased it according to request, fixing the conflicts in the process.  (I assumed the delay was because of difficulty with the conflicts.  Let me know if this was a bad decision)

A few comments on that.  Here follows a --stat of the **original** changes requested in the relevant commit (eac5de53bca1b85dc05773aafe1ca848cbe738b6) that was on the `issues/#84` branch (note this branch is now changed since I had to force push it to update the PR).

```
 VERSION                                            |     2 +-
 spec/Recorder/Recorder.vspec                       |   130 +
 spec/Vehicle/Vehicle.vspec                         |   126 +-
 spec/VehicleSignalSpecification.id                 |    20 +
 spec/VehicleSignalSpecification.vspec              |     9 +
 ....0-alpha+004.csv => vss_rel_2.0.0-alpha+005.csv |  2769 +--
 vss_rel_2.0.0-alpha+005.fidl                       | 10000 +++++++++
 ...-alpha+004.json => vss_rel_2.0.0-alpha+005.json | 20246 ++++++++++---------
 ....0-alpha+004.vsi => vss_rel_2.0.0-alpha+005.vsi |   466 +-
```
The VERSION was updated to 006 in master already, but the PR had version 005 (changed from 004).  Therefore this was a conflict.  I now set it to 006 but leave it to the maintainer to figure out how the versioning _should_ be updated here.  Maybe this should rather be version 007 then, because it is different from master at 006.

The .id file was in conflict because it has been updated in both places.  I am checking my result now and it seems that I took the IDs from the master version, not from the PR which I suppose is wrong.  Does this matter?  (See comment about checking in generated files later)

The .csv, .json and .vsi files were also conflicting because they were previously checked in but have been deleted from master in other commits, and here they were renamed.  So there's a conflict about if they should exist or not.

In this case **I added them** as the PR requested but I think that's wrong.   I assume we follow what master had already done and avoid  _generated_ files to be included in the repository,, or what is our opinion here?

UlfBj:
Why has this not been merged?
I think it should either be merged, or rejected.

PR #94, opened 2019-01-15
intptr_t recasting and python3 updates

UlfBj:


danielwilms:
Just saw that there are merge conflicts. will do a rebase and let you know, so that you can check, that I didn't break something ;) @UlfBj

UlfBj:
I guess the merge conflict is in vssparserutilities.h, in the nodeTypes_t declaration, for the enum value STREAM, that I earlier had another name on. If that is the case, then STREAM should be the correct enum value.

Issue #95, opened 2019-02-26
Need to assign IDs to branches, not only signals


magnusfeuer:
We are at a situation where we need to atomically transmit an entire signal sub-tree with all signals hosted under a specific branch. 
While individual signals can be identified by their signal ID, no IDs are assigned to branches. This means that we have to transmit full path names over the network, which is inefficient.
The proposed solution is that we assign signal IDs to branches as well as signals.
I will branch off master, modify the python tooling to do just this, and then submit a merge request.

PR #96, opened 2019-02-26
Added signal ID to branches.

magnusfeuer:
Bumped version to 2.0.0 Alpha+005.
Created new JSON, CSV, vsi, fidl, and VehicleSignalSpecification.id artifacts (+005 suffix)

Implements issue #95


tools/vspec.py
```diff
@@ -1,6 +1,6 @@
 #
 # (C) 2018 Volvo Cars
-# (C) 2016 Jaguar Land Rover
+# (C) 2016 Jaguar Land Rover (makes sexier cars than Volvo)
 #
```
danielwilms: haha...wanna check the review process :)) maybe we could remove this one before we merge.

klotzbenjamin:
Another remark @magnusfeuer 
It was already an issue some time ago, but the IDs are a bit surprising to me. It starts at 1178, and skips 1193, 1194 and everything between 1216 and 3421.

They are at least consistent between generated files (as it was the case before).

tools/vspec.py
magnusfeuer: Dang - you caught it.
Fixed.

magnusfeuer:
Commenting on @klotzbenjamin 
 VehicleSignalSpecification.id is the master map between signals and their ID. Can you point me to an example where there is a mismatch? That may be a bug.
The ID assignment is a bit stochastic since it depends on the traversal order of the signal tree loaded by vspec.py. Please do tell if there is a better way of assigning IDs, and I'll see if I can make it happen.

magnusfeuer:
Are we ok if I merge this PR into master?

PR #118, opened 2019-08-13
Introduce GitHub Pages Documentation

danielwilms:
Instead of the README file, I used HUGO as a static web page generator to structure the docs in a better way. How the result looks like, can be seen [here](https://danielwilms.github.io/vehicle_signal_specification/introduction/). As soon as the PR gets accepted, I'll enable GitHub Pages for this repository.


danielwilms:
I have to push to another branch, as there would be a conflict with the private github pages otherwise. I will create a new PR.

PR #119, opened 2019-08-13
Introduce GitHub Pages Documentation

### LINKED ISSUE #109: README refactoring
I think the VSS root directory README needs a refactoring to more clearly explain what this repo contains, which in my view is three separate, but related parts: 
- A data model specification,
- An instance of this data model ,
- Tools to transform theYAML based instance into other formats.
The README should have a layout presenting each of these parts.
Further, I think the specification part should reside in a separate document (mirrored in the README) to that it can easily be handled separately.

danielwilms:
Instead of the README file, I used HUGO as a static web page generator to structure the docs in a better way. How the result looks like, can be seen [here](https://danielwilms.github.io/vehicle_signal_specification/introduction/). As soon as the PR gets accepted, I'll enable GitHub Pages for this repository.

Fixes #109


tguild:
Perfectly fine with web pages instead of a README markdown although have no strong preference. Perhaps too many pages and a flatter view would be better keeping the current page fragments as sections on larger pages. Especially like how tooling is separated and should remain so.

gunnarx:
I like it, Daniel.  It is something to build on.

klotzbenjamin:
Thank you Daniel, this looks good !

I have a few superficial comments, mostly related to what we already have in the README:
- In the [quickstart](https://danielwilms.github.io/vehicle_signal_specification/introduction/quickstart/), the json viewer needs a file that is no longer in the repository and linked to a specific version, this should be more robust.
- In the [data entry](https://danielwilms.github.io/vehicle_signal_specification/rule_set/data_entry/), "attributes" is the only term listed in plural.
- In the [unit section](https://danielwilms.github.io/vehicle_signal_specification/rule_set/data_entry/data_unit_types/), you wrote that we use SI units, but it is not fully true. We additionally have domain-specific standards (l/100km for instance).
- In the ["sensor and actuator" section](https://danielwilms.github.io/vehicle_signal_specification/rule_set/data_entry/sensor_actuator/), the introduction implies that this description only applies to sensors, but it is for both sensors and actuators.
- In the [rbranch section](https://danielwilms.github.io/vehicle_signal_specification/rule_set/resource_branch/), I still think we lack a bit of motivation for using this node type. Likewise, we do not clearly define the "element" node that it contains in practice.
- In the [declaring vs defining attributes paragraph](https://danielwilms.github.io/vehicle_signal_specification/rule_set/private_branch/#declaring-vs-defining-attributes) could be already mentionned in the attribute section.
- The version visible in the bottom-left part of the site is correct ?

UlfBj:
>> In the rbranch section, I still think we lack a bit of motivation for using this node type. 
The reason this was introduced was to make possible to support the "resource model" used in VIWI. If there will not be any "porting" from VIWI, then I agree the motivation for this node type is lacking. But to not close that door prematurely, I think we can keep it for now.
 
>>Likewise, we do not clearly define the "element" node that it contains in practice.
Correct. For completeness, I think we shall add the "element" node type, if we decide to keep the rbranch type.

danielwilms:
Hi @UlfBj 
> Correct. For completeness, I think we shall add the "element" node type, if we decide to keep the rbranch type.

Now merging the PR. Do you want to have a look at the rbranch docs?

PR #125, opened 2019-09-30
Using reproducible UUID namespace

rstreif:
This commit updates the creation of UUID5 (SHA-1 hash) from using
the UUID5 of their parent as the namespace to a UUID5 namespace
derived from the object identifier 'vehicle_signal_specification':

> uuid.uuid5(uuid.NAMESPACE_OID, 'vehicle_signal_specification')
> UUID('4635ee41-2db8-5e71-94e0-9ad0f157522c')

This means that a UUID5 for any node in the VSS tree can always
be computed with:

> uuid.uuid5(UUID('4635ee41-2db8-5e71-94e0-9ad0f157522c'), 'Vehicle')
> UUID('ccc825f9-4139-544d-bb5f-4bfd033bece6')

This fixes the criticism that the previous signal ID, which was simply
a numbering, already had and that was carried over to the UUID scheme
that with any new creating of the ID database the IDs would change
unless the database was used as a lookup table (in that case the ID
would simply be read from the database).

This carried over to the UUID-based database because the root UUID
(the one for the first node called 'vehicle') was created using a
UUID1 which is based on hostid, sequence number and time. Furthermore,
the previous UUID creation used the parent node's UUID as the namespace,
hence nodes with different parents would use a different namespace (and
all nodes with 'vehicle' as parent would use the UUID1 for 'vehicle' as
their namespace).

Technically, with this change the ID database becomes somewhat obsolete
as a lookup table to create UUIDs as they always can be reccomputed as
shown above. However, the database still has its place as a reverse lookup
table since UUIDs are of course hashes and hash functions are not inverse.


gunnarx:
Looks good to me and I think it takes a step in the right direction, if IDs are to be used.

I have rather promoted the opinion that the fully-qualified-path ***is*** the unique ID and nothing else is needed as part of VSS itself.  Anything beyond it such as achieving quicker lookup by some integer identifier could be implementation specific.  But I'm not violently opposed.  Having a specification of a shared, stable ID does not hurt, and this is better than before.  It might also be efficient if handled as an integer inside implementations.  (As a string it's on the same scale as the FQN in terms of size)

Maybe in the future, a function to calculate the UUID for any signal, that can be called by any python implementation (hiding the implementation).  I'd like it better than the get_namespace_uuid function.

+1 to merge

danielwilms:
> I have rather promoted the opinion that the fully-qualified-path is the unique ID and nothing else is needed as part of VSS itself.

Agreed, but we have a use-case on-board, where a fixed length identifier is needed. I really like this approach.

danielwilms:
merged with approval of @gunnarx, @UlfBj

PR #128, opened 2019-10-30
Specify safe YAML loader, which avoids warning

gunnarx:
This change deals with the warning that was printed before.
(read link for more info):

   YAMLLoadWarning: calling yaml.load() without Loader=... is
   deprecated, as the default Loader is unsafe. Please read
   https://msg.pyyaml.org/load for full details.

Signed-off-by: Gunnar Andersson <gandersson@genivi.org>


gunnarx:
By the way, I checked that the output files (vss_rel_*) were generated identically after running 
make, before and after this change.

UlfBj:
+1

PR #130, opened 2019-11-15
POSITIONING: Added extensions to the vspec files

danielwilms:
- introduced extensions
- removed duplication from the vspec file
- tooling separated for discussion, on how it should look like

Part of: #81

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -83,33 +83,18 @@
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-- ObstacleDetection.DistanceToObject:
+- ObstacleDetection.Input:
+  extension:
+    - ["Front","Rear"]
+    - ["Left", "Right"]
   type: branch
-  description: Distance signals form Obstacle Sensor System
+  description: Direct sensor input on obstacle detection
 
-- ObstacleDetection.DistanceToObject.FrontLeft:
+- ObstacleDetection.Input.DistanceToObject:
   datatype: uint16
   type: sensor
   unit: m
-  description: Front left distance to object in meters
-
-- ObstacleDetection.DistanceToObject.FrontRight:
-  datatype: uint16
-  type: sensor
-  unit: m
-  description: Front right distance to object in meters
-
-- ObstacleDetection.DistanceToObject.RearLeft:
-  datatype: uint16
-  type: sensor
-  unit: m
-  description: Rear left distance to object in meters
-
-- ObstacleDetection.DistanceToObject.RearRight:
-  datatype: uint16
-  type: sensor
-  unit: m
-  description: Rear right distance to object in meters
+  description: Distance to object in meters
 
 
 #
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -83,14 +83,10 @@
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-- ObstacleDetection.Input:
-  extension:
+- ObstacleDetection.DistanceToObject:
+  instances:
     - ["Front","Rear"]
     - ["Left", "Right"]
-  type: branch
-  description: Direct sensor input on obstacle detection
-
-- ObstacleDetection.Input.DistanceToObject:
   datatype: uint16
   type: sensor
   unit: m
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -83,85 +83,42 @@
 #
 - Windshield:
   type: branch
+  extension: ["Front", "Rear"]
   description: Windshield signals
 
 #
 # Front windshield
 #
-- Windshield.Front:
-  type: branch
-  description: Front windshield signals
 
-- Windshield.Front.Wiping:
+- Windshield.Wiping:
   type: branch
   description: Front windshield wiper signals
 
-- Windshield.Front.Wiping.Status:
+- Windshield.Wiping.Status:
   datatype: string
   type: actuator
   enum: ["off", "slow", "medium", "fast", "interval", "rainsensor"]
   description: Front wiper status
 
-- Windshield.Front.Heating:
+- Windshield.Heating:
   type: branch
   description: Front windshield heater signals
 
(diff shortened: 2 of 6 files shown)


spec/ADAS/ADAS.vspec
```diff
@@ -83,33 +83,18 @@
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-- ObstacleDetection.DistanceToObject:
+- ObstacleDetection.Input:
```
gunnarx: I guess it's no surprise that I question this one, first of all.  In #81 we discussed if an extra node is required in the YAML tree to hold the instance declarations or if those could be put directly under DistanceToObject.  I preferred not creating the extra node.  

However, I also thought we were converging in #81 towards that **if** an extra node is required in the YAML tree to hold the instance declarations , but **if** it ithen we seemed to talk about naming the node itself "instance" instead of an extra word like "Input"  which is only created because some name is needed (but later on, the idea of using a wildcard character like _ instead seemed to be considered also by @UlfBj ).

Furthermore, you just suddenly use the word "extension" instead of "instance" in the actual instance itself?  So are you then saying that you prefer this word over instance and is the intention that it shall mean the same thing?  I feel like you could explain/discuss that first.   (Is it something you have discussed in another forum which I missed?)

danielwilms: @gunnarx: At the beginning when we were talking about the structure and we had "position" as a name, when I think @UlfBj requested to change it to extension. From back then I used the code, which I've created to experiment. So it slipped through. If it's requested I can call it instance.

> However, I also thought we were converging in #81 towards that if an extra node is required in the YAML tree to hold the instance declarations then we seemed to talk about naming the node itself "A.B.C.instance" instead of an extra proper word like "Input"

I guess I got lost in the discussions. I thought the request was to have it on branches only. It's doable :)

danielwilms:
I applied the changes you requested, please check, if you can live with those and if I've overseen something.

UlfBj:
It looks OK to me.

danielwilms:
> It looks OK to me.

@gunnarx?

danielwilms:
@gunnarx any objections?

danielwilms:
> Adding a "request changes" comment, just to indicate that I would like to vote against this PR.

@gunnarx, fixed now?

gunnarx:
In #81 one of the later comments from @UlfBj was:

>    I think wildcard is actually a better choice than "instances" as placeholder node name.

I had just before that proposed :  _ or * instead of the word "instances" *in the actual node name*.  This is the kind of wildcard that could be acceptable to me and still use Ulf's semantics where this wildcard name would in effect "disappear" in the expansion of the instance expression.

Now I see that for the *node name* in the YAML you are not adding ".instances",  so that is good, but I don't know if the PR can/should be accepted before we have agreed on what the change proposal should mean (i.e. #81).  

The way we will address the created instances after merging this PR, is to me equal to the "semantics" of the VSS.  We must agree on the semantics, right?

For example, the PR now includes:

```
- Axle.Wheel:
  instances: ["Left","Right"]
```

This seems OK with me, **assuming that** we are using the original semantics I proposed, where the word Wheel **remains** in the path, and the effect of the above is basically to generate these two nodes:
```
<the-path-leading-up-to-it>.Axle.Wheel.Left
<the-path-leading-up-to-it>.Axle.Wheel.Right
```

However, if it is supposed to be translated as:
```
<the-path-leading-up-to-it>.Axle.Left
<the-path-leading-up-to-it>.Axle.Right
```

which was @UlfBj 's original proposal then this PR seems misleading to me.   (Because surely you are not intending to instantiate a Left and Right Axle.

I don't think creating a PR and asking for it to be accepted really helps the process here.   It does not take away the need for completing #81.   As I think I have said, we **must**  get through the discussion and try to find consensus on the chosen solution.

@UlfBj how do you see this, and can you confirm if you are then accepting "my" semantics when you 
approved the content of this PR?

@danielwilms My problem with our interaction is that you seem to miss or just not mention some of the details.   I think it would be better if we don't hide away from the discussions.  Could you take a stance on the open points in #81?  If it is too much, maybe we have to again summarize where we stand?
Then say which you prefer.  Such as : "I disagree with person X in comment Y, and think we should stick to the word so and so.

To be clear, in this PR you did not give opinion on the proposal of ".instances" or "._" or ".*" but did something completely different which is to not include any particular node name at all.  When you do that I wish you would explicitly say if this is in support of "my" semantics, or if you wish something else, or if you did not think about it and just propose something to drive the decision process between me and @UlfBj .  Thanks.

danielwilms:
@gunnarx, don't agree completely. IMHO, the discussion is about two things:
1. simplifying the yaml, and
1. simplifying the resulting tree

I see as well that we have open issues on the latter, but I don't want to implement the tools, when we don't have an agreement on the yaml structure as such. I think we can only get to a conclusion, when we see the result in json or csv or whatever format helps to browse through the resulting tree. IMHO the simplification of the specification is good to go. The rest we can take from there. That would help me to work on the example. I don't even have to merge it. I just would like to know, if I resolved the issues you had with the yaml specification, so that we can jump into resolving the rest.

danielwilms:
@gunnarx: Sorry, I've simply overseen the last two comments on #81, which were about the yaml. I tried to put in all we discussed. Will add my comment there.

UlfBj:
>> @UlfBj <https://github.com/UlfBj> how do you see this, and can you
confirm if you are then accepting "my" semantics when you
approved the content of this PR?

My approval is in line with your semantics view.

Den mån 6 jan. 2020 kl 05:58 skrev gunnarx <notifications@github.com>:

gunnarx:
Sorry for reply via email. Hopefully it works. Can't log into Github at the
moment:

@danielwilms @UlfBj Thanks for your replies.  It seems we can go ahead and
try this, and implement an expansion into csv to confirm our agreement.

Also, due to travel I will be less responsive. Thanks for pushing forward.

magnusfeuer:
Long story short: I am for the changes, assuming that we go with Gunnar's semantics.
Any thoughts on who has time / wants to modify the tooling?

gunnarx:
As far as I can see this PR is obsolete since #131 includes all the same commits.
@danielwilms If you agree, please close.

danielwilms:
@gunnarx, closed in favour of #131.

PR #131, opened 2019-11-15
POSITIONING: Added tooling for CSV support

danielwilms:
First attempt on how the tooling could look like
with the positions as instances. Dot notation might
not be the best form. But as discussion item for now
to conclude #81.

### DIFF
--- a/tools/vspec2csv.py
+++ b/tools/vspec2csv.py
@@ -83,73 +83,73 @@ def format_data(json_data):
     return f"{Uuid},{Type},{DataType},{Unit},{Min},{Max},{Desc},{Enum},{Sensor},{Actuator}"
 
 
-def json2csv(json_data, file_out, parent_signal, ext=[]):
+def json2csv(json_data, file_out, parent_signal, instances=[]):
     for k in json_data.keys():
         if (len(parent_signal) > 0):
             signal = parent_signal + "." + k
         else:
             signal = k
 
-        local_ext = ext
-        if ('extension' in json_data[k].keys()):
-            local_ext = ext + [(json_data[k]['extension'])]
+        local_instances = instances
+        if ('instances' in json_data[k].keys()):
+            local_instances = instances + [(json_data[k]['instances'])]
 
         # if it's a branch, create an entry and continue
         # with its children
         if (json_data[k]['type'] == 'branch'):
             file_out.write(signal + ',' + format_data(json_data[k]) + '\n')
             # if postion attribute exists, keep it
-            json2csv(json_data[k]['children'], file_out, signal, local_ext)
+            json2csv(json_data[k]['children'], file_out, signal, local_instances)
 
-        # if it's a leave, make an entry, check the extension entries and
-        # create an entry for every extension
+        # if it's a leave, make an entry, check the entries for instances and
+        # create an entry for every instance
         else:
-            file_out.write(signal + ',' + format_data(json_data[k]) + "," + str(ext) + '\n')
-            createExtensionEntries(local_ext, file_out, json_data[k], signal)
+            file_out.write(signal + ',' + format_data(json_data[k]) + "," + str(instances) + '\n')
+            createInstantiationEntries(local_instances, file_out, json_data[k], signal)
 
 
 
-def createExtensionEntries(ext, file_out, json_data, prefix=''):
-    """Function creates the extension entries to a given node as
+def createInstantiationEntries(instances, file_out, json_data, prefix=''):
+    """Function creates the entries for instances to a given node as
        postfix of the node name (e.g. node.Row1.LEFT)
 
     Parameters
     ----------
-    ext : list
-        list of branches with extension argument in the path
+    instances : list
+        list of branches with instance argument in the path
     file_out : file
         csv file to write into
     json_data : json object
         json encoded information about the node
     prefix : string
-        node name before the extension information
+        node name before the instance information
     """
 
     REG_EX = "\w+\[\d+,(\d+)\]"
 
-    if not ext and file_out and json_data:
+    if not instances and file_out and json_data:
         return
     rest = None
     i = []
-    if len(ext) == 1:
-        i = ext[0]
+    if len(instances) == 1:
+        i = instances[0]
     else:
-        i = ext[0]
-        rest = ext[1:]
+        i = instances[0]
+        rest = instances[1:]
 
 
     if prefix and not prefix.endswith("."):
         prefix += "."
 
 
-    # parse string extension elements (e.g. Row[1,5])
+    # parse string instantiation elements (e.g. Row[1,5])
     if isinstance(i,str):
         if re.match(REG_EX, i):
-            extRangeArr = re.split("\[+|,+|\]",i)
-            for r in range(int(extRangeArr[1]),int(extRangeArr[2])+1):
-                nextPrefix = prefix + extRangeArr[0]+str(r)
+            instRangeArr = re.split("\[+|,+|\]",i)
+            for r in range(int(instRangeArr[1]),int(instRangeArr[2])+1):
+                nextPrefix = prefix + instRangeArr[0]+str(r)
                 if rest:
-                    createExtensionEntries(rest, file_out, json_data, nextPrefix)
+                    createInstantiationEntries(rest, file_out, json_data, nextPrefix)
                 else:
                     json_data["type"] = "instance"
                     file_out.write(nextPrefix + ',' + format_data(json_data) + ",***" + '\n')
@@ -157,14 +157,14 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
 
         # TODO: right now dynamic extensions not supported
         else:
-            raise vspec.VSpecError("","","Extension type not supported")
-    # Use list elements for extension (e.g. ["LEFT","RIGHT"])
+            raise vspec.VSpecError("","","instantiation type not supported")
+    # Use list elements for instances (e.g. ["LEFT","RIGHT"])
     elif (isinstance(i,list)):
         complexList = False
         for r in i:
-            # if in case of multiple extensions in one branch
+            # if in case of multiple instances in one branch
             # it has to be distinguished from a list of
-            # string extensions, like ["LEFT","RIGHT"]
+            # string instantiations, like ["LEFT","RIGHT"]
             if (isinstance(r,str)):
                 if re.match(REG_EX, r):
                     if (rest):
@@ -175,7 +175,7 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
                 else:
                     nextPrefix = prefix + str(r)
                     if rest:
-                        createExtensionEntries(rest, file_out, json_data, nextPrefix)
+                        createInstantiationEntries(rest, file_out, json_data, nextPrefix)
                     else:
                         # create a new type instance for now for
                         # better recognition, of what it is
@@ -183,7 +183,7 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
                         file_out.write(nextPrefix + ',' + format_data(json_data) + "," + '\n')
 
             else:
-                # in case of multiple extensions, the list is
+                # in case of multiple instances, the list is
                 # has to be parsed, like ["LEFT","RIGHT"]
                 if (rest):
                     rest.append(r)
@@ -192,7 +192,7 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
                 complexList = True
 
         if complexList:
-            createExtensionEntries(rest, file_out, json_data, prefix)
+            createInstantiationEntries(rest, file_out, json_data, prefix)
 
     else:
         print (i)
--- a/tools/vspec2csv.py
+++ b/tools/vspec2csv.py
@@ -16,6 +16,7 @@ import vspec
 import json
 import getopt
 import csv
+import re
 
 
 def usage():
@@ -82,18 +83,121 @@ def format_data(json_data):
     return f"{Uuid},{Type},{DataType},{Unit},{Min},{Max},{Desc},{Enum},{Sensor},{Actuator}"
 
 
-def json2csv(json_data, file_out, parent_signal):
-    for k in list(json_data.keys()):
+def json2csv(json_data, file_out, parent_signal, ext=[]):
+    for k in json_data.keys():
         if (len(parent_signal) > 0):
             signal = parent_signal + "." + k
         else:
             signal = k
 
+        local_ext = ext
+        if ('extension' in json_data[k].keys()):
+            local_ext = ext + [(json_data[k]['extension'])]
+
+        # if it's a branch, create an entry and continue
+        # with its children
         if (json_data[k]['type'] == 'branch'):
             file_out.write(signal + ',' + format_data(json_data[k]) + '\n')
-            json2csv(json_data[k]['children'], file_out, signal)
+            # if postion attribute exists, keep it
+            json2csv(json_data[k]['children'], file_out, signal, local_ext)
+
+        # if it's a leave, make an entry, check the extension entries and
+        # create an entry for every extension
         else:
-            file_out.write(signal + ',' + format_data(json_data[k]) + '\n')
+            if not local_ext:
+                file_out.write(signal + ',' + format_data(json_data[k]) + "," + str(ext) + '\n')
+
+            createExtensionEntries(local_ext, file_out, json_data[k], signal)
+
+
+
+def createExtensionEntries(ext, file_out, json_data, prefix=''):
+    """Function creates the extension entries to a given node as
+       postfix of the node name (e.g. node.Row1.LEFT)
+
+    Parameters
+    ----------
+    ext : list
+        list of branches with extension argument in the path
+    file_out : file
+        csv file to write into
+    json_data : json object
+        json encoded information about the node
+    prefix : string
+        node name before the extension information
+    """
+
+    REG_EX = "\w+\[\d+,(\d+)\]"
+
+    if not ext and file_out and json_data:
+        return
+
+    rest = None
+    i = []
+    if len(ext) == 1:
+        i = ext[0]
+    else:
+        i = ext[0]
+        rest = ext[1:]
+
+
+
+    if prefix and not prefix.endswith("."):
+        prefix += "."
+
+
+    # parse string extension elements (e.g. Row[1,5])
+    if isinstance(i,str):
+        if re.match(REG_EX, i):
+            extRangeArr = re.split("\[+|,+|\]",i)
+            for r in range(int(extRangeArr[1]),int(extRangeArr[2])+1):
+                nextPrefix = prefix + extRangeArr[0]+str(r)
+                if rest:
+                    createExtensionEntries(rest, file_out, json_data, nextPrefix)
+                else:
+                    file_out.write(nextPrefix + ',' + format_data(json_data) + "," + '\n')
+
+
+        # TODO: right now dynamic extensions not supported
+        else:
+            raise vspec.VSpecError("","","Extension type not supported")
+    # Use list elements for extension (e.g. [LEFT,RIGHT])
+    elif (isinstance(i,list)):
+        for r in i:
+            # if in case of multiple extensions in one branch
+            # it has to be distinguished from a list of
+            # string extensions, like [LEFT,RIGHT]
+            if (isinstance(r,str)):
+                if re.match(REG_EX, r):
+                    if (rest):
+                        rest.append(r)
+                    else:
+                        rest = [r]
+                    createExtensionEntries(rest, file_out, json_data, prefix)
+                else:
+                    nextPrefix = prefix + str(r)
+                    if rest:
+                        createExtensionEntries(rest, file_out, json_data, nextPrefix)
+                    else:
+                        # create a new type instance for now for
+                        # better recognition, of what it is
+                        json_data["type"] = "instance"
+                        file_out.write(nextPrefix + ',' + format_data(json_data) + "," + '\n')
+
+            else:
+                # in case of multiple extensions, the list is
+                # has to be parsed, like [LEFT,RIGHT]
+                if (rest):
+                    rest.append(r)
+                else:
+                    rest = [r]
+                createExtensionEntries(rest, file_out, json_data, prefix)
+
+
+    else:
+        print (i)
+        raise vspec.VSpecError("","","not supported")
+
 
 
 if __name__ == "__main__":
--- a/tools/vspec2csv.py
+++ b/tools/vspec2csv.py
@@ -104,9 +104,7 @@ def json2csv(json_data, file_out, parent_signal, ext=[]):
         # if it's a leave, make an entry, check the extension entries and
         # create an entry for every extension
         else:
-            if not local_ext:
-                file_out.write(signal + ',' + format_data(json_data[k]) + "," + str(ext) + '\n')
-
+            file_out.write(signal + ',' + format_data(json_data[k]) + "," + str(ext) + '\n')
             createExtensionEntries(local_ext, file_out, json_data[k], signal)
 
 
@@ -131,7 +129,6 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
 
     if not ext and file_out and json_data:
         return
-
     rest = None
     i = []
     if len(ext) == 1:
@@ -139,7 +136,6 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
     else:
         i = ext[0]
         rest = ext[1:]
-
 
 
     if prefix and not prefix.endswith("."):
@@ -155,25 +151,27 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
                 if rest:
                     createExtensionEntries(rest, file_out, json_data, nextPrefix)
                 else:
-                    file_out.write(nextPrefix + ',' + format_data(json_data) + "," + '\n')
+                    json_data["type"] = "instance"
+                    file_out.write(nextPrefix + ',' + format_data(json_data) + ",***" + '\n')
 
 
         # TODO: right now dynamic extensions not supported
         else:
             raise vspec.VSpecError("","","Extension type not supported")
-    # Use list elements for extension (e.g. [LEFT,RIGHT])
+    # Use list elements for extension (e.g. ["LEFT","RIGHT"])
     elif (isinstance(i,list)):
+        complexList = False
         for r in i:
             # if in case of multiple extensions in one branch
             # it has to be distinguished from a list of
-            # string extensions, like [LEFT,RIGHT]
+            # string extensions, like ["LEFT","RIGHT"]
             if (isinstance(r,str)):
                 if re.match(REG_EX, r):
                     if (rest):
                         rest.append(r)
                     else:
                         rest = [r]
-                    createExtensionEntries(rest, file_out, json_data, prefix)
+                    complexList = True
                 else:
                     nextPrefix = prefix + str(r)
                     if rest:
@@ -186,13 +184,15 @@ def createExtensionEntries(ext, file_out, json_data, prefix=''):
 
             else:
                 # in case of multiple extensions, the list is
-                # has to be parsed, like [LEFT,RIGHT]
+                # has to be parsed, like ["LEFT","RIGHT"]
                 if (rest):
                     rest.append(r)
                 else:
                     rest = [r]
-                createExtensionEntries(rest, file_out, json_data, prefix)
+                complexList = True
 
+        if complexList:
+            createExtensionEntries(rest, file_out, json_data, prefix)
 
     else:
         print (i)
--- a/tools/vspec2csv.py
+++ b/tools/vspec2csv.py
@@ -48,10 +48,28 @@ Usage: {sys.argv[0]} [options] vspec_file csv_file
     sys.exit(255)
 
 
-def format_data(json_data):
-    Uuid = '""'
+def format_data(json_data, complex=False):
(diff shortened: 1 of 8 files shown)


UlfBj:
Below are two excerpts from the YAML together with the CSV that it generates. 
I see that you introduce a new node type "instance", which can represent any of the node types sensor/actuator/attribute/stream I assume? 
If so, I see an issue with that, and that is it is not in the CSV representation possible to find which it is, only in the YAML source. 

Another issue is shown by the following CSV rows:
Vehicle.Body.Mirrors,"4553","branch","","","","","All mirrors","","",""
Vehicle.Body.Mirrors.Tilt.Left,"4967","instance","int8","percent","","","Mirror tilt as a percent. 0 = 
The node Tilt does not physically exist in the generated tree, it is replaced by Tilt.Left (and Tilt.Right). 
So the Tilt node has actually become a virtual node. Addressing it with Vehicle.Body.Mirrors.Tilt I guess will return both the Left and Right instances?
Virtual nodes can be fine, but also lead to some issues. In W3C automotive there is a discussion about dynamic registration of subtrees. What would happen if this registration tries to anchor the subtree to a virual node?

A third issue is found in the second example below, in the CSV row addressing the node
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Front
I believe this is a Tool bug, it should not be generated?

```
YAML:

- Mirrors:
  type: branch
  extension: ["Left", "Right"]
  description: All mirrors

ExteriorMirrors.vspec:
- Tilt:
  datatype: int8
  unit: percent
  type: actuator
  description: Mirror tilt as a percent. 0 = Center Position. 100 = Fully Upward Position. -100 = Fully Downward Position.

- Pan:
  datatype: int8
  type: actuator
  unit: percent
  description: Mirror pan as a percent. 0 = Center Position. 100 = Fully Left Position. -100 = Fully Right Position.

- Heating:
  type: branch
  description: Mirror heater signals

- Heating.Status:
  datatype: boolean
  type: actuator
  description: Mirror Heater on or off. True = Heater On. False = Heater Off.


CSV file produced by Tool:

Vehicle.Body.Mirrors,"4553","branch","","","","","All mirrors","","",""
Vehicle.Body.Mirrors.Tilt.Left,"4967","instance","int8","percent","","","Mirror tilt as a percent. 0 = Center Position. 100 = Fully Upward Position. -100 = Fully Downward Position.","","","",
Vehicle.Body.Mirrors.Tilt.Right,"4967","instance","int8","percent","","","Mirror tilt as a percent. 0 = Center Position. 100 = Fully Upward Position. -100 = Fully Downward Position.","","","",
Vehicle.Body.Mirrors.Pan.Left,"4968","instance","int8","percent","","","Mirror pan as a percent. 0 = Center Position. 100 = Fully Left Position. -100 = Fully Right Position.","","","",
Vehicle.Body.Mirrors.Pan.Right,"4968","instance","int8","percent","","","Mirror pan as a percent. 0 = Center Position. 100 = Fully Left Position. -100 = Fully Right Position.","","","",
Vehicle.Body.Mirrors.Heating,"4969","branch","","","","","Mirror heater signals","","",""
Vehicle.Body.Mirrors.Heating.Status.Left,"4970","instance","boolean","","","","Mirror Heater on or off. True = Heater On. False = Heater Off.","","","",
Vehicle.Body.Mirrors.Heating.Status.Right,"4970","instance","boolean","","","","Mirror Heater on or off. True = Heater On. False = Heater Off.","","","",


YAML:

- ObstacleDetection.Input:
  extension:
    - ["Front","Rear"]
    - ["Left", "Right"]
  type: branch
  description: Direct sensor input on obstacle detection

- ObstacleDetection.Input.DistanceToObject:
  datatype: uint16
  type: sensor
  unit: m
  description: Distance to object in meters

CSV file produced by Tool:

Vehicle.ADAS.ObstacleDetection,"4895","branch","","","","","Signals form Obstacle Sensor System","","",""
Vehicle.ADAS.ObstacleDetection.IsActive,"4362","actuator","boolean","","","","Indicates if obstacle sensor system is enabled. True = Enabled. False = Disabled.","","","",[]
Vehicle.ADAS.ObstacleDetection.Error,"4363","sensor","boolean","","","","Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.","","","",[]
Vehicle.ADAS.ObstacleDetection.Input,"5048","branch","","","","","Direct sensor input on obstacle detection","","",""
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Front,"5049","instance","uint16","m","","","Distance to object in meters","","","",
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Rear,"5049","instance","uint16","m","","","Distance to object in meters","","","",
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Front.Left,"5049","instance","uint16","m","","","Distance to object in meters","","","",
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Front.Right,"5049","instance","uint16","m","","","Distance to object in meters","","","",
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Rear.Left,"5049","instance","uint16","m","","","Distance to object in meters","","","",
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Rear.Right,"5049","instance","uint16","m","","","Distance to object in meters","","","",
```

danielwilms:
> I see that you introduce a new node type "instance", which can represent any of the node types sensor/actuator/attribute/stream I assume?
If so, I see an issue with that, and that is it is not in the CSV representation possible to find which it is, only in the YAML source.

The instance is the one for discussion here. I just thought it's easier talking about it with an example. Thanks for the bug reports. I'll fix them.

I'd propose to go in two steps. First let's discuss the specification in #130 and then on the tooling and representation, to eventually close #81 :)

tools/vspec2csv.py
```diff
@@ -81,18 +82,121 @@ def format_data(json_data):
...
+                if rest:
+                    createExtensionEntries(rest, file_out, json_data, nextPrefix)
+                else:
+                    file_out.write(nextPrefix + ',' + format_data(json_data) + "," + '\n')
+
+
+        # TODO: right now dynamic extensions not supported
+        else:
+            raise vspec.VSpecError("","","Extension type not supported")
+    # Use list elements for extension (e.g. [LEFT,RIGHT])
```
rtroncy: # Use list elements for extension (e.g. ["LEFT","RIGHT"])

UlfBj: >> I'd propose to go in two steps. First let's discuss the specification in #130 and then on the tooling and representation

I think the hard one is the resulting tree, not the YAML, so I think they need to be discussed together, or start with the Tool result. 
Personally I find the "virtual" node model (see my previous comment) a bit problematic, that for example the node Tilt physically exists in YAML, but not in CSV.
I mentioned the scenario with dynamic registration into this virtual node as an example, please comment how that could be handled.

UlfBj:
>> I'd propose to go in two steps. First let's discuss the specification in #130 and then on the tooling and representation

I think the hard one is the resulting tree, not the YAML, so I think they need to be discussed together, or start with the Tool result. 
Personally I find the "virtual" node model (see my previous comment) a bit problematic, that for example the node Tilt physically exists in YAML, but not in CSV.
I mentioned the scenario with dynamic registration into this virtual node as an example, please comment how that could be handled.

danielwilms:
> Personally I find the "virtual" node model (see my previous comment) a bit problematic, that for example the node Tilt physically exists in YAML, but not in CSV.

This is a bug, I'll fix. Thinking of dynamic registration it should anyway only be allowed on branch nodes I suppose.

tools/vspec2csv.py
danielwilms: Thanks, @rtroncy. Noted to be fixed.

UlfBj:
>> This is a bug, I'll fix.

Good. With that node physically present in the tree I think the new node type "instance" is Ok. The properties in an instance node I think should only be the minimum mandatory - name, type, descr, as it inherits the properties of its "parent leaf node". 

>> Thinking of dynamic registration it should anyway only be allowed on branch nodes I suppose.

You are right. 
I think I was also wrong on the previous comment:
>> A third issue is found in the second example below, in the CSV row addressing the node
Vehicle.ADAS.ObstacleDetection.Input.DistanceToObject.Front
I believe this is a Tool bug, it should not be generated?

They should be generated, but I think they are a good example of that properties from parent leaf should not be copied to them. This is logically equivalent to a branch node, to then have datatype and unit properties in it is confusing.

danielwilms:
@UlfBj: I changed now the code and fixed the bugs. We should still discuss the instance representation. Two things in that regard:

1. should the meta information be present fir the instances as well?
1. is the dot-notation appropriate for instances?

UlfBj:
>> should the meta information be present fir the instances as well?

I think is should not be present. Only the VSS mandatory properties should be present: name, nodetyp, and description (the latter being something like "Instance of <node name>" where node name is the "leaf" node name it instantiates. 
>> is the dot-notation appropriate for instances?

I think the instantiated nodes shall be addressable, so yes. A client should not have to handle them differently than any other node.

danielwilms:
> I think the instantiated nodes shall be addressable, so yes. A client should not have to handle them differently than any other node.

Makes sense. And the protocol decides how they are served then. I'll change the PR then.

UlfBj:
In the latest CSV it looks like below
The row 
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel
is shown, i.e. the node is "physically" in the tree. Good:).
But the row for the nodes
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Left
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Right
are now missing (which was what I asked for, but I changed my mind:). 
As the nodes are "physically" in the tree (I hope?) then they should appear in the CSV. 
They also then are a good example for why I think metadata from the leaf node should not be present in instances, as these two nodes are logically equivalent to a branch node, not a leaf node.
```
Vehicle.Chassis.Axle.Wheel.Brake,"5056","branch","","","","","Brake signals for wheel","","",""
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","sensor","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",[['Row[1,2]'], ['Left', 'Right']]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Left.Row1,"5057","instance","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",***
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Left.Row2,"5057","instance","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",***
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Right.Row1,"5057","instance","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",***
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Right.Row2,"5057","instance","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",***
```

danielwilms:
@UlfBj:
> But the row for the nodes
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Left
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Right
are now missing (which was what I asked for, but I changed my mind:).

I browsed again through the positioning thread and maybe the dot notation isn't the best way to use for the instances, when I think of addressing them. If we add a field to the CSV including the instance information, like you described [here](https://github.com/GENIVI/vehicle_signal_specification/issues/81#issuecomment-463140954), that would leave the addressing to the implementation.

UlfBj:
>>  If we add a field to the CSV including the instance information, like you described here, that would leave the addressing to the implementation.

Could you please exemplify?

danielwilms:
> Could you please exemplify?

Of course :) That's what I thought:

```
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","sensor","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",[['Row[1,2]'], ['Left', 'Right']]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row1","Left"]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row1","Right"]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row2","Left"]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row2","Right"]
```

UlfBj:
>> That's what I thought:

In my understanding of the example above, there is room for many interpretations.
How would the address in the client request to one of the instances look like?
Would the four instances in the example above be direct children to the FluidLevel node?
What would be the node name of the node represented by the row below?
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row1","Left"]

danielwilms:
> In my understanding of the example above, there is room for many interpretations.

I would separate spec and usage here. From the spec perspective it's just an instance of relationship:

example_tree

From a protocol perspective you have to think on how you access the instances.

tools/vspec2csv.py
UlfBj: Generalize multidimensional instantiation also to list elements.
E.g.
```
extension:
    - Row[1,4]
    - ["Left","Right"]
```
becomes
```
extension: [ Row[1,4], ["Left","Right"]]
```

UlfBj:
The current data model with its parent-child relationship will have to be extended with the "instanceOf" relationship, which I think "costs more than it tastes". 
As Tools are part of this project, we need to define it down to the implementation details of the actual tree. 
The alternative to represent the instance nodes with "physical" nodes basically works "out of the box" with the existing model (the node type "instance" must be added). Less is more:).

gunnarx:
> The current data model with its parent-child relationship will have to be extended with the "instanceOf" relationship

I think I agree with Ulf.  Overall, we started with what seemed to me a fairly simple extension/instance mechanism in #81.  And primarily it was there to reduce duplication.  We then found that the devil was in the details. 

Of course it can be useful to draw the instance relationship (like the picture you showed) just for our shared understanding, so that's fine.  But overall, my wish for this is that we don't complicate VSS itself and therefore we in particular don't "highlight" a lot of complex computer-science theory unless truly required for the model definition.

I mean to avoid introducing these concepts too much into the actual model explanation, and into its usage.  It's so easy to over-engineer things and difficult to keep things simple.  VSS should still be expressive enough, so we need just enough, but don't fall into the trap of growing the number of concepts too much.  (I'm in fact already a little skeptical about the many branch types and on, but that is a separate issue).

I also feel that "equivalence" is key here.  I thought it should not matter much if something is an instance or not?  As Ulf implies, an instance just becomes a child.  Isn't it that simple?  It's just a way to efficiently avoid duplication in the definition of the tree?

Would you agree first that it should be a free choice of the data model designer, and the end result should be equivalent:

``` 
A.B.C:   # (or A.B.C.instance or A.B.C._ , syntax is being discussed still)
   - description: This is ABC
   - unit: milliflops
   - datatype: int32
   - instance: ["Left","Right"]
```
should yield nodes:
```
A.B.C.Left
A.B.C.Right
```
each of the nodes having basically identical/duplicated metadata (except the instances of course), that is the following:
```
   - description: "This is ABC"      <- here tools could choose to add something like ", instance Left"
   - unit: milliflops
   - datatype: int32
```

and I think this should be equivalent, in usage, to if the data model designer chose to write this in YAML instead:
```
A.B.C.Left:
   - type: ...
   - description: "This is ABC"
   - unit: milliflops
   - datatype: int32
A.B.C.Right:
   - type: ...
   - description: "This is ABC"
   - unit: milliflops
   - datatype: int32
```

With that said, I don't immediately like to focus on the CSV format that much.   I suppose maybe in communication between humans and as documentation it can be nice to bring it into spreadsheet.  But mostly I see this just as a way to generate to see what effect the instantiation had?  Is that also your intention with the PR, Daniel?  Or what do we think is really the intention of the CSV format, strictly speaking.  The tool has been there for a while, but why?  Just because?
 
If that CSV format makes a note of that a node was created as an instantiation instead of being listed in the original YAML (as done in Daniel's example) I can of course see how this information can be useful to have - just to know it came from an instance.  But the rest should show equivalence.

In your example however:

```
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","sensor","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",[['Row[1,2]'], ['Left', 'Right']]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row1","Left"]
```

...here you are writing "instance" in the same column as the type "sensor" and thereby suggesting it is a new type of node.  No, I don't like it really...  The type of an instance is surely still "sensor" isn't it?

As already mentioned, separate nodes with the same UUID feels also not right.  Each one must be separately addressable and **if** we allow for UUID addressing it must be unique.  

It's not that you could not generate a CSV like this - you could generate like 10 different formats if you wish (again, for understanding, documentation, whatever the user feels like) but if we are speaking of a _definition_ of the model and we use it to explain to explain VSS others then I do not like it if it adds new concepts and more complexity.

The most typical expansion of the YAML file into its instances (and I guess therefore the most typical TXT or CSV format file) should in my opinion list 4 lines in which the full name is fully expanded:

```Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row1.Left
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row1.Right
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row2.Left
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row2.Right
```

In fact, I thought this has been very clear in examples from both me and Ulf in #81?

Stepping back from this to the big picture, we must focus on simplicity -- not to become the most complex and expressive model ever.   I don't think VSS will ever become a full fledged object model.
If further work on VSS-Ontology could create that however I'd love to see this, but I suggest it should be defined as a separate extension/layer on top of standard VSS in that case.

For a perfect and fully formal model we would have to use or define a much more capable modelling language, and that becomes too much here, IMHO.   Also don't forget that in the grander scheme of things, VSS is complemented by things like Franca IDL (which covers functions/methods, interface hierarchies, and arbitrarily complex structs / data types as well), Franca Plus (which covers components and architectures, in principle replacing parts of UML even), and therefore I think VSS should aim first to be simple but capable data/signal model that is directly usable with minimum problems.

gunnarx:
Sorry for the wall of text above, just spinning off of this into principles and thinking :)

> From a protocol perspective you have to think on how you access the instances.

This is the key I think.  We need to consider how protocols and data model are related, it is fairly unaviodable in my opinion.  I want us to envision that there is no new and different way to address instances, because it would then not have the equivalence I mentioned above.  If we get to a point where instances are children, quite simply, then whatever the protocol the way to address them is already defined (the same way as you address a child node).

gunnarx:
@UlfBj comments suggest that this...
```
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","sensor","uint8","percent","","","Brake fluid level as percent. 0 = Empty. 100 = Full.","","","",[['Row[1,2]'], ['Left', 'Right']]
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel,"5057","instance","","","","","","","","",["Row1","Left"]
```
...is fortunately not what the tool is actually generating.  Maybe I missed some of the details on the discussion in between.  Anyway, my comments was on Daniel's most recent entry so they should be valid.

> They also then are a good example for why I think metadata from the leaf node should 
> not be present in instances, as these two nodes are logically equivalent to a branch node, 
> not a leaf node.

Here we get to the crux of some different way of thinking I think.  All in all, I care about what is written in the YAML.  That is the formal spec and everything else is kind of just extra.  
I further believe that if we speak of the semantics and meaning of that YAML, then we have talked about creating "instances" and to me those are in effect several copies created from the node where the "instances: [Left, Right]" appears.  In other words they must have the metadata in each of those.  So when you say "should not be **present**" it is really not clear to me in what context.  Present where and in what format?   We can agree that copies of metadata *are* not present in the YAML (because only the defining node exists, with its instance/extension lines).  And so, you mean in a particular CSV format, or inside of some understanding of what the resulting data model is, after instantiation occurred?

As a parallel to explain how I think about this, in the internal representation of the data model inside of an actual program, it would be seem useful if instantiated nodes simply "have" all of their meta data (so you do not have to traverse a tree to find it) but that is of course completely an implementation detail.  Both ways of doing it is possible and it does not affect external behavior.  So that part has nothing to do with the spec definition.

For any other generated format, such as CSV, as I mentioned I think they are just superfluous formats and have very little to do with a definition of the data model.  Useful for discussion, documentation, etc.  but not the formal files.  So if we decide to duplicate the metadata in such a generated file, I don't see a problem with it.  Actually in my view of instantiation that would be the "correct" definition of a generated/instantiated node.

But that gets us to your actual opposition which is branch vs. leaf.  I think details depend on which proposal we go for (in #81) but in general I see no problem with changing *that* metadata only, as needed (i.e. let a leaf become a branch if it gets instances below it), but I know this is where we differed in our thinking and proposals before.  In any case, the key in my opinion is that some particular metadata (in this case node type only) is required to change to be consistent, just let it change.  It does not mean to me that all other meta data (datatype, unit, whatever) cannot or should not be copied.   (Again I'm not 100% comfortable with the word "copied"... because in some implementations this might be a non-issue even - it might make use of the YAML directly + semantic rules of how to interpret that YAML).  So it all depends on what output format we are actually talking about as I see it.  But I could imagine tooling that reads the YAML and outputs a CSV file to show the instances with all of their relevant data that they "inherited" from their parent being copied therein, and branch/leaf just defined to whatever it should be according to the model we agree upon.

If you don't agree with this, I think I need further understanding of what you meant by "not present".

UlfBj:
>> So when you say "should not be present" it is really not clear to me in what context

I am talking about the Tool output. 
In "multidimensional" instantiations (Row1.Left as an example) I think copying all metadata from the YAML leaf into all of these should be avoided. Only the mandatory properties (name, type, descr) should be copied. This of course means a server would need to traverse up the branch to the (YAML) leaf node to get the metadata. This eliminates redundant data. It also in my view makes it clearer, particularly for a node like Row1 in my example above, where this node logically represents a branch node. 
Besides the above, Isee only two other points where your (@gunnarx ) view and mine differs (here we go again:), and that is that I think in YAML the instantiation should only be allowed in branch nodes, and that the instantiated nodes should have node type "instance" (or whatever name we agree on).

danielwilms:
> But mostly I see this just as a way to generate to see what effect the instantiation had? Is that also your intention with the PR, Daniel?

Yes, that's it! Further, the dot notation shows how it should be addressed

> If we agreed that instances are quite simply children, then for every protocol, the way to address them is already defined (the same way as you address a child node).

The idea with the instances came from the fact that on a protocol level you should be able to decide to just address all left tires, or in opposite all of row1, but I think you are right with the over-engineering.

```
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row1.Left
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row1.Right
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row2.Left
Vehicle.Chassis.Axle.Wheel.Brake.FluidLevel.Row2.Right
```

This looks good and logical to me. But I second to have the meta-data there. @UlfBj, can you agree to this.

I only would like to have a column in the csv, where it's mentioned that it's an instance, please :)

UlfBj:
>>  @UlfBj, can you agree to this.

I am soon at a level where I agree to anything and everything:)
I stated my view just before your latest comment, but if you two agree I will not persist.

UlfBj:
We could list the differences between our views, and then vote on them separately. 
My take on such a list:
- Shall in YAML the instance declarations be allowed in only branch nodes, or only/also in leaf nodes?
- Shall in Tool output, the instantiated nodes have the node type "instance" (or other name), or inherit from the YAML leaf node?
- Shall all properties be copied from the YAML leaf node, or should only mandatory properties be present in the instance nodes?
Please add to the list what you think is missing.

gunnarx:
> The idea with the instances came from the fact that on a protocol level you should be able to decide to just address all left tires, or in opposite all of row1

Yeah I thought that might be the thinking behind.  But in my opinion the query syntax, for example simple wildcards (part of protocol definition, not VSS definition) already solve this problem.

gunnarx:
It's a good start for a list, Ulf.  

But this one... I wonder just to understand what you envision on this one:

> Shall all properties be copied from the YAML leaf node, or should only mandatory properties be present in the instance nodes?

I'm still not sure how you mean the alternative can work.  I believe an instance must copy all characteristics of the defining node, otherwise they are not really instances?  The scary thing here is that we get into complexity because this suggests in my mind the following question: Shall overriding be allowed?  E.g. by defining an explicit node in YAML that overlaps an instance and redefines some property.  If you cannot add or change the non-mandatory properties somehow, then they would simply be missing.  In that case I imagine getting a copy of the "parent" is really the right way.

Otherwise I feel like the net effect is that instantiation is only useful for nodes that do not have any additional/proprietary/non-mandatory properties?

Sorry if I missed it but is there anyone (you) proposing the copying of mandatory properties only?  Or was it just a possibility you thought of now.

gunnarx:
@danielwilms  wrote
> I only would like to have a column in the csv, where it's mentioned that it's an instance, please :)

Yes, after some rambling this is also what I found to be OK.  I think the type should be the parent, but indicating that the node is created as a result of an instance definition is just good additional information.

gunnarx:
> This eliminates redundant data

These words are the key.  In my world view there is nothing redundant because the YAML is king.
I believe you are focusing on the "result" more clearly than I do.

Honestly in any auto-generated information (whether it is CSV, source code, or the binary data structure that a particular implementation chooses to store the tree) I don't think we should care whether there is redundancy.  There might be...  There will be...  And there's really no controlling that if you look at the entire chain all the way to a compiled program.

What are the main reasons to avoid redundancy in the definition of computer systems?   In my opinion the following:

A) Avoid any risk of inconsistent or different results
(especially in the case that some part is changed)
B) Avoid duplicated (human) effort

Then there are two points to highlight with regard to _generated_ results, IMHO:

1) Since it is generated from a computer program (tooling) it cannot be incorrect or inconsistent, once the generator is debugged.  

2)  Second, the generated result should not be seen as the official specification anyway so it does not matter because the YAML is that specification, in my world view.  

Both 1) and 2) independently eliminate or make non-applicable the concerns of A) and B), as I see it.

danielwilms:
> 2. Second, the generated result should not be seen as the official specification anyway so it does not matter because the YAML is that specification, in my world view.

I agree, only the resulting dot notation helps to interpret the specification correctly by visualising it, especially after instantiation.

>>This eliminates redundant data
> These words are the key. In my world view there is nothing redundant because the YAML is king.
I believe you are focusing on the "result" more clearly than I do.

I do focus on it more, because that's what developers see and look at, because this is what they have to use at the end. If I talk about it, people ask immediately: show me the tree. So, the serialization is used for communication, the yaml to formally describe it.

UlfBj:
>> I believe you are focusing on the "result" more clearly than I do

Yes, I also do. Besides the reasons given by @danielwilms above, another argument for keeping redundancy low in the Tool output tree is that it is transmitted to the client as the response to a getmetadata request, so it reduces traffic load.

>> I'm still not sure how you mean the alternative can work. I believe an instance must copy all characteristics of the defining node, otherwise they are not really instances?

Whether you copy the metadata into the instance or not, logically the instance inherits it from its (YAML) parent node. So if it is not copied, the server parsing the instance node needs to go to the parent node for the metadata. Logically no difference, implementation wise a small difference. 

Shall we try to list the differences, or how can we come to a final resolution?

danielwilms:
After some thinking and discussions I hope I can close the two PRs with this proposal. I tried to solve especially Ulfs concerns about a leaf not being the one carrying the data, but rather acting as a branch. As base I take the spec from #130.

I this PR I introduced a new field "complex":
> In order to map the instances to the csv structure, complex types are introduced. This means, that a leaf with attached instances is a typed list and that instances have to be anticipated.

At the end you create with the instances a kind of typed list of items.

UlfBj:
+1.
I think the simplification of the YAML files is a great improvement. Regarding the addition of the new property "complex", I think it should be replaced by adding a new node type "instance" that is assigned to the nodes where "complex"= true. But that is more a matter of taste I guess.

magnusfeuer:
@danielwilms Asked me to have a look at this PR (and #130). 

It seems like you have it covered. My only question is why instances are implemented in `vspec2csv.py` instead of the `vspec2.py` module? It would be nice if we could utilize instances in more output formats than only CSV.

That said, since this has been a long run, we could approve this PR and then do a second run at migrating the code over to `vspec.py`.

danielwilms:
> That said, since this has been a long run, we could approve this PR and then do a second run at migrating the code over to vspec.py.

@magnusfeuer: you're right. Will open an issue after it's merged and move it.

gunnarx:
Looks OK to me. +1 on putting this in the common module of course.

Your branch is now very outdated and rebasing would help.   
As GitHub indicates, rebase or merge would create a small conflict.   I have rebased and fixed the conflict on my side. (Result https://github.com/gunnarx/vehicle_signal_specification/tree/pr_131)
Let me know what you want to do.

gunnarx:
Oh, I just pushed the rebased results.  I forgot I can do that on the GENIVI/ repositories
(Note that since it is a PR this actually modifies your source branch, Daniel)

danielwilms:
@gunnarx, Thanks for the rebase. 

@magnusfeuer: I created #136 

With approvals from @klotzbenjamin, @UlfBj, @tguild, @magnusfeuer and @gunnarx I'll merge.

Issue #132, opened 2019-11-19
VSS versioning


tguild:
Include VSS version within the tree itself, that way VISS, Gen2 etc can distinguish which is being provided by the service

https://github.com/w3c/automotive/issues/307

PR #136, opened 2020-02-11
Versioning: Added VSS Versions to VSS tree

danielwilms:
As the concept of the tree changes over time, it is important for the
client to know, which version it's expecting. Split it into Major,
Minor, Patch to avoid unnecessary string parsing.

Fixes: #132

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -12,6 +12,29 @@
 #
 
 #
+# VSS Versioning information
+#
+
+- VersionVSS:
+  type: branch
+  description: Supported Version of VSS
+
+- VersionVSS.Major:
+  datatype: uint8
+  type: attribute
+  description: Supported Version of VSS - Major version
+
+- VersionVSS.Minor:
+  datatype: uint8
+  type: attribute
+  description: Supported Version of VSS - Minor version
+
+- VersionVSS.Patch:
+  datatype: uint8
+  type: attribute
+  description: Supported Version of VSS - Patch version
+
+#
 # Vehicle identification attributes.
 #
 
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -20,19 +20,24 @@
   description: Supported Version of VSS
 
 - VersionVSS.Major:
-  datatype: uint8
+  datatype: uint32
   type: attribute
   description: Supported Version of VSS - Major version
 
 - VersionVSS.Minor:
-  datatype: uint8
+  datatype: uint32
   type: attribute
   description: Supported Version of VSS - Minor version
 
 - VersionVSS.Patch:
-  datatype: uint8
+  datatype: uint32
   type: attribute
   description: Supported Version of VSS - Patch version
+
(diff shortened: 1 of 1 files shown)


gunnarx:
When going from the very flexible string to ints we have to be really careful that we don't introduce unnecessary limitations.

It's hard to tell exactly how this will be used.  The official VSS signal content will of course not be released that often.   But internal proprietary VSS-files (additions, VSS-layer files or whatever it may be) could very likely be revised a whole bunch of times.  Maybe some companies will bump the number rarely, but some other will bump the patch number daily, in a very agile and often changing process.

So I' d say why not make each number at least a 16 bit int, or even 32...    This information is not something that is repeated over and over, so saving memory is not the concern...

tguild:
@gunnarx raises an interesting point.

Those using private branches will want to know which GENIVI VSS version *and* their proprietary branch. It may be worth adding a common attribute, perhaps string as they may prefer name/label to number, to identify their branch version.

danielwilms:
@gunnarx, @tguild check 575e677

UlfBj:
>> So I' d say why not make each number at least a 16 bit int, or even 32

I agree.
Good point.

BR
Ulf

Den fre 14 feb. 2020 kl 12:38 skrev gunnarx <notifications@github.com>:

gunnarx:
@tguild 
> Those using private branches will want to know which GENIVI VSS version and 
> their proprietary branch. It may be worth adding a common attribute, perhaps string 
> as they may prefer name/label to number, to identify their branch version.

My thinking was basically not to define the versioning for the private branches, but only envisioning that the mechanism we use might have the *ability* to be reused the same way  for (frequently updated) private layers.   That said, I don't mind the name/label addition in the latest commit.

To elaborate a bit, it is  hard to predict all the ways that private/proprietary signals and metadata might be used. (It would be ideal to avoid proprietary VSS information as far as possible, but it cannot be fully avoided).   This is even more so in the light of the larger VSS layers discussion.    Going even further in adding a standard for versioning of proprietary extensions, I fear could risk being inadequate.  There might be more than one private branch for example and there might be layers we haven't thought of yet... -> it is hard to not inadvertently create limitations here.  (still open for more suggestions though)

So one idea is to say that **since** anyone can define signals/attributes in their private branch / layer, they can also decide on their own how to write their own version information in there.  My attempt was just to remove one possible incompatibility.

We could put that note (the best practice for versioning) in a guidance document, but still allow those that write the private layers to decide in the end how to do it.

Thanks for bringing it up.  In my opinion we likely have the right balance with the latest commit.  The string could be used to put a git commit hash or similar thing that uniquely identifies the combination of standard+proprietary extension.

gunnarx:
+1 on 575e677

danielwilms:
> In my opinion we likely have the right balance with the latest commit.

Thanks for all the input, I second that. It'll be interesting to see how it's used. From my perspective, I wouldn't use the version for individual signals, but rather for the structure I've to expect. At the end, you cannot rely on that a specific node is there. You'll need in the protocol later some discovery. 

With that said. I'd merge it based on acceptance of @tguild, @gunnarx and @UlfBj

PR #137, opened 2020-02-11
Example VSS layers for defining access control (Discussion material)

gunnarx:
DO NOT MERGE.  This is discussion material.

The intention was to show access control groups useful for the W3C protocol.  Unfortunately an Android-related example was pushed instead.   Maybe later I have time to clean up and force push an update.

For now, the concepts should be still possible to derive from reading this.

### DIFF
--- a/vss_layers/README.md
+++ b/vss_layers/README.md
@@ -20,20 +20,103 @@ There is one main way to add a metadata relationship in the plain VSS
 (meta)model.  In the VSSo (ontology) environment there may also be additional
 ways.
 
-List new metadata below a signal definition:
+
+Example how to list new metadata below a signal definition:
 
 VSS core model metadata:
 ```
-Motor.CoolantTemperature
+Motor.CoolantTemperature:
   datatype: int16
   type: sensor
   unit: celsius
   description: Motor coolant temperature (if applicable).
 ```
 
+### Removal of nodes
+
+Various ideas:
+
+Removal example, using empty value = removal
+```
+Motor.CoolantTemperature:
+  datatype: int16
+  type:              <- removes "type:" entirely (*if* we allow type to be removed)
+  unit: celsius
+  description: Motor coolant temperature (if applicable).
+```
+
+Removal example, using no meta-data = remove signal
+```
+Motor.CoolantTemperature:  <- removed because meta-data is empty
+
+Vehicle.Powertrain.Transmission.Speed:
+  datatype: int32          <- modified
+```
+
+Removal example, requires all signals to be named in "deployment file" otherwise they are not included
+(in a code generator for example):
+
+"deployment file"
+```
+Motor.CoolantTemperature:   <- signal mentioned, therefore included
+   deployment_stuff: foo
+```
+
+### Tool invocation
+
+Example with flags to define if it's a (m)odel or (R)emoval file:
+```
+ $ mytool -m vss_rel_2.yaml -m privacy_layer.yaml -m deployment_info.yaml -R stuff_to_remove.yaml
+```
+... in this case nodes listed in stuff_to_remove.yaml gets removed.
+
+
+Example with flags to define if it's a (m)odel or (I)nclusion file:
+```
+ $ mytool -m vss_rel_2.yaml -m privacy_layer.yaml -m deployment_info.yaml -I deployment_whitelist.yaml
+```
+... in this case any node NOT listed in deployment_whitelist.yaml gets removed (not used in output).
+
+Proposal: "Allow" tools to include equivalent of -R and -I flags, one or the
+other or both, depending on need.
+
+Also support a manifest file instead of listing on command line
+```
+ $ mytool -f filelist.txt
+```
+
+filelist.txt example 1:
+```
+ vss_rel_2.2
+ +vss_privacy_info
+ +deployment
+ -removed_signals
+```
+
+filelist.txt example 2:
+```
+ basemodel: vss_rel_2.2
+ layer: vss_privacy_info
+ modification: ...
+ deployment: mydepl.yaml
+```
+
+filelist.txt example 3:
+```
+ model: vss_rel_2.2
+ model: vss_privacy_info
+ model: ...
+ Removal: mydepl.yaml
+```
+
+
+TODO:
+- Define how to modify instances
+
+
 Example of layered additional metadata:
 ```
-Motor.CoolantTemperature
+Motor.CoolantTemperature:
   my_unique_concept: true
 ```
 
--- a/vss_layers/README.md
+++ b/vss_layers/README.md
@@ -1,0 +1,243 @@
+# VSS-Layers
+
+VSS layers is an extension mechanism that has been implicitly included since
+the VSS development started, and recently there have been more discussions to
+formalize the concept.
+
+Even before the concept had a name it was in practice implemented in the
+initial tools because they behaved like layers in the following ways:
+
+1. Overriding definitions was possible with subsequent re-definition of
+   the same signal
+2. It was possible to add custom metadata fields to each signal, in
+   addition to the VSS "core" model metadata ("name", "type", "datatype",
+   "description", and in recent editions "instances", "deprecation" etc.)
+
+There are two main ways to add a metadata relationship in the plain VSS
+(meta)model.  In the VSSo (ontology) environment there may also be additional
+ways.
+
+The most obvious one is to list new metadata below a signal definition.
+
+VSS core model metadata:
+```
+Motor.CoolantTemperature
+  datatype: int16
+  type: sensor
+  unit: celsius
+  description: Motor coolant temperature (if applicable).
+```
+
+Example of layered additional metadata:
+```
+Motor.CoolantTemperature
+  my_unique_concept: true
+```
+
+### A note about signal names and paths
+
+The VSS is defined by a hierarchy of sub-trees each defined in their own file.
+
+An example file (ElectricMotor.vspec) may include:
+
+```
+Motor.CoolantTemperature
+  ...
+  ...
+```
+
+but since this information is in the file ElectricMotor.vspec, and
+that file was included at the location of Vehicle.Powertrain.ElectricMotor,
+the above shortened definition actually spells out the following full-path
+signal:
+
+`Vehicle.Powertrain.ElectricMotor.Motor.CoolantTemperature`
+
+This principle for file-inclusions and namespacing applies to VSS-layers in
+the same way as the core VSS.  For full details, refer to the VSS
+documentation.
+
+Note however, that later examples in this text will use the full path for
+clarity.  E.g.: `Vehicle.Chassis.Wheel.Brake.Fluidlevel`
+
+## Usage of VSS-layers
+
+VSS-layers is a fully generic extension mechanism that is likely to find new
+uses over time, but some of the primary drivers are:
+
+1. The general separation of a "deployment model" that is not included in the core definition.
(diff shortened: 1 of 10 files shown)


gunnarx:
The diff is quite useless, probably better to just view the files here:
https://github.com/gunnarx/vehicle_signal_specification/tree/vss_layers_generic/vss_layers

gunnarx:
A [README](https://github.com/gunnarx/vehicle_signal_specification/blob/vss_layers_generic/vss_layers/README.md) has been added to define the VSS layers concept.

BUT once it is agreed, the description should probably be placed with documentation and the examples should probably not be merged into the actual repo either.  They are just here for viewing.

Therefore, still Do-not-merge, but please give feedback on the definition.

erikbosch:
Would the "layer" concept in some way be possible to use to specify which datapoints that are supported/mandatory/optional. Today we have a lot of datapoints, but no way to specify if the datapoints are mandatory or optional. 

Assume e.g. the fleet management system of Hertz require certain data. Then Hertz could define a layer (or profile?) specifying which datapoints that are required, and whether read or write access is necessary. Maybe something like below (ignore the syntax) - all vehicles in the fleet must support a small standardized set of datapoints (e.g. VIN, length, weight) and in addition to this information on whether the vehicle has HVAC and a Hertz specific datapoint on whether the car has StuddedTyres mounted or not.

With this information it would be possible to verify or test that a vehicle fulfils the requirements of the rental company.

```
/* A rental company like Hertz could add additional requirements that must
   be supported by all vehicles in their fleet  - all */
Hertz.mandatory_read {
  default: 
    cvii.default_passenger_vehicle; /* A small standardized list of datapoints all vehicles must support, like e.g. VIN */
  true: 
    Vehicle.VehicleIdentification.ACRISSCode,
    Vehicle.Cabin.HasHVAC, /* Info if vehicle has hvac or not */
    Hertz.StuddedTyres; 
}

/* Another layer/property can be used to define which datapoints that Hertz must have write access to */
Hertz.mandatory_write {
  default: 
    false;
  true: 
    Hertz.StuddedTyres;
}
```

UlfBj:
I have the following comments after reading the README file.

>> The new concept can use namespacing as deep as desired (in this example, the prefix vss.remote-interface.permissions. is used by all definitions). The definitions themselves may also add additional branches (in this case gnss. ), as desired.

I do not understand the above. To keep it understandable I think the Reverse layer example should use the same metadata as used in the Standard layer example (extended with multiple ECUs to show the array concept). Then the above namespacing concept could be explained separately. 

The Reverse layer concept is only applicable for bullet 3 "Add new metadata", I believe?

I think the Tool support chapter should describe one design at a level that can be used for implementation, and then leave the door open for alternative designs/implementations. An implementation of the described design should then also be found at the repo.

vss_layers/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+1. The general separation of a "deployment model" that is not included in the core definition.
+1. Defining supplementary layers that deal with, for example, data classification.
+1. Ability for companies to tie VSS into existing software and processes that needs additional concepts or information
+
+Some of these usages may lead to agreed-upon layers used by the whole industry, a.k.a. standard layers.
+Others will be system, product, or company specific.
+
+### Deployment model
+
```
SebastianSchildt, line 76: I feel this whole section is a lot of text, without saying much. 

I would suggest keeping paragraph two  `A Deployment Model keeps the information necessary to realize the
 described interface in a particular environment...` and probably adding some more clear examples such as "values" for attributes or min/max metadata

vss_layers/README.md
```diff
@@ -0,0 +1,243 @@
...
+1. Modify existing metadata and ultimately redefine any signal (i.e. change
+   its metadata) compared to a previous definition for example the definition
+   in the standard catalog.
+1. Add new signals. It can be done in the private/ branch as recommended but the
+   mechanism allows adding them anywhere.
+1. Add new metadata, or relationships to new concepts for the purpose of
+   creating deployment-models, bindings to existing technology, classification
+   and many other things.
+1. Remove signals from the final data model (SYNTAX TO BE DEFINED)
```
SebastianSchildt, line 153: I know we discussed this, but I am not super-sure we need it.
After all,  an implementation can still give "404" or "40x not implemented" on signals from the "core" spec. I think it would be a more clean design, than letting layers "delete" anything (i.e. I would rather provide a deployment/stack specific layer that tags "not-implemented: true" to relevant leaves, and letting the implementation act accordingly)

Just an opinion though, I won't fight to the death for it

vss_layers/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+#### Reverse Layer
+
+The second way is to flip the definition around.  First define the new
+concept, and below it list metadata (that is specific to the concept).
+The definitions of the metadata is is likely to refer back to the VSS signal
+names.
+
+Let's call that a `Reverse Layer`.
+
```
SebastianSchildt, line 189: I think this section "breaks" the model.

I think it is "legal", and "possible", because it is akin to just extending private data (i.e. creating new branches/subtrees) and filling whatever data you like.

But in this specific example I would assume it is bad practice, so we should not endorse it.

SebastianSchildt (commented):
_Generally_  speaking (because the individual comments are, of course all nitpicks :) )I think this is largely fine and pretty much describes, what we want.

My only strong feeling is, that the "Reverse layer" is a very  bad (and actually redundant)  concept. But maybe I didn't get it.

gunnarx:
@erikbosch wrote: 
> Would the "layer" concept in some way be possible to use to specify which datapoints that are supported/mandatory/optional. Today we have a lot of datapoints, but no way to specify if the datapoints are mandatory or optional.

Yes for sure, technically it could of course be done by adding a layer like this:
```
Vehicle.X.Y.a
   - mandatory : true

Vehicle.X.Y.b
   - mandatory : false
```
and so on...

But I am not convinced we have a plan or need to do that and therefore I would prefer not using it as an example in the README, even if it would be possible.  I have moved the question of **whether this should be done or not** to #238 so we can keep that away from the more technical discussion here.

I will try to answer the Hertz example after some more thinking.

gunnarx:
@UlfBj wrote:
> I do not understand the above.

OK, I will look over and try to rewrite that section.

> I think the Reverse layer example should use the same metadata as used in the Standard layer example

OK, maybe we can have a simple example that is shown in both ways.  The reason it is not using the same now is that (as I see it) some use cases are more convenient with the Reverse format, and others are more convenient with the standard format.   Therefore I thought it would actually be more understandable to use different examples to explain why there should be two different methods.

I will try to simplify the explanation first.  After that I want to ask again if it is really still hard to understand the examples and concepts.

> The Reverse layer concept is only applicable for bullet 3 "Add new metadata", I believe?

That appears true, yes.  I haven't thought much about that, but I'm not sure if we have to point that out or if it is simply implied once the concept is understood.  

But we could also avoid special cases by allowing in theory, that the same thing is defined using both forward and reverse layers, and if so things could, again theoretically be redefined or removed also with reverse syntax.  (N.B. syntax for removal is not specified yet in either forward or reverse ways).

> I think the Tool support chapter should describe one design at a level that can be used for implementation, and then leave the door open for alternative designs/implementations. An implementation of the described design should then also be found at the repo.

First I'd say those requirements probably do not belong in a "README"  (Actually maybe none of this belongs in a README in the end - I expect it to become part of the combined documentation after we agree).   Then, I want to discuss a bit more because I think it is needed that we write down _some_ rules for how tools behave with layers, otherwise the layers concept is not clearly defined.  But we can definitely discuss the details more.

gunnarx:
@SebastianSchildt wrote:

> I feel this whole section is a lot of text, without saying much.

OK, I will look at it again.


(on reverse layers)
> I think this section "breaks" the model.

If I understand correctly, you don't see the need for the whole "Reverse layer" description?

Break the model?  Hmm... for me the "model" (it's perhaps a metamodel really, the underlying model behind the data descriptions) is  what it contains in the end, regardless of the input format.  There could be more than one way to express it in text and still be the same underlying "model" (in my very broad view).  But I do agree that we are also focusing on having _primarily_ one way of expressing (the standard VSS YAML) as far as we can.

Here I am proposing that some data relationships are better expressed in this "Reverse Layer" way (happy to get other name proposals) than the normal way.   Instead of "break",  we might see it maybe as actually an extension of the model we have, I'm not sure...   Consider then if extending could be allowed, since it is under the different name of "VSS layers", compared to the name "VSS"?

(continued)
> ...I think it is "legal", and "possible", because it is akin to just extending private data (i.e. creating new branches/subtrees) and filling whatever data you like.

I basically don't think of the example as any new branches (or new leaves) at all.  The way it is written it is basically defining a new concept (permissions), and then adding leaves (signals) to the new concept.  This could alternatively be seen in the standard/forwardd way as adding data to existing leaves (the leaves/signals that are referred to with their full paths).
So the same thing could be transformed into putting new meta-data on existing signals, yes.  But the idea is that some mappings are more convenient to define as A->B than with B->A.  

I think, just like Ulf's feedback, that the fact that I used a hierarchical names like vss.permissions.whatever here, might be a bit confusing though.  It is also possible that we should keep the ability I describe but find another way of explaining it without the "Reverse Layer" name.

(on deletion):
> I know we discussed this, but I am not super-sure we need it. After all, an implementation can still give "404" or "40x not implemented" on signals from the "core" spec. I think it would be a more clean design, than letting layers "delete" anything 
> 

Consider for example the "fat API" approach where you generate one static function for each item. Then there is no runtime 404, and you probably want to avoid generating for things that should not be there.  Even if that's not a common case, I don't want to put in limitations.  I think there could be other similar cases.  I just think a fundamental model like VSS must very flexible and we should not too much think about the only way it is used right now, today.   Also, "downstream" projects (users of VSS) could choose to add more restrictions (for example "deleting is not allowed"), if we start with a flexible definition at first.

> (i.e. I would rather provide a deployment/stack specific layer that tags "not-implemented: true" to relevant leaves, and letting the implementation act accordingly)

Sure, that's an alternative approach, with almost exactly the same effect.  You're using a different syntax to effectively delete it.  Actually we haven't proposed any syntax for deletion yet, so it's all open.

There are more details I think we might get into regarding possible ways to express these things more efficiently.  Sorry but not covered in the description yet but I think we may find that wildcards (xxx.*) could be useful in some layer definitions.  Let's see what we can come up with.

One key driver, is that while I can see things *could* be expressed in the standard way I'm concerned not with that a software program is able to read it in that alternative format -- it doesn't care about the format -- but that a human can easily write it.

vss_layers/README.md
adobekan: Remove is really specific, something like _omit signals_ could work if we talk about management of nodes for implementation phase.

vss_layers/README.md
```diff
@@ -0,0 +1,243 @@
...
+The layer concept definition should impose very little restrictions.
+Conventions and particular tools may introduce their own restrictions
+for certain usage, at a later time.
+
+The layer concept itself is able to do all of the following:
+
+1. Modify existing metadata and ultimately redefine any signal (i.e. change
+   its metadata) compared to a previous definition for example the definition
+   in the standard catalog.
```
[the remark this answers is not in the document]
adobekan: Modify/Change. Can we also expand optional metadata?

gunnarx:
Reverse Layers has been removed from description to simplify the discussion

gunnarx:
Latest consensus is being written into this Wiki document:
https://github.com/GENIVI/vehicle_signal_specification/wiki/VSS-Layers-Concept

Issue #140, opened 2020-02-16
Update Release process


danielwilms:
Release process is outdated and needs an update.

gunnarx:
My proposal for a working process is that there is no need to create an "Issue" if the PR can explain the problem directly.  Such as in this case, there is no waiting for discussion before the PR is created.  
So it just becomes like two duplicate things for the same thing?  I.e. only the PR would be enough for me if no longer discussion is expected.

danielwilms:
Now, I addressed it immediately, so in this case you're right. But in general I think it is as well a way to organise work and to have a backlog. Not necessarily for discussion only.

danielwilms:
Now, I addressed it immediately, so in this case you're right. But in general I think it is as well a way to organise work and to have a backlog. Not necessarily for discussion only.

PR #141, opened 2020-02-16
ReleaseProcess: Adapted to the current working mode

danielwilms:
Fixes: #140


RELEASE_PROCESS.md
```diff
@@ -8,33 +7,25 @@ provisions of the license provided by the LICENSE file in this repository.
 This document describes the process for creating a new version of the
 signal specification.
 
-The process, driven by the GENIVI Remote Vehicle Interaction Expert
-Group (RVI EG), is aimed at being lightweight and carried out in public, giving both
-GENIVI members and non-members a say in the creation of a new version.
+The process, driven by GENIVI, is aimed at being lightweight and carried
```
gunnarx: I might think to consider adding:
"...by GENIVI, with significant input from the W3C Automotive WG, is aimed at...
It's not the only stakeholder, and other GENIVI projects are increasingly working with this but lately a lot of drive has come from the discussions held in W3C context.

gunnarx: This is a comment for others to consider first and discuss.  My second comment is a "required" change from me.

danielwilms: @gunnarx: I thought about a similar thing at the beginning, but decided against it, as discussion is taking place in different places (e.g. layers discussion started in the android group, etc.). The important thing is, that it's open and the information is collected and linked here. But I'm not strongly against it.

Thanks for spotting the typo. I fixed it.

danielwilms:
> vss_release_process.png has a spelling error in the title. Please fix it first.

done.

gunnarx:
@danielwilms I think you should have amended the previous commit instead, i.e. combining them, which in effect would replace one picture with the other but in the first (only) commit.

As it is now, with two commits, when this is merged we are storing **two** binaries (the PNG) in the repository.  The faulty one does not really need to be stored at all.  Can you fix this and forcepush?  Or do you want some help to do that?

This branch is also out of date with master from what I can tell.  It ought to be automatically fixed with "Rebase and merge" this time, but then please try to have an up to date master branch for the next PR.

gunnarx:
It's easy so I'll just type it now, in case you did not know how:
```
git rebase --interactive HEAD~2
```
Then in the editor, change "pick" to "fixup" **on the second commit**.  This will combine (squash) the second commit into the first one.  Since the PNG file gets new content, it should just blow away the errored version from the commit and thus the history.

RELEASE_PROCESS.md
tguild: I welcome additional wording to reflect the regular coordination between GENIVI and W3C on this topic. As this is GENIVI's repo, the decision is theirs.

danielwilms:
@gunnarx, we could use as well the squash and rebase of github. but I can do the squasing as well on my branch. will be there in a minute.

danielwilms:
@gunnarx: done

gunnarx:
> @gunnarx, we could use as well the squash and rebase of github

Yes I suppose it would have the same effect.  I have never used that function actually.

RELEASE_PROCESS.md
danielwilms: @tguild: added it

PR #144, opened 2020-02-21
Replace tools/ with vss-tools.git submodule

gunnarx:
There are a few more issues to solve so for now I put this on a separate branch and open a PR.
@magnusfeuer @danielwilms 
The branch is stored in this same repo, so you can just checkout.


gunnarx:
You can still start developing vss-tools in here:  https://github.com/genivi/vss-tools

gunnarx:
OK, so I found that there are a lot of references to the **tools/** directory, which is now renamed **vss-tools**.  The path is mentioned in the Makefile, and in scripts like **vspec2c.py** and **vspec2ocf.py** but also in _many_ documentation files.  If we keep this directory name we have to change all those references.

We also need to restore the HUGO tool, according to @danielwilms because it is used during documentation generation (published on the GitHub-pages).

Before doing all of it, I'd like to understand more how the documentation is created:

It seems that some or all of the old **tools/documentation** should maybe not be in vss-tools, but rather in vehicle_signal_specification?  Since the old **tools/** is now "removed", they should be moved to a new path.   Unless we keep both names tools/ and vss-tools, but that's not the nicest maybe.

1. @danielwilms Which files/directories should we move from the old **tools/documentation** and still keep here in this VSS repo?
2. What should we call the directory?  tools/documentation seemed to be a mix of documentation files, and the actual tools to build the wiki (e.g. **hugo**).  Can we separate text files from doc tools?  If so, can the tools go into a directory called **documentation/tools** instead?  
3. What is the relationship between the text files under **tools/documentation** and those that are under **docs/**.  Are both needed?  Can we put them all in one place?

Also @magnusfeuer since you started this request I asked if you would agree to sign the commit for the split, in vss-tools and in vehicle_signal_specification.

magnusfeuer:
If you by signature you mean approval, then yes.
I approve this PR.

gunnarx:
I meant Signed-off-by: in the commit history.

danielwilms:
> @danielwilms Which files/directories should we move from the old tools/documentation and still keep here in this VSS repo?

Yes.

> What is the relationship between the text files under tools/documentation and those that are under docs/. Are both needed? Can we put them all in one place?

In the docs folder is the generated html code and it gets overwritten with the newly generated documentation. The files in the `tools/documentation` folder have to stay together, in `tools/documentation/content` you find the markdown files

> What should we call the directory? tools/documentation seemed to be a mix of documentation files, and the actual tools to build the wiki (e.g. hugo). Can we separate text files from doc tools? If so, can the tools go into the directory docs/tools instead?

@gunnarx, maybe we could move the `documentation` to a new folder called `utils`. Then we keep the tools folder clean.

gunnarx:
Updated PR.  This is now pointing the submodule to GENIVI instead of gunnarx.  The documentation generation was moved to a directory named docs-gen/ in the main repo, and *hopefully* everything is fixed.

gunnarx:
Study the [branch](https://github.com/GENIVI/vehicle_signal_specification/tree/vss-tools-submodule) for the final result.

gunnarx:
Unfortunately, all git history of the documentation was lost.  It does not look like a "git move" from documentation to docs-gen, because it was done after the repositories were split into two unrelated ones using subtree.  So now it looks like all the docs were re-added by me in a single commit.    This should have been possible to avoid (I think) but hard without redoing the whole split. Let me know if we can live with that.
The history of the documentation is of course available from the [git tree history before the split](https://github.com/GENIVI/vehicle_signal_specification/tree/f31feebdb4cdf9373ca0fe926d5e0590f1be49f1/tools/documentation).

gunnarx:
> The history of the documentation is of course available from the git tree history before the split.

This fact was now also reflected in the commit, see c4bff3e

danielwilms:
@gunnarx, I fixed the path in the makefile and committed that one. If you agree, I'd squash and merge.

gunnarx:
@danielwilms I fixed Makefie -> Makefile.
I think I prefer to keep the steps as I arranged them here and not squash them.  (Otherwise I typically do the rearranging before pushing the PR). Just in case there is any issue, I want to know what we did and when.  So no squash needed IMHO.  I see no reason to squash your Makefile fix either - it was an important step in the process.  Go ahead and merge, or I will :)

danielwilms:
Decided to merge 04/14

PR #152, opened 2020-04-13
Cleaned up inconsistent unit specification

Floix:
Found some inconsitencies in the unit specifications (e.g., unit: kilometer instead of unit: km). This PR cleans that up


danielwilms:
will be merged.

Issue #147, opened 2020-03-26
New EV signals


UlfBj:
Geotab has been asked by the W3C automotive group to propose what signals that could be relevant for the EV use case. This resulted in the document on the link below.
https://mail.google.com/mail/u/0?ui=2&ik=80e1234f77&attid=0.1&permmsgid=msg-a:r-4429506179679065946&th=170f9242229c39bf&view=att&disp=inline&realattid=f_k80hx7nn1

Below you will find a proposal for the VSS integration of the EV "signals" being discussed. 
There are some design questions still to answer, such as the resolution of the energy signals, and how to handle counter wrap-around.

- Delivered energy to the battery during charging (kWh).
- Consumed energy from the battery during "non-charging" (kWh).
-  Delivered energy to the battery from regenerative braking. (kWh).
- Battery state of charge. (%)

These must represent accumulated energy over time as I see it. That is, the values are monotonically increasing (until the value hits max possible, when it wraps around to zero).
What resolution do we believe is needed on these kWh values?
Should the number of "wrap-arounds" be stored by the vehicle?

The existing VSS tree contains the following:
Vehicle.Drivetrain.BatteryManagement.BatteryTemperature
Vehicle.Drivetrain.BatteryManagement.LowBatteryLevel
Vehicle.Drivetrain.BatteryManagement.ChargingInlet

I modified it slightly by adding one more branch -  Battery, which changes the existing entries to:

Vehicle.Drivetrain.BatteryManagement.Battery.Temperature
Vehicle.Drivetrain.BatteryManagement.Battery.LowLevel
Vehicle.Drivetrain.BatteryManagement.ChargingInlet

The proposal for the new EV signals below uses uint32 integers to hold the values, which have the unit Wh. Is the resolution Wh sufficient?
As it says in the presentation that a typical yearly load is about 4000kWh, then a uint32 would take about 1000 years before wrapping around.
For these vehicles there would be no need to keep track of an eventual wrap-around, but maybe for heavy trucks?
It would be sufficient with one wrap-around counter, as  ChargeEnergy and  ConsumedEnergy will not differ significantly. A uint8 should cover the needs.

- Vehicle.Drivetrain.BatteryManagement.Battery.ChargeEnergy:
  datatype: uint32
  type: sensor
  unit: Wh
  description: The accumulated energy delivered to the battery during charging.

- Vehicle.Drivetrain.BatteryManagement.Battery.ConsumedEnergy:
  datatype: uint32
[...]

PR #143, opened 2020-02-21
cnative tool refactoring.

UlfBj:

PR #154, opened 2020-04-16
Implementation of Issue GENIVI#147:New EV signals

UlfBj:


spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
+  datatype: float
+  type: sensor
+  unit: kWh
+  description: The accumulated energy delivered to the battery during charging.
+
+- Battery.AccumulatedConsumedEnergy:
+  datatype: float
+  type: sensor
+  unit: kWh
+  description: The accumulated energy leaving HV battery for propulsion and auxiliary loads.
```
gunnarx, line 46: Since when?  How and when is this signal reset to zero?

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
+
+- Battery.AccumulatedBrakingEnergy:
+  datatype: float
+  type: sensor
+  unit: kWh
+  description: The accumulated energy delivered to the battery from regenerative braking.
```
gunnarx, line 52: Since when?

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
+- Battery.LowLevel:
   datatype: Boolean
   type: sensor
   description: Indicates that the battery level is low
 
+- Battery.AccumulatedChargeEnergy:
+  datatype: float
+  type: sensor
+  unit: kWh
+  description: The accumulated energy delivered to the battery during charging.
```
gunnarx, line 40: ...since the most recent charging session started

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
 
-- BatteryCapacity:
-  datatype: int8
+- Battery.StateOfCharge:
+  datatype: uint8
   type: sensor
   unit: percent
   min: 0
   max: 100
-  description: Remaining capacity of the batter pack
```
gunnarx, line 25: Why was it removed? 

Also, it seems there should be a read-only attribute to indicate the nominal battery capacity of this particular car.

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
+  description: The current charging level as a percentage of original battery performance.
 
-- LowBatteryLevel:
+- Battery.LowLevel:
   datatype: Boolean
   type: sensor
   description: Indicates that the battery level is low
```
gunnarx, line 34: Agree, this is replaced by StateOfCharge

gunnarx:
The PR has conflicts.  It is sort of based on master before #144 was merged.
It conflicts because there are commits that modify cnative implementation (the same ones as in #143).
They are changing  tools/c_native/cnativenodelib.c and other files, but tools/ directory exists.  The directory is now named vss-tools and is a separate repo.  Therefore it conflicts.  

In particular I think the merge commit and the one you did before it: f032f53 did not quite achieve what is needed.

I fixed this locally by simply taking the current master and cherry-picking the only commit that actually matters here:  **38f4e8b  Implementation of Issue #147:New EV signals** 

and this worked fine.

@UlfBj If you want I can force push this fix back to the source branch for the PR, but be aware that this means changing your **Ulfbj/master** branch on GitHub unless you explicitly prevented updating the PR source branch.  Because **UlfBj/master** is where the PR was sent from, as we can see above.  (But maybe that's actually not a bad idea - you will be back in sync).  Let me know.

gunnarx:
It looks like this: https://github.com/gunnarx/vehicle_signal_specification/tree/pr_154_fixed

UlfBj:
>> @UlfBj If you want I can force push this fix back to the source branch for the PR, but be aware that this means changing your Ulfbj/master branch on GitHub unless you explicitly prevented updating the PR source branch. Because UlfBj/master is where the PR was sent from, as we can see above. (But maybe that's actually not a bad idea - you will be back in sync). Let me know.

Sounds like a good idea, so please go ahead.

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
 # Battery management system for EV / Hybrids
 #
 
-- BatteryTemperature:
+- Battery:
```
alexanderdomin, line 13: I have a proposal to do several things:

1. Move BatteryManagement from Drivetrain to Powertrain branch. Due to Wikipedia Drivetrain is the wrong location for this leafs.  The powertrain term includes all of the components used to transform stored (chemical, solar, nuclear, kinetic, potential, etc.) energy into kinetic energy for propulsion purposes. Drivetrain is a group of components that deliver power to the driving wheels.

2. In a car we have different Batteries.
    a. Automotive battery -->  Its main purpose is to feed the starter, which starts the engine.
    b. Electric vehicle battery --> Power the electric motor.

So I think we need also differentiation also hear. I am not native speaker in english but I would like to propose following:

- Vehicle.Powertrain.Storage.HighVoltage --> for Electric vehicle battery
- Vehicle.Powertrain.Storage.AutomobileBattery --> automobile battery

or

- Vehicle.Powertrain.Battery.ElectricVehicle --> for Electric vehicle battery
- Vehicle.Powertrain.Battery.Automobile --> automobile battery

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
+
+- Battery.Temperature:
   datatype: float
   type: sensor
   unit: celsius
   description: Temperature of the battery pack
 
-- BatteryCapacity:
-  datatype: int8
+- Battery.StateOfCharge:
```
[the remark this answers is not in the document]
gunnarx: There could be "Voltage" which is the absolute value and is allowed to be the actual voltage, regardless of what the normal voltage is of the battery (different for different cars).  In fact, adding a fixed attribute specifying the nominal voltage for the installed battery type also makes sense to me.
and "State of Charge" which can be in percent (and intended to be shown to a customer)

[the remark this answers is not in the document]
alexanderdomin: I think we can avoid this issue adding the "rated capacity" to the tree. In this case you will be able to show to the customer informations he needs.

spec/Drivetrain/BatteryManagement.vspec
gunnarx: Very good point, Alexander.  I think this PR might have by mistake removed the normal car battery because everyone is so focused on EVs?

> In a car we have different Batteries.
> a. Automotive battery --> Its main purpose is to feed the starter, which starts the engine.

In fact, often also called the Starter Battery I think?  I'm not sure why the other battery would not be "automotive"?

> or
> Vehicle.Powertrain.Battery.ElectricVehicle --> for Electric vehicle battery
> Vehicle.Powertrain.Battery.Automobile --> automobile battery

or even the opposite, in my opinion.  The branch named ElectricVehicle should be higher up since the whole branch can be avoided completely if it is not an EV:

Vehicle.Powertrain.Battery.**ElectricVehicle** --> Vehicle.Powertrain.**ElectricVehicle**.Battery  ?

gunnarx:
As indicated above, I resolved the conflicts and pushed a new clean version to the PR branch (UlfBj/master)

spec/Drivetrain/BatteryManagement.vspec
danielwilms: EV Battery under powertrain -> see comment @gunnarx.
Support battery in a different issue, as not only EV.

danielwilms: vehicle.powertrain.electric.battery

spec/Drivetrain/BatteryManagement.vspec
danielwilms: `lifetime of battery` in description
Rename to `AccumulatedChargedEnergy`

spec/Drivetrain/BatteryManagement.vspec
danielwilms: directly under powertrain.

spec/Drivetrain/BatteryManagement.vspec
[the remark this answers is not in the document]
gunnarx: @alexanderdomin wrote:

> I think we can avoid this issue adding the "rated capacity"

I don't fully understand. Avoid which issue, and are you proposing to add rated capacity, or to not add it?  It sounds like a independent of what is discussed here.

- StateOfCharge (% full)
- is different from the current measured Voltage
- and that is different from rated capacity (which could also be called Nominal Capacity, I think)

spec/Drivetrain/BatteryManagement.vspec
```diff
@@ -10,25 +10,47 @@
...
+  datatype: uint8
   type: sensor
   unit: percent
   min: 0
   max: 100
-  description: Remaining capacity of the batter pack
+  description: The current charging level as a percentage of original battery performance.
```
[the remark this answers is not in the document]
gunnarx: Seems OK for me.  State of charge would say how much the battery is charged compared to the currently usable capacity and would then always be at or near 100% after a full charge.  But there must then also be another item that states the actual amount of stored energy, so that there is some absolute value to rely upon.

spec/Drivetrain/BatteryManagement.vspec
gunnarx: I believe this could be used to estimate how "worn out" a battery is likely to be.  (together with AccumulatedChargeEnergy).  It is similar to odometer, the "complete distance travelled" for evaluating how worn a vehicle is, but more accurate for the battery itself.

spec/Drivetrain/BatteryManagement.vspec
gunnarx: also lifetime of battery, should be in description.

danielwilms:
@UlfBj, will create new proposal.

spec/Drivetrain/BatteryManagement.vspec
gunnarx: I must have misunderstood the change.  I thought it was removed.  I agree with @alexanderdomin that this is not needed - as I wrote "StateOfCharge".

danielwilms:
@danielwilms: rebase it after PR merge

spec/Drivetrain/BatteryManagement.vspec
danielwilms: I would conclude this one.

alexanderdomin: I still think this leaf is statistics relevant. So I think it is better to introduce following leaf: Battery.Statistics.AccumulatedConsumedEnergy

I also think we need tree structure which is generic and can be instantiated for the life time, from last charging or from date x, 

Please also add a description seans it is for the lifetime of the vehicle.

spec/Drivetrain/BatteryManagement.vspec
danielwilms: removed leaf

spec/Drivetrain/BatteryManagement.vspec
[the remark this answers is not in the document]
danielwilms: Added gross/net capacity and nominal/referent voltage. I hope this concludes the discussion

spec/Drivetrain/BatteryManagement.vspec
danielwilms: Done as discussed.

danielwilms:
I rebased @UlfBj s commit,  added a commit based on the comments you gave. Please check, if we can merge it like that @alexanderdomin, @gunnarx, @Floix, @UlfBj

UlfBj:
I am ok with the PR updates.

spec/Body/Body.vspec
```diff
@@ -146,3 +146,18 @@
   description: All mirrors
 # Include mirrors specification and attach it to the Mirrors branch
 #include ExteriorMirrors.vspec Mirrors
+
+##
+# Charging Inlet definition
+##
+
+- ChargingInlet:
```
danielwilms, line 154: * Bi-Directional. e.g. ChargingPort?

Issue #160, opened 2020-06-09
charging inlet


danielwilms:
ChargingInlet is leading to the idea, that it's unidirectional. That might not be true for all ports. So, renaming would be needed.

@SebastianSchildt argued for ChargingPort.

PR #161, opened 2020-06-10
Changing ChargingInlet > ChargingPort

danielwilms:
ChargingInlet is leading to the idea, that it's unidirectional.
That might not be true for all ports. So, renaming would be needed.

@SebastianSchildt

Fixes: #160

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -151,13 +151,13 @@
 # Charging Inlet definition
 ##
 
-- ChargingInlet:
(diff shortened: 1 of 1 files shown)


gunnarx:
I'm OK with this.  In general for these issues I promote that the most commonly used name in the industry should be preferred (even more than it being perfectly accurate in terms of behavior).  
But I expect ChargingPort to be common enough, and I don't know what is the most popular name, so it seems OK to me.

PR #162, opened 2020-06-10
Describing Data Streams

danielwilms:
Extended the documentation about the usage of data streams.

Fixes: 112

### DIFF
--- a/docs-gen/content/rule_set/data_entry/data_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_types.md
@@ -41,3 +41,39 @@ datatype: UInt32[100]
 
 Examples for the usage of `arrays` are status information about battery cells or a list of DTCs, which are present in a
 vehicle.
+
+
+
+## Data Streams
+
+Data Entries, which describe sensors offering binary streams
+(e.g. cameras), are not supported directly by VSS with a
+dedicated data type. Instead, they are described through the
+meta data about the sensor itself and how to retrieve the
+corresponding data stream.
+
+A camera can be a good example to it. The Data Entry for the camera
+and the corresponding video stream could look like:
+
+```YAML
+- Camera:
+  type: branch
+  description: Information about the camera and how to connect to the video stream
+
+- Camera.IsActive:
+  type: actuator
+  datatype: boolean
+  description: If the camera is active, the client is able to retrieve the video stream
+
+- Camera.URI:
+  type: sensor
+  datatype: string
+  description: URI for retrieving the video stream
+```
+
+In this example, it shows the usage of meta data about
+the status of the sensor. The camera can be set to active through
+the same data entry (`actuator`). A dynamic data entry (`sensor`)
+is used for the URI of the video stream. As long as the camera
+`IsActive`, in this example the `URI` is set and empty otherwise.
+Datatypes are not specified for ```branch```
--- a/docs-gen/content/rule_set/data_entry/data_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_types.md
@@ -68,17 +68,12 @@ and the corresponding video stream could look like:
 - Camera.URI:
   type: sensor
   datatype: string
-  description: URI for retrieving the video stream
+  description: URI for retrieving the video stream, with information on how to access the stream (e.g. protocol)
 
-- Camera.Protocol:
-  type: attribute
-  datatype: string
-  description: Protocol used for the video stream, connection done via Camera.URI
 ```
 
 In this example, it shows the usage of meta data about
 the status of the sensor. The camera can be set to active through
 the same data entry (`actuator`). A dynamic data entry (`sensor`)
-is used for the URI of the video stream. As long as the camera
-`IsActive`, in this example the `URI` is set and empty otherwise.
-Datatypes are not specified for ```branch```
+is used for the URI of the video stream. Information on how to access
+the stream is expected.
--- a/docs-gen/content/rule_set/data_entry/data_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_types.md
@@ -68,7 +68,7 @@ and the corresponding video stream could look like:
 - Camera.URI:
   type: sensor
   datatype: string
-  description: URI for retrieving the video stream, with information on how to access the stream (e.g. protocol)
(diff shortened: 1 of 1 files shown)


danielwilms:
> In the example IsActive is an actuator. I would have thought it more often would be a sensor, i. e. it is the vehicle control system that sets it to on/off. But as an example, either way is ok.

@UlfBj Thanks for the review! In the example I wanted to show that you could request to activate the camera from client side to extend it a bit.

docs-gen/content/rule_set/data_entry/data_types.md
```diff
@@ -23,3 +23,38 @@ String     | character string           | n/a  | n/a
...
+
+## Data Streams
+
+Data Entries, which describe sensors offering binary streams
+(e.g. cameras), are not supported directly by VSS with a
+dedicated data type. Instead, they are described through the
+meta data about the sensor itself and how to retrieve the
+corresponding data stream.
+
+A camera can be a good example to it. The Data Entry for the camera
```
gunnarx, line 37: example of, not example to.

docs-gen/content/rule_set/data_entry/data_types.md
```diff
@@ -23,3 +23,38 @@ String     | character string           | n/a  | n/a
...
+
+- Camera.IsActive:
+  type: actuator
+  datatype: boolean
+  description: If the camera is active, the client is able to retrieve the video stream
+
+- Camera.URI:
+  type: sensor
+  datatype: string
+  description: URI for retrieving the video stream
```
gunnarx, line 53: The description should give information about data format / streaming protocol.

docs-gen/content/rule_set/data_entry/data_types.md
```diff
@@ -23,3 +23,38 @@ String     | character string           | n/a  | n/a
...
+```
+
+In this example, it shows the usage of meta data about
+the status of the sensor. The camera can be set to active through
+the same data entry (`actuator`). A dynamic data entry (`sensor`)
+is used for the URI of the video stream. As long as the camera
+`IsActive`, in this example the `URI` is set and empty otherwise.
```
gunnarx, line 60: > As long as the camera `IsActive`, in this example the `URI` is set and empty otherwise.

I heard this in the previous call but I did not have time or reason to interject.  I do not think it's required to set the URL to empty if the IsActive is there to give this information?  In any case, if this **is** the behavior of some particular stream, it ought to be in the description I think.  I would not write it here as if it is a required behavior for all streams.

danielwilms: >  I heard this in the previous call but I did not have time or reason to interject. I do not think it's required to set the URL to empty if the IsActive is there to give this information? In any case, if this is the behavior of some particular stream, it ought to be in the description I think. I would not write it here as if it is a required behavior for all streams.

Fair enough. It just adds complexity to the example and definitely it's not mandatory. Will remove that one.

docs-gen/content/rule_set/data_entry/data_types.md
danielwilms: This should be in a different node then. My thinking was mostly, that the protocol is part of the URI. 

But we could make it explicit with a node, like
```YAML
- camera.Protocol
   type: attribute
   datatype: string
   description: Protocol used for the video stream, connection done via Camera.URI
```

In private discussions that came up already once before, so I'll add it to the PR.

docs-gen/content/rule_set/data_entry/data_types.md
danielwilms: done

docs-gen/content/rule_set/data_entry/data_types.md
gunnarx: > This should be in a different node then

Why?  My proposal was only that the description should explain everything you need to know to use the data item (in this case the URL).   It is generally the case that the description needs to give enough information.  In this case you have to explain what is actually being transferred.

I don't think Protocol needs to be a data item in VSS.  It would only need to be a separate item if it is something that can change dynamically, and that is usually not the case.

If "camera" was part of the standard VSS catalog, we might consider being a bit more formal.  But this is just a guideline for how anyone can define streaming URLs, for any purpose.  In my opinion there is no reason to have a _guideline_ to put Protocol in a separate attribute for every definition of a streaming URL.

gunnarx: > My thinking was mostly, that the protocol is part of the URI.

On some protocol level it is, of course.  But in reality behind **http://something** lies a lot of different data formats typically, even sometimes some embedded "protocols". 
 In particular, "the video stream" that was mentioned in the comment is not enough to define it clearly.   It might not always make sense to describe these things, but in some cases it should and the video stream was a particular case, in my opinion.

danielwilms: @danielwilms: add description, remove protocol node for simplicity

docs-gen/content/rule_set/data_entry/data_types.md
```diff
@@ -50,16 +50,12 @@ and the corresponding video stream could look like:
 - Camera.URI:
   type: sensor
   datatype: string
-  description: URI for retrieving the video stream
+  description: URI for retrieving the video stream, with information on how to access the stream (e.g. protocol)
```
gunnarx, line 53: Sorry to propose a change again but this documentation is a really important part to be able to say we have "solved" the streaming data issue.

What do you think about this for even more clarity:

"...information on how to interpret the stream (e.g. protocol, data format, encoding, etc.)"

gunnarx: ^^ because I agree with your fundamental point that the main protocol is already in the URI itself.  Only some "tunneled" protocol is valid, and therefore it is good to mentioned here, but there could be also things like video encoding parameters and so on.

danielwilms: > Sorry to propose a change again but this documentation is a really important part to be able to say we have "solved" the streaming data issue.

I agree and sometimes things are only noticeable, when you see it in writing. @gunnarx, added the description and rebased. Will squash it during merge into master.

PR #164, opened 2020-06-17
Instance documentation

danielwilms:
Fixes: #81


docs-gen/content/rule_set/instances.md
```diff
@@ -0,0 +1,86 @@
+---
+title: "Instances"
+date: 2019-07-31T15:27:36+02:00
+weight: 5
+---
+
+As VSS resembles primarily the physical structure of the vehicle. So
```
gunnarx, line 7: VSS resembles primarily the physical structure of the vehicle, so ...

docs-gen/content/rule_set/instances.md
```diff
@@ -0,0 +1,86 @@
...
+2. The instantiation is done for every data entry in the following path.
+3. Instances are defined with the key-word `instances`, followed by its
+   definition, which can be either:
+   * a list of strings, where each element defines a single instance, e.g.
+     `['Left','Right']` results into two instances of every following
+     data entry in the path, named `Left` and `Right`
+   * a string, followed by a range, which defines the number of instances:
+     `Position[1,4]` results into 4 instances of every following
+     data entry in the path, named `Position1`, `Position2`, `Position3`
+     and `Position4`.
```
gunnarx, line 38: I assume ranges must start at 1, or could they start at some other number?  It ought to be specified.
Also, I thought we had discussed the syntax [1..4] instead of [1,4]?
The [1..4] makes the range syntax more clearly different from ['a',b']

docs-gen/content/rule_set/instances.md
```diff
@@ -0,0 +1,86 @@
...
+4. If multiple instances occur in one node or on the path to a data entry,
+   the instances get combined, by the order of occurrence. Following the example above, for each instance `Left` and `Right`, 4 position
+   instances shall be created, addressed by the dot-notation.
+
```
gunnarx, line 42: s/shall/will/?


Also the sentence flows more naturally if you flip it around:

Four position instances will be created for each of the 'Left' and 'Right' instances.

instead of 

For each of the Left and Right instances, 4 instances will be created.

also I would end with "... a total of 8 instances" (for clarity).

To shorten the text - I don't think "...addressed by the dot-notation." immediately adds any understanding.  It is either obvious (because you always address by dot notation), or it will be hard to understand _exactly_ what it means (and in that case an example is needed in any case)

docs-gen/content/rule_set/instances.md
```diff
@@ -0,0 +1,86 @@
...
+Vehicle.Cabin.Door.IsOpen.Row2
+Vehicle.Cabin.Door.IsOpen.Row2.Left
+Vehicle.Cabin.Door.IsOpen.Row2.Right
+Vehicle.Cabin.Door.IsOpen.Row3
+Vehicle.Cabin.Door.IsOpen.Row3.Left
+Vehicle.Cabin.Door.IsOpen.Row3.Right
+Vehicle.Cabin.Door.IsOpen.Row4
+Vehicle.Cabin.Door.IsOpen.Row4.Left
+Vehicle.Cabin.Door.IsOpen.Row4.Right
+```
```
gunnarx, line 86: As noted many times before, I still think it should be instantiated in the order it is logically written in the tree.  In other words:

```
Vehicle.Cabin.Door.Row1.Left.IsOpen
```
I guess I have still not understood the point of modifying this order...

gunnarx: To add to the above,  a more logical tree in the VSS seems to be to instantiate Row node, and the Left & Right Door being defined below the Row branch.  But according to the definition you gave above, the instantiation would become as written above (I think)

SebastianSchildt: > As noted many times before, I still think it should be instantiated in the hierarchical order it is was defined in the original files. In other words:
> 
> ```
> Vehicle.Cabin.Door.Row4.Right.IsOpen
> ```
> 
> I guess I have still not understood the point of modifying this order...
The order you suggested order also "feels" more natural to me, however I am not sure how mich that is a "core" VSS question and how much it is a question of just how we generate outputs (CSV, JSON), or if it even just a protocol question. (the YAML as discussed here would be the same in any case)

danielwilms: @SebastianSchildt: create JSON example

danielwilms: @danielwilms: create an example for better understanding

SebastianSchildt:
As requested, a walkthrough using fine hand-crafted JSON, maybe it helps in discussion.

For starters I assume that we want instances, and the YAML notation in master is fine and expresses what we want. 

It looks like this

```yaml
- Door:
  type: branch
  instances:
    - Row[1,4]
    - ["Left","Right"]
  description: All doors, including windows and switches
#include SingleDoor.vspec Door
```

Currently the tooling generates this JSON

```json
"Door": {
            "type": "branch",
            "instances": [
              "Row[1,4]",
              [
                "Left",
                "Right"
              ]
            ],
            "description": "All doors, including windows and switches",
            "uuid": "fd7f4d16f8965419a9a69fd66b40c1d7",
            "children": {
              "IsOpen": {
                "datatype": "boolean",
                "type": "actuator",
                "description": "Is door open or closed",
                "uuid": "5a04d833cf2e55e3a3ceb1f1e525a188"
              },
```
Screenshot 2020-07-07 at 13 41 11

I think that is almost fine, as far as structure is concerned: It like a single Door, except the "instances" describe that there are more, well, door instances. I think instances themselves are coded in a rubbish way though

```json
"instances": [
              "Row[1,4]",
              [
                "Left",
                "Right"
              ]
        ],
```

This is "unpareseable" or it least complicates parsing a lot, and it is not at all clear what this means.  I think the intention might be something like

```json
"instances": {
               "Row1": ["Left","Left"] ,
               "Row2": ["Left","Right"] ,
               "Row3": ["Left","Left"] ,
              "Row4": ["Left","Right"] ,
        },
```

Screenshot 2020-07-07 at 14 06 32

I think this is the minimal way of "unrolling" is needed in JSON export: When using a JSON parser tooling should have the "structure" of instances accessible without needing to interpret/parse some "inner" format hidden in the data.


However, we do need to define (for the YAML), how powerful our instance construct should be. In order of ascending implementation pain/bug density

 - Allow a simple one-dimensional list of instances : This would simplify instances to a list, but int he example it means ending up with `[ "row1_left", "row1_right", "row2_left",......   ]`
 - Allowing multiple dimension (limited to 1,2,n or arbitrary -> We need to define this), and they can either be 
      - regularly shaped (won't capture different amount of tyres per axle)
      - irregularly shaped -> Most powerful
    
I think the only valid options are: "Just a list" (simple, VSS like), or go all out. In the JSPN example everything works, in the complex case: Dictionaries/Maps until the last level, then lists. However, just "encoding" it is easy, the point is tooling _using_ the data model would need to be able to deal with the complexity.

I think these are the strictly VSS related questions. 

To be continued in next comment....

SebastianSchildt:
Welcome to instantiation two of our nice JSON show!

In all cases discussed above implementations/users of VSS need to understand the instance concept, and do some internal management/instantiation of data structures. 

I think it is also a valid point to view instances as purely  "Syntactic sugar" in terms of YAML only and flatten this using existing VSS concepts. Therefore,  I would (probably as an alternate mode) llke the tooling to generate the instances like this:

```json
         "Door": {
            "type": "branch",
            "children": {
                "Row1": {
                    "type": "branch",
                    "children" :{ 
                        "Left": {
                            "IsOpen": {
                                "datatype": "boolean",
                                "type": "actuator",
                                "description": "Is door open or closed",
                                "uuid": "5a04d833cf2e55e3a3ceb1f1e525a188"
                              },
                              "IsLocked": {
                                "datatype": "boolean",
                                "type": "actuator",
                                "description": "Is door locked or unlocked. True = Locked. False = Unlocked.",
                                "uuid": "6cc24552c4185026a1ead2e2aee24b1b"
                              },
                              "Window": {
                                "type": "branch",
                                "description": "Door window status",
                                "uuid": "0295961e129e5706ac46da25daf2e142",
                                "children": {
                                  "isOpen": {
                                    "datatype": "boolean",
                                    "type": "sensor",
                                    "description": "Is door open or closed",
                                    "uuid": "f760a77693455d41b43c8d1ffc9261d7"
                                  },
                                  "Position": {
                                    "datatype": "uint8",
                                    "type": "sensor",
                                    "min": 0,
                                    "max": 100,
                                    "unit": "percent",
                                    "description": "Window position. 0 = Fully closed 100 = Fully opened.",
                                    "uuid": "c861bfaaaf795ab28597f684427277ba"
                                  },
                                  "ChildLock": {
                                    "datatype": "boolean",
                                    "type": "sensor",
                                    "description": "Is door child lock engaged. True = Engaged. False = Disengaged.",
                                    "uuid": "76175d9563705cc791e8eb6ba61861cc"
                                  },
                                  "Switch": {
                                    "datatype": "string",
                                    "type": "actuator",
                                    "enum": [
                                      "Inactive",
                                      "Close",
                                      "Open",
                                      "OneShotClose",
                                      "OneShotOpen"
                                    ],
                                    "description": "Switch controlling sliding action such as window, sunroof, or blind.",
                                    "uuid": "e729ded092f055f0b813ccc37d49fc77"
                                  }
                                }
                              },
                              "IsChildLockActive": {
                                "datatype": "boolean",
                                "type": "sensor",
                                "description": "Is door child lock engaged. True = Engaged. False = Disengaged.",
                                "uuid": "0246b897c0325faf869330d2e96b36af"
                              },
                              "Shade": {
                                "type": "branch",
                                "description": "Side window shade",
                                "uuid": "2e3b255557e15f81b0169d7133f8f5af",
                                "children": {
                                  "Switch": {
                                    "datatype": "string",
                                    "type": "actuator",
                                    "enum": [
                                      "Inactive",
                                      "Close",
                                      "Open",
                                      "OneShotClose",
                                      "OneShotOpen"
                                    ],
                                    "description": "Switch controlling sliding action such as window, sunroof, or blind.",
                                    "uuid": "e842a21dc81358d1b9df58e10542f002"
                                  },
                                  "Position": {
                                    "datatype": "uint8",
                                    "type": "actuator",
                                    "min": 0,
                                    "max": 100,
                                    "unit": "percent",
                                    "description": "Position of side window blind. 0 = Fully retracted. 100 = Fully deployed.",
                                    "uuid": "abee29324b4658db8146b3ca604c5461"
                                  }
                                }
                              }
                        },
                        "Right": {
                            "IsOpen": {
                                "datatype": "boolean",
                                "type": "actuator",
                                "description": "Is door open or closed",
                                "uuid": "5a04d833cf2e55e3a3ceb1f1e525a188"
                              },
                              "IsLocked": {
                                "datatype": "boolean",
                                "type": "actuator",
                                "description": "Is door locked or unlocked. True = Locked. False = Unlocked.",
                                "uuid": "6cc24552c4185026a1ead2e2aee24b1b"
                              },
                              "Window": {
                                "type": "branch",
                                "description": "Door window status",
                                "uuid": "0295961e129e5706ac46da25daf2e142",
                                "children": {
                                  "isOpen": {
                                    "datatype": "boolean",
                                    "type": "sensor",
                                    "description": "Is door open or closed",
                                    "uuid": "f760a77693455d41b43c8d1ffc9261d7"
                                  },
                                  "Position": {
                                    "datatype": "uint8",
                                    "type": "sensor",
                                    "min": 0,
                                    "max": 100,
                                    "unit": "percent",
                                    "description": "Window position. 0 = Fully closed 100 = Fully opened.",
                                    "uuid": "c861bfaaaf795ab28597f684427277ba"
                                  },
                                  "ChildLock": {
                                    "datatype": "boolean",
                                    "type": "sensor",
                                    "description": "Is door child lock engaged. True = Engaged. False = Disengaged.",
                                    "uuid": "76175d9563705cc791e8eb6ba61861cc"
                                  },
                                  "Switch": {
                                    "datatype": "string",
                                    "type": "actuator",
                                    "enum": [
                                      "Inactive",
                                      "Close",
                                      "Open",
                                      "OneShotClose",
                                      "OneShotOpen"
                                    ],
                                    "description": "Switch controlling sliding action such as window, sunroof, or blind.",
                                    "uuid": "e729ded092f055f0b813ccc37d49fc77"
                                  }
                                }
                              },
                              "IsChildLockActive": {
                                "datatype": "boolean",
                                "type": "sensor",
                                "description": "Is door child lock engaged. True = Engaged. False = Disengaged.",
                                "uuid": "0246b897c0325faf869330d2e96b36af"
                              },
                              "Shade": {
                                "type": "branch",
                                "description": "Side window shade",
                                "uuid": "2e3b255557e15f81b0169d7133f8f5af",
                                "children": {
                                  "Switch": {
                                    "datatype": "string",
                                    "type": "actuator",
                                    "enum": [
                                      "Inactive",
                                      "Close",
                                      "Open",
                                      "OneShotClose",
                                      "OneShotOpen"
                                    ],
                                    "description": "Switch controlling sliding action such as window, sunroof, or blind.",
                                    "uuid": "e842a21dc81358d1b9df58e10542f002"
                                  },
                                  "Position": {
                                    "datatype": "uint8",
                                    "type": "actuator",
                                    "min": 0,
                                    "max": 100,
                                    "unit": "percent",
                                    "description": "Position of side window blind. 0 = Fully retracted. 100 = Fully deployed.",
                                    "uuid": "abee29324b4658db8146b3ca604c5461"
                                  }
                                }
                              }
                        }
                    }
                },
                "Row2": {
                    "type": "branch",
                    "children" :{ 
                        "Left": {
                            "IsOpen": {
...
```

Screenshot 2020-07-07 at 14 14 11

This would be 100% "backward" compatible, consequently, when subscribing to a value using W3C web socket protocol you would go for

`Vehicle.Cabin.Door.Row1.Left.isOpen`

whereas in an "instance-aware" server you might go for

`Vehicle.Cabin.Door[Row1][Left].isOpen`

However, I think this is not "Vehicle signal specification"-relevant. If somebody likes to put instances last in a specific protocol due to  some platform limitations or conventions: More power to him.

gunnarx:
Conversations like these are always hard when we speak different languages, and come at the problem from very different angles.  I'll try to write down my thoughts based on Sebastian's answer:

1. Why use JSON as the language to understand this?
 - A flat list of fully qualified VSS paths tell us everything we need to know, in my opinion?
 - We usually say "CSV" for this although only the first column of the comma-separated-values is relevant here.

2. What is even the purpose of our YAML to JSON generator in the first place (general question).  I have not any use for it yet, but maybe we can define some useful reason?

3. The example **itself** is unfortunately not perfectly crafted, for my view of the world:
```
- Door:
  type: branch
  instances:
    - Row[1,4]
    - ["Left","Right"]
  description: All doors, including windows and switches
#include SingleDoor.vspec Door
```

... because it is more logical that Rows include Doors that include Left and Right Door.
And I'm not talking about a different way of "translating this".  I am talking about the definition itself which here puts Rows under Doors and is therefore (in my world view) a bad example, but maybe makes sense for other views.
So we can talk about the mechanics of how to translate this example as it is written (which I did in my reply) even if the example is unfortunately misleading then and probably leads to confusion.  


```
"instances": [
              "Row[1,4]",
              [
                "Left",
                "Right"
              ]
        ],
```

This might be unparseable but again it is about the JSON translation, so I don't know why that matters.  Let's get the flat list / CSV done first?

> I think it is also a valid point to view instances as purely "Syntactic sugar" in terms of YAML only 
> and flatten this using existing VSS concepts.

Yes, I think that is a fair view.

>  Therefore, I would (probably as an alternate mode) llke the tooling to generate the instances like this:

I could not follow the JSON perfectly but a first view looks like the same translation I did above, basically the instances appear just like they are defined...

gunnarx:
> Allow a simple one-dimensional list of instances : This would simplify instances to a list, but int he example it means ending up with [ "row1_left", "row1_right", "row2_left",...... ]

No, the translation should create a VSS hierarchy...
Row1.Left and Row1.Right and so on..

```
"instances": {
               "Row1": ["Left","Left"] ,
               "Row2": ["Left","Right"] ,
               "Row3": ["Left","Left"] ,
              "Row4": ["Left","Right"] ,
        },
```
I will not right now discuss if the above is the right syntax because in my opinion the core question is if "instances" are a concept that should remain in the _result_ tree or if it gets expanded to a plain VSS tree.  If it gets expanded as I see it, then there is no special syntax appearing in the JSON (like the one you describe here), so there is no discussion if it is the right syntax.

danielwilms:
@gunnarx, @SebastianSchildt, @UlfBj: Please review, if the new documentation suits your idea of instances. In parallel I'll work on the implementation on the tools side.

docs-gen/content/rule_set/instances.md
danielwilms: done

docs-gen/content/rule_set/instances.md
danielwilms: done.

docs-gen/content/rule_set/instances.md
danielwilms: done.

UlfBj (approved):

SebastianSchildt (approved):
Current doc is fine for me, So I vote merge.

I still strongly suggest that we should move the task of clearly separating between "model" and "deployment" in VSS and our tooling forward

danielwilms:
meeting 22/09/20

Issue #167, opened 2020-07-14
Merge and structure Powertrain and Drivetrain branches


dakrbo:
Hello everyone, as pre-sounded in Slack, I'd like to table a proposal:

In the current state of the VSS tree, there exists a somewhat incoherent split between Drivetrain and Powertrain,  where Powertrain contains some EV-Battery stuff (energy storage), whereas EV Motor (power source) is located in Drivetrain. At the same time, fuel system (energy storage) and combustion enginer (power source) are both located in Drivetrain. There currently is no structural distinction between energy storage and power source.

I would like to propose some structural changes to address these two topis:
1. merge Powertrain and Drivetrain branches
2. introduce structure elements to distinguish energy storage from power sources

Note: I suggest the get rid of 'Drivetrain' as a top-level element, and use 'Powertrain' as the top-level branch name. Rationale: [Drivetrain](https://en.wikipedia.org/wiki/Drivetrain) is a subset of [Powertrain](https://en.wikipedia.org/wiki/Powertrain) (power delivery vs power generation and delivery), as per the corresponding wikipedia definitions.

So I would propose the following changes:
- create Powertrain.EnergyStorage
- create Powertrain.PowerSource
- move Drivetrain.ElectricMotor to Powertrain.PowerSource.ElectricMotor
- move Drivetrain.Engine to Powertrain.Powersource.CombustionEngine
- move Drivetrain.FuelCell to Powertrain.EnergyStorage.FuelCell
- move Drivetrain.FuelSystem to Powertrain.EnergyStorage.FuelSystem 
- move Drivetrain.Transmission to Powertrain.Transmission 
- move Powertrain.Electric to Powertrain.EnergyStorage.Battery
- get rid of Drivetrain 

Note: I'm not sure what to do with Powertrain.Powertrain, and whether Powertrain.Transmission is what we want. So in addition, we might do something like:
- create Powertrain.Drivetrain
- move .Transmission to Powertrain.Drivetrain.Transmission
- move .Powertrain to Powertrain.Drivetrain.Electric 

That's it for the restructuring proposal - let me know what you think!

klotzbenjamin:
This is interesting, especially since there is some overlap on electric motor branches.

Just for understanding, the taxonomy would look like this, right?

- Powertrain
   - EnergyStorage
      - FuelCells
      - FuelSystem
      - Battery
   - PowerSource
      - ElectricMotor
      - CombustionEngine
   - Transmission
   - Drivetrain
[...]

PR #169, opened 2020-07-22
#167 merged Drivetrain and Powertrain trees, updated vspec include directives

dakrbo:
Pull request to address issue #167


SebastianSchildt:
Ran it through tooling and seems to work for me.

This structure is an improvement to the status-quo, and if it works for others too, I would suggest merging it and discussing the potential renaming/moving of specific datapoints within this hierarchy separately. 

👍

danielwilms:
Fine for me. I'm always a bit reluctant when changing the structure only, as there are always different opinions on where a leaf belongs. But cleaning up the drivetrain/powertrain structure is a real improvement I'd say.

PR #176, opened 2020-08-14
Occupant driver

chheis:
Proposal for Driver/Occupant branches, to enable the indentification of the driver and other occupants


spec/Occupant/Identifier/Identifier.vspec
```diff
@@ -0,0 +1,22 @@
...
+# Identifier
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+
+- Subject:
+  datatype: string
+  type: sensor
+  description: Subject for the authetification of the occupant.
```
gunnarx: typo: authentification

spec/Occupant/Identifier/Identifier.vspec
```diff
@@ -0,0 +1,22 @@
...
+  
+- Issuer:
+  datatype: string
+  type: sensor
+  description: Issuer for the authetification of the occupant.
```
gunnarx: same...

spec/Occupant/Identifier/Identifier.vspec
```diff
@@ -0,0 +1,19 @@
...
+#
+#
+# Identifier
+#
+
+- Subject:
+  datatype: string
+  type: sensor
+  description: Subject for the authentification of the occupant.
+
```
danielwilms, line 15: @gunnarx: improve the documentation (point to OAuth)

chheis: @danielwilms added OAuth 2.0

spec/Occupant/Identifier/Identifier.vspec
chheis: done

spec/Occupant/Identifier/Identifier.vspec
chheis: done

spec/VehicleSignalSpecification.vspec
```diff
@@ -226,3 +227,27 @@
...
+#
+- Vehicle.Occupant:
+  type: branch
+  description: Occupant data.
+
+# Include the Occupant/Occupant.vspec file and attach all its data under the
+# Occupant branch created above.
+
+#include Occupant/Occupant.vspec Vehicle.Occupant
+
```
danielwilms, line 242: @chheis: As I understand it, now you have exactly one occupant and one driver. And you define one time the role of the Occupant through the path (vehicle.driver) and one time through the Role attribute. Maybe we should instantiate the Occupant or Only use driver for now and not use the attribute. What do you think?

chheis: @danielwilms: As I mean it, there are no instances here, just Type definitions. So there can be 0 to n Occupants, with the role passenger. one with the role CoPilot. As I rethink about it, it is really not necessary to have the role "driver". As it is a special type already. So your suggestion would be to leave out the occupant currently? Or is there a better solution with the inheritance between Occupant and driver?
For the general instance topic (@SebastianSchildt) we should have a separate discussion.

danielwilms: > So there can be 0 to n Occupants, with the role passenger

Right now in the vspec definition, there's just one. But this hits especially the instance discussion we're having. One solution could be to split it. Having 
1. `driver` signals with or without occupant directly under the vehicle and 
1. the occupants next as sub-branch of the seat, as it then resembles the structure and you have already sensors like hasPassenger, isBelted etc.

Other opinions @SebastianSchildt, @gunnarx, @UlfBj?

chheis: @danielwilms I have followed your suggestion. Removed Role inside Occupant. I also made Occupant as branch under "Cabin.Seat" 
Driver is a top level branch and includes the attributes of the "Occupant" (without sub branch / transparent use of all attributes of occupant)

SebastianSchildt (approved):
👍

danielwilms (approved):

Issue #165, opened 2020-06-19
DTC node missing


UlfBj:
@UlfBj : After reading this article
https://otonomo.io/blog/extended-vehicles-a-new-eu-regulation-based-framework-for-car-data-access/ 
where it says that DTCs are some of the data that must be accessible from Sept 2020.
I searched on DTC in the VehicleSignalSpecification.id file, and found the signals below. 
To my understanding none of them provide the DTCs (the freeze DTC being an exception). 
I think we should add DTC to the tree, with the new array datatype (unlimited).
Vehicle.OBD.DistanceSinceDTCClear: 0da628e2c69d561eb86216ddcb6e7b2a
Vehicle.OBD.DriveCycleStatus.DTCCount: 312856f746ff560e8098c19196964d3b
Vehicle.OBD.FreezeDTC: 5b87fae8dda4522aae209ae528960782
Vehicle.OBD.Status.DTCCount: 4afdf65e788c5f69baf682597e69fb67
Vehicle.OBD.TimeSinceDTCCleared: 66ea3984a2585dcdaaf6452eef835c0d
Vehicle.OBD.WarmupsSinceDTCClear: a63ba60721785fc591e3dd067c4dc2ae
The article also says "It should be in standardized formats (while acknowledging that standards are not yet available)". If we want VSS to become the standard for this, then we should at least ensure it meets what is required.

Schildt: Yes, we need DTC, that is always the „Hello world”, when using OBD… I think I already mentioned it as an example in the array discussion, so I agree: This is exactly a usecase for arbitrary length arrays.
So at least we need a list of DTCs as array
(Maybe we want to separate P/U/C.. codes or maybe generic from manufacturer specific, but I suppose a flat list is enough, that is also what OBD gives you)

UlfBj:
@gunnarx 
> After reading this article
> https://otonomo.io/blog/extended-vehicles-a-new-eu-regulation-based-framework-for-car-data-access/
> where it says that DTCs are some of the data that must be accessible from
> Sept 2020.

It is a reasonable extension of the requirement already today to provide
standard diagnostics on the physical OBD-2 port.  In the same way, as the
OEMs are implement diagnostics-over-the-air, the wider industry is asking
that the same capability is made available to other third-party companies
involved in the business of car service station, or other.


> I searched on DTC in the VehicleSignalSpecification.id file, and found
> the signals below.
> To my understanding none of them provide the DTCs

[...]

PR #177, opened 2020-08-18
Add List of DTCs to OBD branch

SebastianSchildt:
The OBD branch was missing the list of DTCs itself (even so DTCCount, DistanceSinceDTCClear etc. was already there).

This is related to #165 

For the OBD branch we should follow OBD semantics for DTCs, and not trying to be clever about presenting this information. Consequently I used an array of strings, I do however think not all our tooling can deal with the array definition yet. I get an error on vspec2c:
``` 
./vss-tools/vspec2c.py -i:./spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).h vss_rel_$(cat VERSION)_macro.h
Illegal data type: 'string[]'
Signal: Vehicle_OBD_DTCList
Try: int8 uint8 int16 uint16 int32 uint32 int64 uint64
     float double string boolean stream
make: *** [Makefile:30: c] Error 255
```
Can/should/will we repair that? Who knows how to?

### DIFF
--- a/spec/OBD/OBD.vspec
+++ b/spec/OBD/OBD.vspec
@@ -1,4 +1,5 @@
 #
+# (C) 2020 Robert Bosch GmbH
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
 #
@@ -42,6 +43,11 @@
   type: sensor
   enum: [ "spark", "compression" ]
   description: Type of the ignition for ICE - spark = spark plug ignition, compression = self-igniting (Diesel engines)
+
+- DTCList:
+  datatype: string[]
+  type: sensor
+  description: List of cuurently active DTCs formatted according OBD II standard ([P|C|B|U]XXXXX )
 
 - FreezeDTC:
   datatype: string


SebastianSchildt:
Ok, vspec2c.py can not handle array datatypes. Travis works because it is only doing json. Is this a blocker? Somebody knows something about vspec2c?

gunnarx:
I think @UlfBj knows about vspec2c?

spec/OBD/OBD.vspec
```diff
@@ -43,6 +44,11 @@
   enum: [ "spark", "compression" ]
   description: Type of the ignition for ICE - spark = spark plug ignition, compression = self-igniting (Diesel engines)
 
+- DTCList:
+  datatype: string[]
+  type: sensor
+  description: List of cuurently active DTCs formatted according OBD II standard ([P|C|B|U]XXXXX )
```
gunnarx: *currently

gunnarx: Is there a more formal way to specify the document you refer to when you say "OBD II standard"?

gunnarx (changes requested):

SebastianSchildt:
Addresses the "Diagnostic Trouble Code" point from #180

spec/OBD/OBD.vspec
SebastianSchildt: Added reference to SAE standard, see current commit

spec/OBD/OBD.vspec
```diff
@@ -43,6 +44,11 @@
   enum: [ "spark", "compression" ]
   description: Type of the ignition for ICE - spark = spark plug ignition, compression = self-igniting (Diesel engines)
 
+- DTCList:
+  datatype: string[]
+  type: sensor
+  description: List of currently active DTCs formatted according OBD II (SAE-J2021DA) standard ([P|C|B|U]XXXXX )
```
SebastianSchildt, line 50: It is a type :-/ Needs to be SAE-J2012

chheis: changed to SAE-J2012DA_201812

SebastianSchildt: Thx is @chheis , looking good now

danielwilms:
meeting 13/10/20

PR #182, opened 2020-08-26
spec/Vehicle: add IgnitionOn signal

### LINKED ISSUE #179: Signal for ignition/key state
I would like the ability to determine, via VSS, the current ignition/engine state of the car. As far as I can tell, there is no such signal. The closest would be `Vehicle.IgnitionOnTime`, however one must compare two successive signals to determine if the ignition is on.

I propose adding an `IgnitionOn` boolean or some kind of `KeyState` which would be one of `[ON, ACC, OFF]` which would also suffice for my purposes. I'm happy to implement this in the spec, but I'm interested to here thoughts on this signal addition and under which node it should live.

rfrowe:
Resolves #179


rfrowe:
Hmm so I generated a JSON spec to trigger UUID generation for `Vehicle.IgnitionOn` so I could commit that as well, but it seems I've picked up a bunch of other sensors that did not have UUIDs already. Let me know what you want me to do, I can remove the unrelated UUIDs, leave them in, or not add any UUIDs.

gunnarx:
This commit has changes to the .id file.  @danielwilms What did we decide on the ID file?
I'm thinking we want to .gitignore it if it is generated material?

danielwilms:
> This commit has changes to the .id file. @danielwilms What did we decide on the ID file?
I'm thinking we want to .gitignore it if it is generated material?

Indeed. I added a reminder #185

danielwilms:
Otherwise fine for me. We could merge it as is, as the issue is now there.

danielwilms (approved):

Issue #186, opened 2020-09-09
Remove instances from leaf


danielwilms:
Instances in a leaf are not supported anymore. This needs a change in the specification, wherever instances in a leaf occur.

PR #188, opened 2020-09-16
Removed instances from leaf nodes

danielwilms:
Removed instances from:
* ADAS > ObstacleDetection > Distance
* Cabin > Lights > Spotlight
* Cabin > Seat > SeatPosCount

Fixed typo in datatype in Driver.vspec

Fixes: #186

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -83,10 +83,14 @@
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-- ObstacleDetection.DistanceToObject:
+- ObstacleDetection.Detector:
+  type: branch
   instances:
     - ["Front","Rear"]
-    - ["Left", "Right"]
+    - ["Left", "Middle", "Right"]
+  description: Device to detect obstacles in range of the vehicle.
+
+- ObstacleDetection.Detector.DistanceToObject:
   datatype: uint16
   type: sensor
   unit: m
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -83,18 +83,6 @@
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-- ObstacleDetection.Detector:
-  type: branch
-  instances:
-    - ["Front","Rear"]
-    - ["Left", "Middle", "Right"]
-  description: Device to detect obstacles in range of the vehicle.
-
-- ObstacleDetection.Detector.DistanceToObject:
-  datatype: uint16
-  type: sensor
-  unit: m
-  description: Distance to object in meters
 
 
 #
--- a/spec/Cabin/InteriorLights.vspec
+++ b/spec/Cabin/InteriorLights.vspec
@@ -56,6 +56,11 @@
 
 - Spotlight.IsPassengerOn:
   datatype: boolean
-  instances: ["Left","Right"]
   type: actuator
-  description: Is passenger light on
+  description: Is light on the passenger side switched on
+
+
+- Spotlight.IsDriverOn:
+  datatype: boolean
+  type: actuator
+  description: Is light on the drivers side switched on
--- a/spec/Cabin/InteriorLights.vspec
+++ b/spec/Cabin/InteriorLights.vspec
@@ -54,13 +54,13 @@
   type: sensor
   description: Is a shared light across a specific row on
 
-- Spotlight.IsPassengerOn:
+- Spotlight.IsLeftOn:
   datatype: boolean
   type: actuator
-  description: Is light on the passenger side switched on
+  description: Is light on the left side switched on
 
 
-- Spotlight.IsDriverOn:
+- Spotlight.IsRightOn:
   datatype: boolean
   type: actuator
-  description: Is light on the drivers side switched on
+  description: Is light on the right side switched on
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -170,9 +170,9 @@
   value: 0
   description: Number of seat rows in vehicle
 
+
 - SeatPosCount:
-  instances: Row[1,4]
-  datatype: uint8
+  datatype: uint8[]
   type: attribute
   value: 0
   description: Number of seats across each row from the front to the rear
--- a/spec/Driver/Driver.vspec
+++ b/spec/Driver/Driver.vspec
@@ -16,26 +16,25 @@
   type: sensor
   unit: percent
   description: Distraction level of the driver will be the level how much the driver is distracted, by multiple factors. E.g. Driving situation, acustical or optical signales inside the cockpit, phone calls
-  
+
 - EyesOnRoad:
   datatype: boolean
   type: sensor
   description: Has driver the eyes on road or not?
-  
+
 - AttentiveProbability:
   datatype: float
   type: sensor
   unit: percent
   description: Probability of attentiveness of the driver.
-  
+
 - FatigueLevel:
   datatype: float
   type: sensor
   unit: percent
   description: Fatigueness level of driver. Evaluated by multiple factors like trip time, behaviour of steering, eye status.
-  
+
 - HeartRate:
   datatype: UInt16
   type: sensor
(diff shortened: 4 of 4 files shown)


spec/Cabin/InteriorLights.vspec
```diff
@@ -56,6 +56,11 @@
 
 - Spotlight.IsPassengerOn:
```
SebastianSchildt, line 57: Spotlight is some interior light, and haveing only one implies, it is only on one seat (and driver)?
Does it make sense "generally"? There might be lights over rear seats also... But i think this comment is general, not specifically tied to this PR...

spec/ADAS/ADAS.vspec
```diff
@@ -83,10 +83,14 @@
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-- ObstacleDetection.DistanceToObject:
+- ObstacleDetection.Detector:
```
SebastianSchildt, line 86: I am fine with "Detector", I have nit checked the whole ADAS structure, but is "ObstacleDetection.Detection not double? So either have branches

Obstacle.Detector. xxxx

or

ADAS.ObstacleDetector -> INsances
(but not sure about isError and isActive... But lets say side cam broken, you could still see forward....

SebastianSchildt (commented):
Generally build, but not with the new code path, but agian maybe not a problem with this specific PR:
```
./vss-tools/vspec2csv.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).csv
Traceback (most recent call last):
  File "./vss-tools/vspec2csv.py", line 97, in <module>
    tree = vspec.load_tree(args[0], include_dirs)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 250, in load_tree
    return render_tree(dict_tree, True)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 654, in render_tree
    render_subtree(child_nodes, tree_root)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 668, in render_subtree
    render_subtree(child_nodes, new_element)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 665, in render_subtree
    new_element = VSSNode(element_name, current_element, parent=parent)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/model/vsstree.py", line 60, in __init__
    self.data_type = VSSDataType.from_str(source_dict["datatype"])
  File "/home/sebastian/vehicle_signal_specification/vss-tools/model/constants.py", line 137, in from_str
    raise Exception("Invalid VSS type %s only support %s" % (name, VSSDataType.values()))
Exception: Invalid VSS type uint8[] only support ['Int8', 'UInt8', 'Int16', 'UInt16', 'Int32', 'UInt32', 'Int64', 'UInt64', 'Boolean', 'Float', 'Double', 'String', 'UNIX Timestamp']
make: *** [Makefile:22: csv] Error 1
```

spec/Cabin/InteriorLights.vspec
danielwilms: Remove Spotlight from the PR and create ticket.

spec/ADAS/ADAS.vspec
danielwilms: Use Obstacle instead of sensor?

gunnarx: @danielwilms wrote:
> Use Obstacle instead of sensor?

I don't understand.  Where do you see "sensor" in the current example?  Is it in comparison to Sebastian's proposal or your own (which uses "Detector")?

I think "Obstacle" is not a synonym for "Sensor", so what more generally are we trying to achieve with the change?
Maybe I'm missing something.

@SebastianSchildt wrote:
>but is "ObstacleDetection.Detection not double? 

I don't see ObstacleDetection.**Detection**, I see ObstacleDetection.**Detector** ?
Sorry to nitpick, but it matters for understanding of the conversation.
(IMHO, Detection+Detector is not duplication in itself.  The only thing that matters then is if we really want to model the "detector" and if that is the best word (compared to for example sensor).

Overall, you might know I'm not a fan of adding too much depth to the tree if not needed.  It is the data that matters to me, not to model the actual physical setup of the car.   So, a "Sensor" node should only be mentioned if we need the extra level to reasonably explain the meaning of the data items.  

Often "<Concept>.<DataItem>" is fine for me.  It is therein _implicit_ and well known that there is some sensor that will be measuring the DataItem, but that's not necessary in the data taxonomy.   We are not _primarily_ modeling a physical car, we are primarily  organizing data items, in my opinion.   (In many cases we also do not know or care where and how the data item was actually measured).

So as an example `ObstacleDetection.DistanceToObject` was fine for me for a single object and I am likely to prefer it over `ObstacleDetection.MeasuringThing.DistanceToObject` no matter if you call MeasuringThing, "Detector" or "Sensor" or something else.

But I think there might be more to do here -  is a single MeasuringThing (sensor) instance able to detect more than one object for example?  If it is, then putting DistanceToObject as a child of a sensor-instance still does not help, because there are still multiple instances of the distance.

I think it might be easier to just use this PR for what it is for (remove instances) and avoid renaming anything.  And then discussing renaming separately?

danielwilms:
> I think it might be easier to just use this PR for what it is for (remove instances) and avoid renaming anything. And then discussing renaming separately?

I think it's a good proposal, but then I would remove it from the PR.

>> Use Obstacle instead of sensor?
> I don't understand. Where do you see "sensor" in the current example? Is it in comparison to Sebastian's proposal or your own (which uses "Detector")?

This was meant as a reminder, that we should discuss, if the sensor should be modelled (e.g. Detector) or the Obstacle, as the sensor is just the vehicle of measuring the distance to the obstacle. I like the latter, but would go for your proposal to solve that separately.

spec/Cabin/InteriorLights.vspec
danielwilms: @SebastianSchildt: looking again at the PR, it is defined for each row. Maybe the solution is not so bad at the end. For every row you have shared, driver side, and passenger side. Option would be to rename it to left/right, so that it matches how we model it in other places.

spec/Cabin/InteriorLights.vspec
```diff
@@ -54,8 +54,13 @@
   type: sensor
   description: Is a shared light across a specific row on
 
-- Spotlight.IsPassengerOn:
+- Spotlight.IsLeftOn:
```
adobekan, line 57: Spotlight is Dome Light within the cabin?

spec/Cabin/InteriorLights.vspec
```diff
@@ -54,8 +54,13 @@
...
-  instances: ["Left","Right"]
   type: actuator
-  description: Is passenger light on
+  description: Is light on the left side switched on
+
+
+- Spotlight.IsRightOn:
+  datatype: boolean
+  type: actuator
+  description: Is light on the right side switched on
```
adobekan, line 66: Punctuation mark missing.

adobekan (approved):

spec/Cabin/InteriorLights.vspec
gunnarx: The spec is named InteriorLights, so from my perspective it must mean this.  But at the same time I do think that it would each fully qualified node should have a description that can be understood on its own.  So it is a good idea to be quite detailed in the description I mean.

spec/Cabin/InteriorLights.vspec
gunnarx: > Spotlight is some interior light, and haveing only one implies, it is only on one seat (and driver)?

Yes, if only one is modelled it might not be enough.  But I want to highlight as I have done before that I think in general we should model a data hierarchy, and not primarily the design of the physical car.  Here, to my understanding there is only one _branch_ named Spotlight, that is true,  but Spotlight is here a category more than a physical thing?   The branch could have many leaves under it, thus implying many lights that are of type Spotlight, do you agree with my thinking?

gunnarx (approved):
Overall I would agree to merge the change.  I'm sure some minor tweaking will happen in the future but it is a good start and it gets us using the instances.

PR #193, opened 2020-09-22
Fix datatype in Driver.vspec

### LINKED ISSUE #191: Datatype value in Driver.vspec should be "boolean", not "bool"
https://github.com/GENIVI/vehicle_signal_specification/blob/e21f5506eaea04b9ebe278d99ada2360c2c6a046/spec/Driver/Driver.vspec#L21

SebastianSchildt:
Fixes #191


danielwilms:
fixed in #188, but makes sense to have it fixed immediately.

PR #196, opened 2020-09-29
Battery charging properties

chheis:
Adding new properties with the Charging branch


spec/Powertrain/EnergyStorage/Battery.vspec
```diff
@@ -83,3 +84,103 @@
...
+  datatype: UInt8
+  type: actuator
+  min: 0
+  max: 100
+  unit: percent
+  description: Maximum charge level for battery, can potentially be set manually.
+
+- Charging.MaximumChargingCurrent:
+  datatype: float
+  type: actuator
```
gunnarx, line 102: Should it be an actuator?  What happens when you write to it?

dakrbo: The idea was that this would limit the current the battery pack takes during charging - I believe eg the BMW i3 offers a feature like that. Not sure whether this is common enough a feature to warrant making this thing an actor, though. 
(I know the 1.5gen Renault Zoe, and Tesla - not even sure if they can do that, never needed that feature)

chheis: @danielwilms @gunnarx As we agreed to move the "sensor vs actuator" discussion in a special issue, can we continue to merge the changes?

dakrbo: Also, we can change this specific thing to be a sensor (instead of an actuator) - as mentioned, I only know one vehicle which offered this option - and then it only worked in certain cases iirc. So definitely not hung up on this specific property...

SebastianSchildt (approved):
Works for me

chheis (commented):
fine for me (approved)

danielwilms:
Rebase & Merge @danielwilms

danielwilms:
Create new PR, as I cannot push on fork.

Issue #197, opened 2020-09-30
Sensors vs. Actuators


danielwilms:
There is an issue with the current understanding of sensor vs. actuator.

@gunnarx argued, that contrary to the current specification, any data entry can be assumed writeable, unless it's an attribute (const). Given example: If an ECU A with a specific sensor wants to distribute its own value, it might want to do that over the one vss-enabled protocol, which the other entities are using to consume that value. In that case ECU A needs the rights to write that value and not being limited by the specification.

Some thoughts on that:
1. With the model we target primarily creating a common vocabulary within the industry and IMO even more one for non-domain experts. I personally think the fact, that there is an indication of whether a data entry is writable or not adds a lot of value, especially to the latter. 
1. The model describes concepts, which are used in the vehicle domain. And this determines, if the value can be changed or not. That a sensor has to communicate the data somehow, doesn't change that fact. If you use the argument above, there wouldn't be any place for attributes either. The additional information serves its purpose describing the concept. 
2. This was one of the first big changes towards v2. It has been discussed and is now in place. We can go forward to have the discussion for v3, and I think it fits into the scope of deployment/layering. But let's not reopen this issue and try to move on for now.

What do you think, @UlfBj, @chheis, @SebastianSchildt, @gunnarx, @adobekan?

UlfBj:
Whether a signal is read-only ,or read/write is key information for how a client will interact with it. To define this follows a well-known pattern, it is not something specific to VSS.

danielwilms:
@danielwilms make proposal how it can be written.

PR #199, opened 2020-10-08
Clarify description of attribute, sensor and actuator

danielwilms:
Fixes: #197

### DIFF
--- a/docs-gen/content/rule_set/data_entry/_index.md
+++ b/docs-gen/content/rule_set/data_entry/_index.md
@@ -7,7 +7,7 @@ weight: 2
 
 # Data Entry
 Leaf nodes of the tree contain metadata describing the data associated to the node.
-This specification makes a distinction between signals - in the following as ```sensor```, ```actuator``` and ```stream``` - and ```attribute```.
+In order to help application developers, who are using the specification, it makes a distinction between signals - in the following as ```sensor```, ```actuator``` and ```stream``` - and ```attribute```.
 The difference between a signal and an attribute is that the signal has
 a publisher (or producer) that continuously updates the signal value while an
 attribute has a set value, defined in the specification, that never changes.
@@ -15,7 +15,7 @@ As summary, besides [```branch```](/rule_set/branches) type can be:
 
 * **```attribute```**: attributes are not expected to change once they're set (e.g. vehicle identification number)
 * **```sensor```**: sensor values describe the current state of the vehicle and change over time, as the state of the vehicle changes (e.g. odometer).
-* **```actuator```**: actuating signals describe current state of the vehicle and change, when the state of the vehicle changes. Actuating on the value, leads to a change of the vehicle state itself (e.g. door lock). 
+* **```actuator```**: actuating signals describe current state of the vehicle and change, when the state of the vehicle changes. Writing the value, leads to a change of the vehicle state itself (e.g. door lock).
 * **```stream```**, data stream like video.
 
 Examples you'll find in the [sensor and actuator chapter](/vehicle_signal_specification/rule_set/data_entry/sensor_actuator).
--- a/docs-gen/content/rule_set/data_entry/_index.md
+++ b/docs-gen/content/rule_set/data_entry/_index.md
@@ -13,9 +13,9 @@ a publisher (or producer) that continuously updates the signal value while an
 attribute has a set value, defined in the specification, that never changes.
 As summary, besides [```branch```](/rule_set/branches) type can be:
 
-* **```attribute```**, which describes static read-only value.
-* **```sensor```**, which describes non-static read-only signal.
-* **```actuator```**, which describes a non-static read-write signal.
+* **```attribute```**: attributes are not expected to change once they're set (e.g. vehicle identification number)
+* **```sensor```**: sensor values describe the current state of the vehicle and change over time, as the state of the vehicle changes (e.g. odometer).
+* **```actuator```**: actuating signals describe current state of the vehicle and change, when the state of the vehicle changes. Actuating on the value, leads to a change of the vehicle state itself (e.g. door lock). 
 * **```stream```**, data stream like video.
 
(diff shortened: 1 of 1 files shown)


UlfBj (changes requested):
I think the original descriptions were quite accurate, so keeping them with additional information, like my proposal below, is what I would prefer.

**```attribute```**, which describes a static read-only value (e.g. vehicle identification number).
* **```sensor```**, which describes a non-static read-only signal. Sensor values describe the current state of the vehicle and change over time, as the state of the vehicle changes (e.g. odometer). Reading the sensor value does not change the state of the vehicle.
* **```actuator```**, which describes a non-static read-write signal. Actuator signals describe current state of the vehicle, and writing to the actuator changes the vehicle state (e.g. door lock).

dakrbo:
Just to state the obvious (I guess): this question is about, do we consider the semantics of 'things in the tree' from the PoV of being-the-system, or from the viewpoint of a consuming-application?
* being-the-system of course implies that any property value can, and most will, change - no matter what is causing that change (a physical sensor output, or a consuming-application action)
* the consuming-application view benefits from a clear distinction between 'this is something you can/cannot change' - in my opinion

So if I were to phrase the question, it would be "which viewpoint do we think the VSS tree spec should address?"

If it's the second one, this has some implications on implementation/behaviour - ie, a consuming-application role shouldn't be allowed to change sensor values (again, stating the obvious).

Sorry if that was redundant, I'm a newbie here and all...

gunnarx:
> Just to state the obvious (I guess): this question is about, do we consider the semantics of 'things in the tree' from the PoV of being-the-system, or from the viewpoint of a consuming-application?

Yes, this is my point, I think.  I feel the importance of the model being generic.  There is a lot more discussion now about its usage inside the vehicle:

Here is the point I tried to make on the previous call with an Example:

`[Sensor]` is wired to `[ECU1]` -> communicates with -> `[(Telematics) ECU2]` -> communicates with `[Cloud system]`

If the system design is such that ECU2 holds the database of latest values then ECU1 must "write" the sensor value it picked up to the database in ECU2.  Thus, here "sensor" is writable from one viewpoint.  It is regardless of protocol since VSS is not tied to a protocol, but just for the example's sake you could imagine **theoretically** that the writing from ECU1 to ECU2 uses something like W3C Gen 2 REST protocol.  So then the protocol must also allow writing to a Sensor type node.  But that is just an example, and in the end it's about looking at the system generically.

Also, I have a few concerns if the VSS model+catalog combination can define behavior that strictly as to say that an action must be taken if an actuator is written.  Door lock is a good (counter) example.  It seems likely that you want more control over the possible failures that may occur and use RPC for that instead.  So... it's rather a "desired" lock state that you are writing?  The problem is we have some parts of the standard catalog are like DoorState.Locked which seems expected to be both readable and writable, (and therefore classed as Actuator), and then comes some expected behavior attached to that.

gunnarx:
> the consuming-application view benefits from a clear distinction between 'this is something you can/cannot change' 
> - in my opinion

Although I might look at this as a completely generic model on top of which you must put an access-control framework anyway, I also agree that there is some usefulness to having such a hin in the database for the users of a typical protocol from outside the vehicle.

I think we may find a compromise and keep the model without major rework if we use the description that Ulf provides with a few slight changes and additions:

- Define the viewpoint from which this is defined
- Add that we recognize flexible system design expectations (as I wrote above)
- Maybe that the associated behavior is more proposed than a rule (maybe in particular regarding the action expected from an actuator?)

docs-gen/content/rule_set/data_entry/_index.md
```diff
@@ -13,9 +13,9 @@ a publisher (or producer) that continuously updates the signal value while an
 attribute has a set value, defined in the specification, that never changes.
 As summary, besides [```branch```](/rule_set/branches) type can be:
 
-* **```attribute```**, which describes static read-only value.
-* **```sensor```**, which describes non-static read-only signal.
-* **```actuator```**, which describes a non-static read-write signal.
+* **```attribute```**: attributes are not expected to change once they're set (e.g. vehicle identification number)
```
gunnarx, line 16: Good summary I think.
In a complementary "system design" description we might clarify that this is either done by being set by a certain system, or by "software-download" of a configuration file, etc.  But what you write here is a short and valid description I think.

docs-gen/content/rule_set/data_entry/_index.md
```diff
@@ -13,9 +13,9 @@ a publisher (or producer) that continuously updates the signal value while an
...
+* **```sensor```**: sensor values describe the current state of the vehicle and change over time, as the state of the vehicle changes (e.g. odometer).
```
gunnarx, line 17: Also short and good, only I'm concerned with having it general enough.
There's some big movements going on as you know in CVII and so on, so in terms of a data model I'm not even sure that the VSS model is only describing "the vehicle" in the future.  But we can decide on that after clarifying the overall "from which viewpoint" discussion.

docs-gen/content/rule_set/data_entry/_index.md
```diff
@@ -13,9 +13,9 @@ a publisher (or producer) that continuously updates the signal value while an
...
+* **```actuator```**: actuating signals describe current state of the vehicle and change, when the state of the vehicle changes. Actuating on the value, leads to a change of the vehicle state itself (e.g. door lock).
```
gunnarx, line 18: > ac·tu·ate  (ăk′cho͞o-āt′)
> tr.v. ac·tu·at·ed, ac·tu·at·ing, ac·tu·ates
> 1. To put into motion or action; activate: electrical relays that actuate the elevator.
> 2. To cause to take action: a speech that actuated dissenters.

Not wrong, but we might be clearer if we use a bit more specific language "writing to the signal".   For the rest, consider the discussion in the comments.

gunnarx:
@UlfBj  
> Reading the sensor value does not change the state of the vehicle.

I feel this part might be unnecessary.  Most systems are unaffected by reading, it is the expected default behavior.

PR #202, opened 2020-10-20
promoting global vehicle data/information to Vehicle root branch

chheis:
Promote some date points like Vehicle position to the top level of the vehicle
Signed-off-by: C. Heissenberger <christian.heissenberger@de.bosch.com>

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -125,7 +125,6 @@
   description: Washer fluid level as a percent. 0 = Empty. 100 = Full.
 
 
-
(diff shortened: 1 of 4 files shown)


danielwilms:
I would vote against moving location, as it's not logically wrong in the current place. The change is more of a nice to have. Location is really used a lot and the change would complicate things for people, who are already using it.

chheis:
Used deprecation and moved to: https://github.com/GENIVI/vehicle_signal_specification/pull/224

PR #209, opened 2020-10-26
Updated vss-tools to support instances

danielwilms:


danielwilms:
Updated the tools for instance support.

PR #211, opened 2020-11-09
Added various combustion engine properties

dakrbo:
Signed-off-by: Daniel (dakrbo) <daniel.krippner@de.bosch.com>

- Added a bunch of combustion-engine properties, though nothing revolutionary (heh)
- Removed 'ambient temperature'  - because this is a vehicle-level property, and already present on that level

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -1,6 +1,7 @@
 #
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
+# (C) 2020 Robert Bosch GmbH
 #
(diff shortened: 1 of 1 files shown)


SebastianSchildt (commented):
Seems fine 👍

danielwilms (approved):
Looks good, just one question, why has `engine.ambientairtemperature` been removed?

SebastianSchildt:
As discussed last meeting, because it is a duplicate and can be found already here: https://github.com/GENIVI/vehicle_signal_specification/blob/1565e2f2426593521f4d81b24e84efaa5cc5cdf5/spec/Vehicle/Vehicle.vspec#L197
(which seems the better position)

PR #212, opened 2020-11-09
Fix/fix units again

### LINKED ISSUE #210: Inconsistency "datatype" definition in json
curl https://genivi.github.io/vehicle_signal_specification/releases/nightly/vss_rel_2.0.0-alpha+006.json | grep datatype | sed "s/^[ \t]*//" | sort | uniq

```
"datatype": "boolean",
"datatype": "Double",
"datatype": "float",
"datatype": "Float",
"datatype": "int16",
"datatype": "Int16",
"datatype": "int32",
"datatype": "int8",
"datatype": "Int8",
"datatype": "string",
"datatype": "string[]",
"datatype": "uint16",
"datatype": "UInt16",
"datatype": "uint32",
"datatype": "Uint32",
"datatype": "uint8",
"datatype": "uint8[]",
"datatype": "UInt8",
```

Some with upper case and some in lower case. In the documentation [Vehicle Signal Specification > Rule Set > Data Entry > Data Types](https://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/data_types/). 
IMO, it shall be unification.

SebastianSchildt:
Harmonize all units to lowercase (that was the majority, and easier to type), and also adapt the documentation.

I am not sure if we _require_ lower casp anywhhere, or both is valid, but in either case we should not be "messy" about this in the core spec.

This should fix #210 

After generating json locally:
```
$ cat vss_rel_2.0.0-alpha+006.json | grep datatype |  sed "s/^[ \t]*//" | sort | uniq 
"datatype": "boolean",
"datatype": "double",
"datatype": "float",
"datatype": "int16",
"datatype": "int32",
"datatype": "int8",
"datatype": "string",
"datatype": "string[]",
"datatype": "uint16",
"datatype": "uint32",
"datatype": "uint8",
"datatype": "uint8[]",
```

Somebody (Travis?) needs to check if the docs still generate, I do not know Hugo, and just installing it lead to some errors on my side (no matter I modify or not)

```
ERROR 2020/11/09 20:21:35 render of "page" failed: execute of template failed: template: _default/single.html:1:3: executing "_default/single.html" at <partial "header.html" .>: error calling partial: execute of template failed: template: partials/header.html:40:7: executing "partials/header.html" at <partial "menu.html" .>: error calling partial: "/home/sebastian/vehicle_signal_specification/docs-gen/themes/learn/layouts/partials/menu.html:100:23": execute of template failed: template: partials/menu.html:100:23: executin
```

### DIFF
--- a/spec/Driver/Driver.vspec
+++ b/spec/Driver/Driver.vspec
@@ -35,6 +35,6 @@
   description: Fatigueness level of driver. Evaluated by multiple factors like trip time, behaviour of steering, eye status.
 
 - HeartRate:
-  datatype: UInt16
+  datatype: Uint16
   type: sensor
   description: Heart rate of the driver.
--- a/spec/Driver/Driver.vspec
+++ b/spec/Driver/Driver.vspec
@@ -35,6 +35,6 @@
   description: Fatigueness level of driver. Evaluated by multiple factors like trip time, behaviour of steering, eye status.
 
 - HeartRate:
-  datatype: Uint16
+  datatype: uint16
   type: sensor
   description: Heart rate of the driver.
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -42,7 +42,7 @@
 
 - Battery.StateOfCharge.Target:
   type: actuator
-  datatype: Uint8
+  datatype: uint8
   min: 0
   max: 100
   unit: percent
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -27,14 +27,14 @@
 - Battery.StateOfCharge.Current:
   type: sensor
   unit: percent
-  datatype: Float
+  datatype: float
   min: 0
   max: 100.00
   description: Physical state of charge of the high voltage battery. This is not necessarily the state of charge being displayed to the customer.
 
 - Battery.StateOfCharge.Displayed:
   type: sensor
-  datatype: Float
+  datatype: float
   unit: percent
   min: 0
   max: 100.00
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -42,7 +42,7 @@
 
 - Battery.StateOfCharge.Target:
   type: actuator
-  datatype: UInt8
+  datatype: Uint8
   min: 0
   max: 100
   unit: percent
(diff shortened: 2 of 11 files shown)


chheis (approved):
looks good

spec/Powertrain/Battery.vspec
```diff
@@ -42,7 +42,7 @@
 
 - Battery.StateOfCharge.Target:
   type: actuator
-  datatype: UInt8
+  datatype: Uint8
```
gunnarx: misspelled.  OK it was fixed in d4fa0876

spec/Driver/Driver.vspec
```diff
@@ -35,6 +35,6 @@
   description: Fatigueness level of driver. Evaluated by multiple factors like trip time, behaviour of steering, eye status.
 
 - HeartRate:
-  datatype: UInt16
+  datatype: Uint16
```
gunnarx: Uint16 -> uint16.
OK, it was changed in a later commit.

spec/Private/CARFIT/Chassis/Wheel.vspec
```diff
@@ -12,7 +12,7 @@
   description: Brake pad wear status. True = Worn. False = Not Worn.
 
 - Brake.RotorWear:
-  type: Uint16
+  type: int16
```
gunnarx, line 15: This changes from unsigned to signed.  Intentional?

gunnarx:
I'm OK with this change in general.  One thing to consider for VSS, in my opinion, is to keep it as simple as possible for "non-programmers" to also define and work with the data.  I don't know if this change improves that or not... Just something to consider.

spec/Private/CARFIT/Chassis/Wheel.vspec
SebastianSchildt: a mistake.... Will fix and push again (Same happened to driver heartrate...)

SebastianSchildt:
I repaired the broken "Uint16" conversion. Needed to force push, to not make it look ugly in history...

Also updated tools, to be more strict: https://github.com/GENIVI/vss-tools/pull/26

gunnarx:
No problem.  Force pushing PRs is normal.  The "final" history is not decided until it is merged.

danielwilms (approved):

PR #218, opened 2020-11-13
Add carets to clarify powers

rfrowe:
Right now, `m/s2` and `cm/s2` are used for for meters per square second and centimeters on square second. However, this is a non-traditional syntax and can lead to ambiguities. Instead, I suggest the traditional caret notation.

For example, you have `l/100km` and `Nm`. If I were to combine these three syntaxes in one, what would `N3m` be?

Finally, many parsing programs won't recognize `s2` as squared seconds. The library that I'm using, pint, doesn't.

### DIFF
--- a/docs-gen/content/rule_set/data_entry/data_unit_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_unit_types.md
@@ -36,8 +36,8 @@ g          | Weight        | Grams
 kg         | Weight        | Kilograms
 g/s        | Flow          | Grams per second
 l/h        | Flow          | Liters per hour
-m/s2       | Acceleration  | Acceleration in meters per second squared
-cm/s2      | Acceleration  | Acceleration in centimeters per second squared
+m/s^2      | Acceleration  | Acceleration in meters per second squared
+cm/s^2     | Acceleration  | Acceleration in centimeters per second squared
 N          | Force         | Newton
 Nm         | Force         | Torque
 l          | Volume        | Liter
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -20,7 +20,7 @@
   datatype: uint16
   type: attribute
   description: Displacement in cubic centimetres.
-  unit: cm3
+  unit: cm^3
 
 - Configuration:
   datatype: string
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -225,19 +225,19 @@
 - Acceleration.Longitudinal:
   datatype: int32
   type: sensor
-  unit: m/s2
+  unit: m/s^2
   description: Vehicle acceleration in X (longitudinal acceleration).
 
 - Acceleration.Lateral:
   datatype: int32
   type: sensor
-  unit: m/s2
+  unit: m/s^2
   description: Vehicle acceleration in Y (lateral acceleration).
 
 - Acceleration.Vertical:
   datatype: int32
   type: sensor
-  unit: m/s2
+  unit: m/s^2
   description: Vehicle acceleration in Z (vertical acceleration).
 
 


SebastianSchildt:
I like this, and would suggest merging this soon (as it will be a pain to keep it in sync with other merges...), however it fails with new style tools, which are checking units as well. 
```
sebastian@xbionic:~/vehicle_signal_specification$ make csv
./vss-tools/vspec2csv.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).csv
Traceback (most recent call last):
  File "./vss-tools/vspec2csv.py", line 97, in <module>
    tree = vspec.load_tree(args[0], include_dirs)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 253, in load_tree
    return render_tree(dict_tree, True)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 797, in render_tree
    render_subtree(child_nodes, tree_root)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 811, in render_subtree
    render_subtree(child_nodes, new_element)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 808, in render_subtree
    new_element = VSSNode(element_name, current_element, parent=parent)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/model/vsstree.py", line 63, in __init__
    self.unit = Unit.from_str(source_dict["unit"])
  File "/home/sebastian/vehicle_signal_specification/vss-tools/model/constants.py", line 90, in from_str
    raise Exception("Invalid Unit %s only support %s" % (name, Unit.values()))
Exception: Invalid Unit m/s^2 only support ['mm', 'cm', 'm', 'km', 'km/h', 'm/s2', 'l', 'celsius', 'degrees', 'degrees/s', 'kW', 'kWh', 'kg', 'V', 'A', 's', 'min', 'weeks', 'months', 'UNIX Timestamp', 'pa', 'kpa', 'percent', 'cm3', 'PS', 'stars', 'g/s', 'g/km', 'kWh/100km', 'l/100km', 'l/h', 'mpg', 'lbs', 'Nm', 'rpm', 'inch', 'ratio']
make: *** [Makefile:22: csv] Error 1
``` 
Do we have a coresponding PR for that?

SebastianSchildt:
I guess it is as simple as modifying here https://github.com/GENIVI/vss-tools/blob/master/model/constants.py

@rfrowe do you volunteer? 😻 ?

@danielwilms Why is travis green?

chheis (approved):
I like the change. As our parsers have to deal with "/" as well it can also deal with "^" :)

danielwilms (approved):

danielwilms:
tool has to be adapted.

rfrowe:
@SebastianSchildt Ah I didn't know there was tooling that did such a thing. I'll try to get around to it today.

As for why travis is green, it only does `make clean json deploy` not `make csv` as you did. Perhaps this code path does not do unit checking.

rfrowe:
@SebastianSchildt @danielwilms yeah I can confirm `make json` does not throw the exception while `make json` does.

rfrowe:
@SebastianSchildt I've made the change and it can be found in https://github.com/GENIVI/vss-tools/pull/29 but I do have some comments.
0. The unit class does not enumerate `cm/s^2` even though it's an allowed unit.
0. `cm^3` is named `CUBICMETER` rather than `CUBICCENTIMETER`.
0. `cm^3` is used in Vehicle.PowerTrain.CombustionEngine.Displacement but is not present in the data type docs.

LMK if you want any of these fixed and I can add them as separate commits to this PR

rfrowe:
This will need another approval now. I missed `cm3` in my original commit.

danielwilms:
> As for why travis is green, it only does make clean json deploy not make csv as you did. Perhaps this code path does not do unit checking.

@rfrowe that's correct. I've opened [GENIVI/vehicle_signal_specification/issues/221] and will fix it.

PR #220, opened 2020-11-23
Makefile updated to support native tool.

UlfBj:
If the PR introducing the native tool n the VSS-Tools repo is accepted, then this update is needed. 
If not, it is to be rejected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,8 @@ install:
   - cd ..
 
 script:
-  - make clean json deploy
+  # NOTE: c and csv not built -- fix later
+  - make clean json franca binary
   - hugo -s ${DOCS_DIR}
 
 deploy:
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 
 .PHONY: clean all json franca c
 
-all: clean json franca csv c
+all: clean json franca csv binary c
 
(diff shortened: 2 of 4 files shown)


UlfBj:
The tool name is changed to binary.

danielwilms:
Checked it against the vss-tools master branch of Ulf, used in the[ corresponding PR](https://github.com/GENIVI/vss-tools/pull/28) and it works.

danielwilms (approved):

gunnarx:
Squashed the three commits together, and rebased on latest master.  Awaiting build check and will then merge.

gunnarx:
Travis file was not building the target all, only a few selected ones.  I pushed a commit to fix this and now we can again do the build check.

gunnarx:
OK, the PR also needs the update of the vss-tools submodule to include the binary tool.
(Can be done now that it is merged).  So we try again...

gunnarx:
One of the files had the wrong Pa(scal) unit.  This is unrelated to this PR but fixed now so we can get a successful build check (hopefully).   I'm not sure why this was not noticed before, but as noted above I changed  .travis build to build "all" now.  I'm still surprised that what we DID build before was not checking the units(??)

Issue #216, opened 2020-11-10
Deprecation


danielwilms:
Some nodes are moved with in the tree. Deprecation would be help to adapt. Therefore, the suggestion is to add a `deprecation` keyword to the nodes, which are moved. Using a keyword in the yaml node works nicely with the current tooling.

Example:
```
Navigation.CurrentLocation:
  type: branch
  aggregate: true
  description: The current latitude and longitude of the vehicle.
  deprication: @v2.1 moved to Vehicle.CurrentLocation
```

SebastianSchildt:
`depr_e_cation` 😁 

Semantic of @: Does it mean moved in 2.1?, so IN 2.1 deprecated

A general question is, do we want more semantic, like "machine readable" this moved to ... or jsut a general hum readable comment? I think human readabe string is enoug (can be put out by the compiler), becasue otherwise we ran into trouble if it is "moved _and_ renamed this and that...."

Just a marker and humnan readable string is also, how it is done C++, see jsoncons as a good example (search for JSONCONS_DEPRECATED_MSG here https://github.com/danielaparker/jsoncons), basically the compiler throws those stirngs into your face as warnings whenever you touch something old.

So tldr; I think this is the approach we want, in case I read correctly between the lines. I suggest putting this into the documentation in a more verbose way, and discuss on a PR

danielwilms:
@danielwilms add a PR with the documentation.

chheis:
First real example in PR here: https://github.com/GENIVI/vehicle_signal_specification/pull/224

danielwilms:
add description for commit

PR #223, opened 2020-12-01
Add deprecation documentation

danielwilms:
Fixes: #216

### DIFF
--- a/docs-gen/content/rule_set/basics.md
+++ b/docs-gen/content/rule_set/basics.md
@@ -78,10 +78,11 @@ During the process of model development, nodes might be
 moved or deleted. Giving developers a chance to adopt to the
 changes, the original nodes are marked as deprecated with the following rules.
 
-* Nodes, which are moved in the tree or removed from the specification are marked with the deprecation keyword.
+* Nodes, which are moved in the tree or are intended to be removed from the specification are marked with the deprecation keyword.
 * The string following the deprecation keyword shall start with the version, when the node was deprecated starting with `@` (e.g. `@2.1`) followed by the reason for deprecation.
-* If the node was moved, it shall be indicated by `moved to` followed by the new node name in dot notation as deprecation reason.
-* If the node is removed from the specification it shall be indicated by `removed` and optionally the reason for the removal as deprecation reason.
+* If the node was moved, it shall be indicated by `moved to` followed by the new node name in dot notation as deprecation reason. This keyword shall be used only
+if the meta-data of the moved node hasn't changed. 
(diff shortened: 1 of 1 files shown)


SebastianSchildt (commented):
I like it. As stated in discussion: Lets not overly complicate this in the first step. I consider my comments mere suggestions at the moment, I would even approve without them.
I think the mechanism with the `deprecation` metadata for nodes is sufficiently generic, that we can specify more semantics on a as-needed basis, once we hit some corner cases.

docs-gen/content/rule_set/basics.md
```diff
@@ -71,3 +71,24 @@ be included as well, as shown below:
...
+* If the node is removed from the specification it shall be indicated by `removed` and optionally the reason for the removal as deprecation reason.
+* Nodes which are deprecated will be removed from the specification, either in the second minor update or, if earlier, the next major update.
+
+### Example
+```
+Navigation.CurrentLocation:
+  type: branch
+  aggregate: true
+  description: The current latitude and longitude of the vehicle.
+  deprecation: @2.1 moved to Vehicle.CurrentLocation
```
erikbosch, line 93: Is it always clear what the next release will be called, e.g. that next release actually will be called 2.1 in this example? Or do we need a temporary syntax in Git like "@X.Y" to use for the "next release" that is replaced with the actual release name as part of the release process?

danielwilms:
Open points:
* Removed -> intent to remove
* clarify move to doesn't work with meta data changes
* recommend/indicate: server should keep deprecated nodes

docs-gen/content/rule_set/basics.md
danielwilms: @danielwilms: add to the release process

gunnarx (changes requested):
It looks like the format we suggest is not valid YAML :-(
Try pasting this into https://yamlvalidator.com/ 
```
- Navigation.CurrentLocation.Example:
  foo: bar
  deprecation: @2.1 moved to Vehicle.CurrentLocation.Longitude
``` 

The relevant specification is here - it states the @ sign is reserved:
https://yaml.org/spec/1.2/spec.html#c-reserved

The ability to use an off-the-shelf parser is critical IMHO.  It is the reason to use a standard format.  Therefore we need to come up with something different that is YAML compliant.

gunnarx:
A proposal 
`   deprecation: V2.1 moved to Vehicle.CurrentLocation`

SebastianSchildt:
Yes, but quotes are allowed in YAML

```
- Navigation.CurrentLocation.Example:
  foo: bar
  deprecation: "@2.1 moved to Vehicle.CurrentLocation.Longitude"
```

(If this is a"better" solution I do not know, but it is not a hack. Should validate)

PR #224, opened 2020-12-11
promoting global vehicle data/information to Vehicle root branch

chheis:
promoting global vehicle data/information to Vehicle root branch **_using deprecation_**

### removed
Navigation.CurrentLocation.Speed **removed** because doubled with Vehicle.Speed

### moved
complete branch: Navigation.CurrentLocation moved to Vehicle.CurrentLocation
complete branch: Chassis.Trailer moved to Vehicle.Trailer
Vehicle masses moved from Chassis to Vehicle

Signed-off-by: Christian Heissenberger <christian.heissenberger@de.bosch.com>

### DIFF
--- a/spec/Chassis/Chassis.vspec
+++ b/spec/Chassis/Chassis.vspec
@@ -226,10 +226,10 @@
 - Trailer:
   type: branch
   description: Trailer signals
-  deprecation: @2.1 moved branch to Vehicle.Trailer
+  deprecation: @2.1 moved to Vehicle.Trailer
 
 - Trailer.Connected:
   datatype: boolean
   type: sensor
   description: Signal indicating if trailer is connected or not.
-  deprecation: @2.1 moved to Vehicle.Trailer.Width
+  deprecation: @2.1 moved to Vehicle.Trailer.Connected
--- a/spec/Chassis/Chassis.vspec
+++ b/spec/Chassis/Chassis.vspec
@@ -19,7 +19,7 @@
   default: 0
   unit: kg
   description: Vehicle curb weight, in kg, including all liquids and full tank of fuel, but no cargo or passengers.
-  deprecation: @2.1 moved to Vehicle.CurbWeight
+  deprecation: V2.1 moved to Vehicle.CurbWeight
 
 - GrossWeight:
   datatype: uint16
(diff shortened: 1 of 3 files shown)


spec/Vehicle/Vehicle.vspec
```diff
@@ -295,3 +295,123 @@
...
+#
+# Vehicle Weight and Dimension attributes
+#
+
+- CurrentOverallWeight:
+  datatype: uint16
+  type: sensor
+  value: 0
+  unit: kg
+  description: Current overall Vehicle weight, in kg, including all liquids and full tank of fuel.
```
erikbosch, line 309: I know that we just moved this item, but is the description mentioning "full tank of fuel" correct? I.e. that the sensor providing this value needs to consider how full the tank is and adjust the value accordingly?

SebastianSchildt: My feeling is that "CurrentOverallWeight" is sort of the "best guess" how heavy a vehicle currently is, including everything potentially even cargo and passengers. (how accurate it is/can be depends on the vehicle).
@chheis may correct me 😀

danielwilms: I follow @erikbosch. from the description I'd rather think it's a static value, but the type `sensor` indicates, that it's dynamic. Maybe we should clarify that.

danielwilms (commented):
I'm undecided if you should have the deprecation just on the branch or on every leaf as well. That we should discuss.

spec/Chassis/Chassis.vspec
```diff
@@ -220,8 +226,10 @@
 - Trailer:
   type: branch
   description: Trailer signals
+  deprecation: @2.1 moved branch to Vehicle.Trailer
 
 - Trailer.Connected:
   datatype: boolean
   type: sensor
   description: Signal indicating if trailer is connected or not.
+  deprecation: @2.1 moved to Vehicle.Trailer.Width
```
danielwilms, line 235: looks like a copy paste error.

gunnarx:
I think each signal should have its own deprecation notice.  Even if it is a bit more work, it simplifies processing and I suppose in the future there should hopefullly not be too many of these major moves of complete branches...

gunnarx:
@danielwilms Check the build failure.  Perhaps this usage of @ is not valid YAML

spec/Vehicle/Vehicle.vspec
chheis: I changed the description according to the discussion

spec/Chassis/Chassis.vspec
chheis: thanks, corrected

gunnarx:
There is a parsing error when encountering the @ symbol on the deprecation line.
Also https://yamlvalidator.com seems to complain about the format?

PR #227, opened 2021-01-12
Release 2.0

danielwilms:
* Changed documentation
* Added release notes
* Release official json, csv & fidl file


VERSION
```diff
@@ -1 +1 @@
-2.0.0-alpha+006
+2.1-develop
```
erikbosch, line 1: Is this the commit that will be tagged as "Release 2.0"? If so shall this version not say "2.0.0"? (I.e. change to "2.1-develop" first in the following commit?

PR #234, opened 2021-02-03
Using float for speed and acceleration (Fixes #228)

### LINKED ISSUE #228: Changing Vehicle.Speed and Accelerations to float without limitations
Due to calculation loses (some function may use m/s internally). Would be better to have more accuracy for the speed.
https://github.com/GENIVI/vehicle_signal_specification/blob/master/spec/Vehicle/Vehicle.vspec 
And also to change the vehicle acceleration to float:
Acceleration.Longitudinal / Acceleration.Lateral / Acceleration.Vertical

erikbosch:
This makes it possible to use data with higher accuracy. This can be especially important with speed where multiple conversions may happen. For example - a vehicle may internally represent speed as m/s, but the end-user might want to work with miles/hour. If using int-representation (without decimals) for the intermediate format, then it might become important whether values shall be rounded upwards/downwards

Example: Assume a car driver wants to set cruise control to 70 mph. This corresponds to 112.6 km/h.  If using integer-representation (without decimals) the service using the actuator must make an active decision on whether it shall request 112 km/h or 113 km/h. In this case just rounding to 113 km/h theoretically might give legal problems (the car will move faster than requested by the user, and possibly exceed speed limits).

### DIFF
--- a/spec/Powertrain/Transmission.vspec
+++ b/spec/Powertrain/Transmission.vspec
@@ -48,6 +48,7 @@
   max: 250
   unit: km/h
   description: Vehicle speed, as sensed by the gearbox.
+  deprecation: V2.1 removed because doubled with Vehicle.Speed
 
 
 #
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -175,12 +175,10 @@
   description: Accumulated idle time in seconds.
 
 - Speed:
-  datatype: int32
+  datatype: float
   type: sensor
-  min: -250
-  max: 250
   unit: km/h
-  description: Vehicle speed, as sensed by the gearbox.
+  description: Vehicle speed
 
 - TravelledDistance:
   datatype: float
@@ -207,10 +205,8 @@
   description: Indicates whether the vehicle is stationary or moving
 
 - AverageSpeed:
-  datatype: int32
+  datatype: float
   type: sensor
-  min: -250
-  max: 250
   unit: km/h
   description: Average speed for the current trip
 
@@ -223,19 +219,19 @@
   description: Spatial acceleration
 
 - Acceleration.Longitudinal:
-  datatype: int32
+  datatype: float
   type: sensor
   unit: m/s^2
   description: Vehicle acceleration in X (longitudinal acceleration).
 
 - Acceleration.Lateral:
-  datatype: int32
+  datatype: float
   type: sensor
   unit: m/s^2
   description: Vehicle acceleration in Y (lateral acceleration).
 
 - Acceleration.Vertical:
-  datatype: int32
+  datatype: float
   type: sensor
   unit: m/s^2
   description: Vehicle acceleration in Z (vertical acceleration).
@@ -281,6 +277,7 @@
   type: attribute
   unit: s
   description: The time needed to accelerate the vehicle from a given start velocity to a given target velocity.
+  deprecation: V2.1 removed as ambiguous definition (start/stop-speed not defined)
 
 - cargoVolume:
   datatype: int16
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -34,7 +34,7 @@
 
 
 - CruiseControl.SpeedSet:
-  datatype: int32
+  datatype: float
   type: actuator
   unit: km/h
   description: Set cruise control speed in kilometers per hour


erikbosch:
This is proposed changes for #228. There are some open topics that better shall be discussed at the next VSS-meeting. I will add comments to highlight them.

spec/Vehicle/Vehicle.vspec
```diff
@@ -175,12 +175,10 @@
   description: Accumulated idle time in seconds.
 
 - Speed:
-  datatype: int32
+  datatype: float
   type: sensor
-  min: -250
-  max: 250
   unit: km/h
```
erikbosch, line 180: If we agree that float is a decent type for all datapoints related to speed, then we can discuss if we should keep "km/h" or if it would make sense to shift to "m/s" (SI-units) . OEMs and others anyway needs conversion capabilities, to e.g. being able to present speed as both km/h and miles/hour.

spec/Vehicle/Vehicle.vspec
```diff
@@ -175,12 +175,10 @@
...
-  description: Vehicle speed, as sensed by the gearbox.
+  description: Vehicle speed
```
erikbosch, line 181: We recently marked "Navigation.CurrentLocation.Speed" as deprecated with the rationale that it was a duplicate. But for that to be true Vehicle.Speed must be agnostic to how speed is sensed/calculated, i.e. we shall remove "as sensed by the gearbox" here?

spec/Vehicle/Vehicle.vspec
```diff
@@ -207,10 +205,8 @@
   description: Indicates whether the vehicle is stationary or moving
 
 - AverageSpeed:
-  datatype: int32
+  datatype: float
   type: sensor
-  min: -250
```
erikbosch, line 212: The proposal is to remove all specific limits concerning speed in the VSPEC standard. The current limits (-250/+250) does not relate to any industry standard (?), and actual limits will be vehicle dependent. If limits are needed/wanted, which e.g. could be useful for "CruiseControl.SpeedSet" then they should be defined in the VSPEC instantiation for that particular vehicle.

spec/Powertrain/Transmission.vspec
```diff
@@ -48,6 +48,7 @@
   max: 250
   unit: km/h
   description: Vehicle speed, as sensed by the gearbox.
+  deprecation: V2.1 removed because doubled with Vehicle.Speed
```
erikbosch, line 51: We recently marked "Navigation.CurrentLocation.Speed" as deprecated with the rationale that it was a duplicate. Should then this datapoint also be marked as duplicate.  (datatype/limits not changed for datapoints that are depricated)

spec/Vehicle/Vehicle.vspec
```diff
@@ -281,6 +277,7 @@
   type: attribute
   unit: s
   description: The time needed to accelerate the vehicle from a given start velocity to a given target velocity.
+  deprecation: V2.1 removed as ambiguous definition (start/stop-speed not defined)
```
erikbosch, line 280: Anyone using this attribute?

If it is supposed to be an attribute (with a constant value) then it needs to be better defined, e.g. by stating that start speed i 0 km/h and stop speed is 100 km/h. But the question is if an attribute like that is needed? And if so - is an attribute sufficient or would rather an RPC be needed (giving start and stop speed as parameters and getting time needed as return value)

My suggestion is to deprecate it so that we later can remove it. Better to have a limited set of well-defined datapoints than a big set of ambiguous ones.

spec/Vehicle/Vehicle.vspec
gunnarx: Agree, I see no need for the limits.

SebastianSchildt: I agree. Most min/max (e.g. speed) are deployment specific (might go into specific layer used for deployment, but not into the core speed)

spec/Vehicle/Vehicle.vspec
SebastianSchildt: I would vote to remove it (if keeping it, datatype also need to be float)

erikbosch: Remove it right away - or first mark it as deprecated? In this case it seems to be ill-defined and not used, so is it OK to remove it instantly, or should we always go via deprecation?

spec/Vehicle/Vehicle.vspec
erikbosch: Today we discussed GNSS-speed vs. speed as measured by the wheels, and in corner cases there could be big differences. For example, if your car is sliding down an icy road - then actual speed might be higher than wheel speed? If the car is tested on a "rolling road", GPS speed and wheel speed will be very different.

Do we see a benefit in adding detailed documentation on how each datapoint shall be interpreted, or do we prefer leaving that to each OEM? We could go "all in" and try to detail something like below, but I am not sure that we want that:

```
# The vehicle may use any method to provide this value (e.g. GPS, Radar, gearbox-sensor, wheel sensor)
# If possible, this sensor value shall represent the true longitudinal speed of the vehicle compared to the ground.
# It is acceptable that this is not always achievable for certain cases 
# Example 1: If the car is tested on a "rolling road", it is preferable that the datapoint reports 0 as speed, 
#                   but it is acceptable if a speed based on wheel or gearbox rotations are reported
# Example 2: If the car is parked on a moving ferry,  it is preferable that the datapoint reports 0 as speed, 
#                   but it is acceptable if a speed based on GPS is reported

```

danielwilms (approved):
Meeting: 2021-02-16 approved

Issue #231, opened 2021-01-28
Improper kPa and Pa units in release 2.0


rfrowe:
As per #206 and #217, all instances of `kpa` and `pa` have been changed to `kPa` and `Pa` respectively. This change was merged well before the 2.0 release of VSS.

However, as I look at the officially released 2.0 JSON [here](https://genivi.github.io/vehicle_signal_specification/releases/v2.0/vss_release_2.0.json), I see no instances of `kPa` and `Pa` and only instances of `kpa` and `pa`. My VSS tooling depends on the correct casing of these units and this release represents a regression that breaks my tooling.

gunnarx:
I found one instance of pa, and a few instances of UInt8 and UInt32 in the spec directory, so a few instances seem to have be missed.  Our build checks were also not running all build variants so for some reason it seems to have slipped through.  Some (hopefully all) of those fixes will be soon merged because I'm working to update and merge #220 

Related to this I however question if JSON should be "released", since it is a generated artifact from the original source code.

danielwilms:
handled in #232

Issue #232, opened 2021-02-02
Content of v2.0, maintenance branch (and branch strategy in the future)


gunnarx:
**1) Does something in this need to be adjusted?**

Commits in **origin/master** but not in **origin/release/2.0**:

* a95f464  danielwilms  7 days ago (authored 7 days ago)  Release 2.0 (#227)
* 891eeed  danielwilms  7 days ago (authored 7 days ago)  Add deprecation documentation (#223)
* a3aeaf3  Christian Heissenberger  2 weeks ago (authored 2 weeks ago)  "@" is a reseved symbol. possible bugfix, as gunnarx proposed
* c3c2e56  Christian Heissenberger  2 weeks ago (authored 7 weeks ago)  bugfixed commit for deprecation
* ef683c4  Christian Heissenberger  2 weeks ago (authored 8 weeks ago)  promoting global vehicle data/information to Vehicle root branch using deprecation
* 87155cc  Gunnar Andersson  7 weeks ago (authored 7 weeks ago)  vss-tools: Bump submodule to include units update
* 153cfbf  Ryan Rowe  8 weeks ago (authored 3 months ago)  Add carets to clarify powers
* 488a45a  Ryan Rowe  8 weeks ago (authored 3 months ago)  Rename units pa, kpa to Pa, kPa
* b35f4f3  Daniel (dakrbo)  9 weeks ago (authored 3 months ago)  Added various combustion engine properties
* 1565e2f  Daniel Wilms  2 months ago (authored 2 months ago)  Updated vss-tools for instance support
* 9c1b89b  SebastianSchildt  3 months ago (authored 3 months ago)  Fix/fix units again (#212)
* 6127e87  Sebastian Schildt  3 months ago (authored 3 months ago)  Bump HUGO version in travis
* d82d16f  Sebastian Schildt  3 months ago (authored 3 months ago)  Update Hugo theme version to work with newer Hugo releases

I'm wondering about the fact that there is a commit named "Release 2.0" on the master branch, but not on the Release branch, and that there are many commits on master also not on the release branch.  Should they be there?

**2) What's the overall plan for maintenance branches for 2.0 and beyond?**

I'd recommend we keep a maintenance branch for each.  Should it be called `/release/2.0` like now?
Another naming convention I have used in the past is `2.x` for the maintenance branch, *if* we will reuse it for all minor releases after 2.0.   It might be a reasonable way to avoid a maintenance branch for every minor release (too many).

[...]

PR #235, opened 2021-02-09
Updated release files

danielwilms:
Updated release branch as discussed in #232 and generated the new files. Those are added to the documentation.

Fixes: #231


SebastianSchildt:
Hopefully fixes #231   @rfrowe  can you check, if it looks fine for you, you seem to have better tooling 😁

rfrowe:
I diffed this against a 2.0 release that I fixed by hand. This has some signals that mine was missing but aside from that it looks the same. Should be good!

danielwilms:
Thanks for checking @rfrowe!

PR #237, opened 2021-02-10
Remove remaining value attributes

danielwilms:
Fixes: 198

### DIFF
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -103,7 +103,7 @@
 - Charging.ChargePortFlap:
   datatype: string
   type: actuator
-  value: closed
+  default: closed
   enum: [ 'open', 'closed' ]
   description: Signal indicating if charge port cover is open or closed, can potentially be controlled manually.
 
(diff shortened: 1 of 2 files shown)


spec/Powertrain/Battery.vspec
```diff
@@ -103,7 +103,7 @@
 - Charging.ChargePortFlap:
   datatype: string
   type: actuator
-  value: closed
+  default: closed
```
erikbosch, line 106: So "default" can be used for all types (actuators7sensors/attributes)? Does the documentation need to be updated, it seems to mention default only for attributes:
https://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/sensor_actuator/ 
https://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/attributes/

spec/Vehicle/Vehicle.vspec
```diff
@@ -304,56 +304,56 @@
...
-  value: 0
+  default: 0
   unit: kg
   description: Current overall Vehicle weight. Including passengers, cargo and other load inside the car.
 
 - CurbWeight:
   datatype: uint16
   type: attribute
-  value: 0
+  default: 0
```
erikbosch, line 314: Have there been any general discussion on when default values shall be used in the VSS standard? What is e.g. the purpose of having a default value here? I.e why is it better to report 0 as default value rather than giving an error if someone ask for the value but the car does not provide it.

(Note - I am not questioning the change in this PR, I am more thinking from a general perspective)

SebastianSchildt: If I remember there was some discussion around it and at least I was promoting the view, those things all belong to a deployment(layer) and not in the core spec, (and I think, but may be selective memory), there was a rough consensus here, however seeing that neither layers nor "deployment" workflow has been defined, the decision was in a first step make it consistent, i.e. only use either value or default for all such instances currently in the spec (and that is where we somehow failed, which is way the issue opened again 😿 ).

PR #242, opened 2021-02-17
Changing volume-related datapoints to float

erikbosch:
Also clarifying meaning of some "average" datapoints.
It can be noted that some OEM, like e.g. Mercedes
(https://developer.mercedes-benz.com/products/connect_your_fleet/details)
differentiate between "SinceReset" and "SinceStart".
While other like Volvo
(https://developer.volvocars.com/volvo-api/extended-vehicle/#Average_Fuel_Consumption)
does not specify how "average" is calculated.

### DIFF
--- a/spec/Powertrain/FuelSystem.vspec
+++ b/spec/Powertrain/FuelSystem.vspec
@@ -25,9 +25,8 @@
   description: Defines the hybrid type of the vehicle
 
 - TankCapacity:
-  datatype: uint16
+  datatype: float
   type: attribute
-  default: 0
   unit: l
   description: Capacity of the fuel tank in liters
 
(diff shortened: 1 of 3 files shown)


spec/Powertrain/FuelSystem.vspec
```diff
@@ -58,18 +57,21 @@
   unit: l/100km
   min: 0
   description: Average consumption in liters per 100 km.
+# The period used to calculate average consumption is OEM-specific
```
erikbosch, line 60: Do we want to have a better explanation on how the value shall be calculated, or do we want to keep that as OEM-specific. And if so, do we want to explicitly state that it is OEM-specific with a comment like here?

For example - Mercedes (https://developer.mercedes-benz.com/products/connect_your_fleet/details)
has two values: Driving.Consumption.Fuel.SinceStart and Driving.Consumption.Fuel.SinceReset

Volvo has just one value (https://developer.volvocars.com/volvo-api/extended-vehicle/#Average_Fuel_Consumption), not pecifying wther it is for current trip, since last reset or sine the car was produced.

danielwilms (approved):

danielwilms:
agreed 23/02/21

PR #248, opened 2021-03-03
Improving documentation for TravelledDistance

erikbosch:
Previously Vehicle.TravelledDistance and Transmission.TravelledDistance had the same description. In VSS meeting it was discussed whether they could be considered as duplicates or if they actually have different meaning. It was agreed that they might be different, if the transmission has been replaced you might want to keep track of the wear/usage of the new/replacement transmission separately.

### DIFF
--- a/spec/Powertrain/Transmission.vspec
+++ b/spec/Powertrain/Transmission.vspec
@@ -52,13 +52,19 @@
 
 
 #
-# Current odometer reading
+# Current odometer reading for the transmission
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
While I still think this is a useless value for the __core__ data model, I think it is good, that with this change it isn't a confusing duplicate anymore.

danielwilms:
09/03/21

PR #253, opened 2021-03-17
Using float for OBD data

### LINKED ISSUE #250: Unit and Description of OBD signals
When reviewing the proposed Android to VSS mapping ( https://at.projects.genivi.org/jira/browse/AASIG-122) i realized that our OBD.vspec is a bit odd. According to https://en.wikipedia.org/wiki/OBD-II_PIDs there are a lot of conversions going on. If we take the two examples below

- EngineSpeed (0C) is sent as uint16, to get the "real value" take (256*A + B)/4
- OilTemperature (5C) is sent as uint8, to get the "real value" take A - 40

And this is not obvious in our current descriptions: 

```
- EngineSpeed:
  datatype: float
  type: sensor
  unit: rpm
  description: PID 0C - Engine speed measured as rotations per minute

- OilTemperature:
  datatype: uint8
  type: sensor
  unit: celsius
  description: PID 5C - Engine oil temperature
```

For EngineSpeed we state what the real value can be (a float, but actually not all floats as we just can support quarters). For OilTemperature we state what is sent (a uint8). Just from looking at OilTemperature above it is not obvious that you can send negative temperatures.  

I can see two approaches for OBD-signals:

**Represent the real value**

I.e. do not care how OBD send it. With this approach 

- OilTemperature shall be an int16 with min: -40 and max +210. Unit shall be celsius.
- EngineSpeed shall be a float with min: 0 and max: 16,383.75. Unit shall be rpm. 

**Represent the value transmitted by OBD**

I.e. do not care what value it represent. With this approach 

- OilTemperature shall be a uint8. 
- EngineSpeed shall be a uint16. 

With this approach I suggest that celsius/rpm shall be removed from the signals. As of now VSS does not have any syntax for scaling/offset. So if we say "celsius+uint8" I assume that I do not need to do any scaling/offset. I.e. in practice - OBD is using a custom unit - not the unit "celsius" standardized by VSS. Of course we can include the conversion as a comment or in the description, and mention celsius in a comment.

As of now it seems that we have a mix of the two approaches. I believe this needs to be corrected!

erikbosch:
Fixes #250

Previously datatypes were inconsistently used in OBD.vspec.
E.g use of an int-type even if OBD support fractions or using an
uint-type even if OBD supports negative values.

With this commit all PIDs representing measurement or count use float.
This makes representation independent from size/range/scaling/offset
used by OBD. Also removing limits where specified to get a consistent representation.

### DIFF
--- a/spec/OBD/OBD.vspec
+++ b/spec/OBD/OBD.vspec
@@ -18,6 +18,19 @@
 # Some of these signals are also available through other nodes in the
 # VSS tree.
 #
+# OBD often use scaling and offset, and has specified limits for each PID
+# (see e.g. https://en.wikipedia.org/wiki/OBD-II_PIDs).
+# In VSS signals represent the real value, the actual encoding used by OBD is not considered.
+# Limits specified by OBD are not explicitly stated in VSS
+# i.e. a VSS OBD signal may theoretically have a value that can not be transferred by OBD.
+#
+# Example: Timing Advance (PID 0E) can in OBD support the range from -64 degrees to +63.5 degrees.
+#          In OBD the value is transmitted as a uint8, to get the real value one must take the
+#          uint8 value, divide by 2 and subtract 64.
+#          E.g. +4.5 degrees is in OBD transmitted as (4.5+64)*2 = 137.
+#
+#          In VSS the signal contains the actual value, i.e. +4.5 degrees is sent as +4.5.
+#
 
 - PidsA:
   datatype: uint32
@@ -34,7 +47,7 @@
   description: Malfunction Indicator Light (MIL) False = Off, True = On
 
 - Status.DTCCount:
-  datatype: uint32
+  datatype: uint8
   type: sensor
   description: Number of sensor Trouble Codes (DTC)
 
@@ -60,11 +73,9 @@
   description: PID 03 - Fuel status
 
 - EngineLoad:
-  datatype: uint8
+  datatype: float
   type: sensor
   unit: percent
-  min: 0
-  max: 100
   description: PID 04 - Engine load in percent - 0 = no load, 100 = full load
 
 - CoolantTemperature:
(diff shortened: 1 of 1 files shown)


spec/OBD/OBD.vspec
```diff
@@ -74,49 +85,39 @@
   description: PID 05 - Coolant temperature
 
 - ShortTermFuelTrim1:
-  datatype: int8
+  datatype: float
   type: sensor
   unit: percent
-  min: -100
-  max: 100
```
erikbosch, line 81: We previously only had limits in a few places, and they were not always correct. Like here where the max limit actually is 99.21875 due to the scaling used.

spec/OBD/OBD.vspec
```diff
@@ -34,7 +47,7 @@
   description: Malfunction Indicator Light (MIL) False = Off, True = On
 
 - Status.DTCCount:
-  datatype: uint32
+  datatype: float
```
SebastianSchildt, line 50: This really is a "count" as in "number of errors. No "sensor", Wouldn't it make more sense to keep this as some sort of int?

erikbosch: Yes we could have it as int. Then possibly uint8 as only seven bits are used in OBD (see https://en.wikipedia.org/wiki/OBD-II_PIDs#Service_01_PID_01)

spec/OBD/OBD.vspec
```diff
@@ -562,7 +561,7 @@
   description: Malfunction Indicator Light (MIL) - False = Off, True = On
 
 - DriveCycleStatus.DTCCount:
-  datatype: uint32
+  datatype: float
```
SebastianSchildt, line 564: Maybe also int

spec/OBD/OBD.vspec
```diff
@@ -465,31 +464,31 @@
...
 - FuelLevel:
-  datatype: uint8
+  datatype: float
   type: sensor
   unit: percent
   description: PID 2F - Fuel level in the fuel tank
 
 - WarmupsSinceDTCClear:
-  datatype: uint16
+  datatype: float
```
SebastianSchildt, line 491: also int?

spec/OBD/OBD.vspec
```diff
@@ -34,7 +47,7 @@
   description: Malfunction Indicator Light (MIL) False = Off, True = On
 
 - Status.DTCCount:
-  datatype: uint32
+  datatype: uint8
```
erikbosch, line 50: As discussed in previous VSS-meeting  - keeping int-based values for PID representing counts. But then it makes sense to select the "right" VSS size. In this case OBD use 7 bits (see https://en.wikipedia.org/wiki/OBD-II_PIDs#Service_01_PID_01) so uint8 shall suffice, no need to have uint32

spec/OBD/OBD.vspec
```diff
@@ -465,31 +464,31 @@
...
 - FuelLevel:
-  datatype: uint8
+  datatype: float
   type: sensor
   unit: percent
   description: PID 2F - Fuel level in the fuel tank
 
 - WarmupsSinceDTCClear:
-  datatype: uint16
+  datatype: uint8
```
erikbosch, line 491: According to https://en.wikipedia.org/wiki/OBD-II_PIDs PID 30 use 1 byte and supports values 0-255 so uint8 shall be sufficient.

SebastianSchildt (approved):

PR #254, opened 2021-03-17
Removing aggregate keyword

### LINKED ISSUE #246: Meaning of "aggregate" for branches?
In https://genivi.github.io/vehicle_signal_specification/rule_set/branches/ we describe aggregate. Well we do not describe it that well:

`Defines whether or not this branch is an aggregate. If not defined, this defaults to false.`

It is used in a few cases where data logically "belongs together". But do we know of any tooling actually relying on "aggregate" in any kind? Or is it just handled as informal information? I think it would be good to improve the documentation on "aggregate" and I can create a PR for that, but then I first need to understand the background for it, and make sure that we still consider it useful (otherwise we can deprecate it)

Example usage:
```
- CurrentLocation:
  type: branch
  aggregate: true
  description: The current latitude and longitude of the vehicle.
```

erikbosch:
Fixes #246

After this has been merged,  a separate PR for vss-tools
shall be merged to give error if keyword "aggregate" is used.

### DIFF
--- a/docs-gen/content/rule_set/branches.md
+++ b/docs-gen/content/rule_set/branches.md
@@ -12,7 +12,7 @@ A branch entry example is given below:
 ```YAML
 - Trunk:
   type: branch
-  aggregate: true
+  aggregate: false
   description: All signals related to the rear trunk
 ```
 
@@ -30,6 +30,10 @@ opposed to a signal entry). This is the default, in case ```type``` is omitted.
 * **```aggregate```** *[optional]*  
 Defines whether or not this branch is an aggregate.
 If not defined, this defaults to ```false```.
+An aggregate is a collection of signals that make sense to handle together in a system.
+A typical example could be GNSS location, where latitude and longitude make sense to read
+and write together. This is supposed to be deployment and tool specific,
+and for that reason no branches are aggregates by default in VSS.
 
 * **```description```**  
 A description string to be included (when applicable) in the various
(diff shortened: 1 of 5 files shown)


SebastianSchildt:
Discussed March 23rd: "aggregate" might be a good deployment information for a layer, but maybe not needed for the core model

So merge this @gunnarx  and close #246

docs-gen/content/rule_set/branches.md
```diff
@@ -27,10 +26,6 @@ well.
 The value ```branch``` specifies that this is a branch entry (as
 opposed to a signal entry). This is the default, in case ```type``` is omitted.
 
-* **```aggregate```** *[optional]*  
-Defines whether or not this branch is an aggregate.
```
erikbosch, line 31: @gunnarx If we as stated in #256 wants to keep aggregate for use in layers, then should we then not keep it in this file? Possibly with a remark like. "Intended for the future layer concept and not used in VSS core model"

SebastianSchildt: I would remove it completely (spec and doc)

docs-gen/content/rule_set/branches.md
```diff
@@ -30,6 +30,10 @@ opposed to a signal entry). This is the default, in case ```type``` is omitted.
 * **```aggregate```** *[optional]*  
 Defines whether or not this branch is an aggregate.
 If not defined, this defaults to ```false```.
+An aggregate is a collection of signals that make sense to handle together in a system.
```
erikbosch, line 33: Documentation updated, as discussed last week

SebastianSchildt (approved):

danielwilms:
merged in 06/04/21

PR #255, opened 2021-03-17
Running vss-tools tests also from vss travis

erikbosch:
We recently added a "tests" target in the vss Makefile and updated .travis.yml to run that target for vss-tools PRs, but we didn't add it to .travis.yml in vehicle-signal-specification.

With this change we run "tests" also for PRs like this one

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,7 @@ install:
   - cd ..
 
 script:
-  # NOTE: c not built -- fix later
-  - make clean json franca binary csv deploy
+  - make travis_targets
   - hugo -s ${DOCS_DIR}
 
 deploy:
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,14 @@
 # Makefile to generate specifications
 #
 
-.PHONY: clean all json franca c
+.PHONY: clean all json franca csv binary c tests travis_targets
 
 all: clean json franca csv binary c tests
+
+# All targets that shall be built on each pull request for
+# vehicle-signal-specification or vss-tools
+# NOTE: c currently not built by travis - does not support latest vss
+travis_targets: clean json franca binary csv tests deploy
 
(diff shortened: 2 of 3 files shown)


erikbosch:
@danielwilms @gunnarx what is our approach to updating the vss-tools submodule version? Here I noticed when adding "tests" to the travis file that the tests did not pass, as we used an old vss-tools. So as of now - this PR include an update of the vss-tools version.

But I do not know if that is the approach we want - that you need to manually bump vss-tools version when you see a need.

.travis.yml
```diff
@@ -19,8 +19,7 @@ install:
   - cd ..
 
 script:
-  # NOTE: c not built -- fix later
-  - make clean json franca binary csv deploy
+  - make travis_targets
```
erikbosch, line 22: Later we can change also .travis.yml in vss-tools. when that is done we only need to change on file (makefile in vss) to change which targets to build for travis tests for both vss and vss-tools

Makefile
```diff
@@ -33,7 +38,7 @@ c:
 	${TOOLSDIR}/vspec2c.py -i:./spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).h vss_rel_$$(cat VERSION)_macro.h
 
 clean:
-	rm -f vss_rel_$$(cat VERSION).json vss_rel_$$(cat VERSION).fidl vss_rel_$$(cat VERSION).binary vss_rel_$$(cat VERSION).csv vss_rel_$$(cat VERSION).h
+	rm -f vss_rel_*
```
erikbosch, line 41: In the deploy target below we copy everything, so why no remove everything starting with vss_rel_* here as well

SebastianSchildt:
March 23rd: Let's merge
@gunnarx

PR #259, opened 2021-03-30
vss-tools & Makefile: include ocf, protobuf

gunnarx:
This is a tools bump to include the most recent additions to vss-tools

Should be straight forward, therefore SHORT REVIEW PERIOD! :) 

Shortlog:

  - c29206f Move vspec2protobuf to contrib/
  - 413c74e Add vspec2protobuf, using anytree
  - 1076588 Go build script and Go parser update.
  - 4dbc6e0 Move ocf related tools and create contrib/ dir
  - e075559 vspec2ocf: Fix printfs to be python3 compliant
  - e905aab Updates to binary tool

Some issues missed in previous merges: Bugfix the protobuf target in Makefile and include target for ocf. (OCF tooling is unmaintained but some fix included so might as well have the Makefile target for further testing)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,9 @@
 # Makefile to generate specifications
 #
 
-.PHONY: clean all json franca csv binary c tests travis_targets
+.PHONY: clean all travis_targets json franca csv tests binary protobuf ocf c install deploy
 
-all: clean json franca csv binary c tests
+all: clean json franca csv binary tests protobuf
 
 # All targets that shall be built on each pull request for
 # vehicle-signal-specification or vss-tools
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,9 +2,9 @@
 # Makefile to generate specifications
 #
 
-.PHONY: clean all json franca csv binary c tests travis_targets
+.PHONY: clean all travis_targets json franca csv tests binary protobuf ocf c install deploy
 
-all: clean json franca csv binary c tests
+all: clean json franca csv binary tests protobuf
```
erikbosch, line 7: Is protobuf supposed to work as of today? (I got a build error, but that could be on my system only)
If it is supposed to be working, I suggest to add it to "travis_targets" as well so that it is included in travis tests automatically

Makefile
```diff
@@ -33,6 +33,12 @@ binary:
 	gcc -shared -o ${TOOLSDIR}/binary/binarytool.so -fPIC ${TOOLSDIR}/binary/binarytool.c
 	${TOOLSDIR}/vspec2binary.py -i:./spec/VehicleSignalSpecification.id ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).binary
 
+protobuf:
+	${TOOLSDIR}/contrib/vspec2protobuf.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).proto
+
+ocf:
```
erikbosch, line 39: My understanding is that ocf does not work as of today, if so i suggest extending the comment on line 11 to state that both c and ocf are deliberately excluded from travis builds

Makefile
erikbosch: On the other hand - protobuf/ocf are in contrib directory so maybe not make sense to require that they shall pass

erikbosch: I experimented a little bit with "optional build targets" in #260. If we think it is a good idea that we try to run c, ocf and protobuf on every PR just to track their status, but still wants Travis to ignore whether they fail, then that could be a solution. But this is something we can do later, I think this PR is good to go.

erikbosch (approved):
A bit unclear to me if the ocf/protobuf commands are supposed to work as of today, but adding the make commands is a good start

gunnarx:
Updated to include latest commit to vss-tools (ae90436 Go module script update)

PR #261, opened 2021-04-01
Adding service signals

### LINKED ISSUE #244: Service Indicator datapoints
In #180 ACEA list a lot of datapoints related to service, but with no extra information on any of them:

```
7. Service Indicator
8. Distance to next service
9. Time to next service
25. Distance until Engine Oil Service
28. Remaining Brake Fluid Life
29. Distance Until Brake Service
30. Due date for Brake Fluid Change
32. Time until next required oil change
```

VSS has only one attribute related to service:

`Vehicle.Powertrain.CombustionEngine.OilLifeRemaining`

Other known APIs typically include datapoints related to service, see references below. It seems to be quite common to have at least 2 datapoints, time and range to next service. The other datapoints listed by ACEA above does not seem to be commonly used in public APIs. It is also possible that OEMs might have other datapoints or service intervals as well (e.g. for air filter change, HVAC filter change, transmission oil change, coolant change, timing belt change, spark plug replacement, airbag inspection, ...)

Topics to discuss: 

- I suggest that we at least add two datapoints, initial suggestion: “Vehicle.TimeToService” (seconds?, float?) and “Vehicle.DistanceToService” (meter?, float?) to VSS. 
- But do we want to add others as well, or wait until we see that they really are needed? I do not know what values vehicles typically maintain internally. (But nothing says that the datapoint needs to be maintained "inside" the car)
- Shall we allow negative values (to indicate service overdue by X seconds/meter)
- Do we need a separate “service indicator” ? Or use an implicit definition “(TimeToService<=0) OR (DistanceToService<=0)). 
- I suggest that we deprecate Vehicle.Powertrain.CombustionEngine.OilLifeRemaining. Alternatively change it to being a sensor value (it is as of now an attribute)

Volvo: https://developer.volvocars.com/volvo-api/extended-vehicle/ 

```
kmToService (km), 
monthsToService (months, integer), 
serviceWarningStatus (NORMAL, ALMOST_TIME_FOR_SERVICE, TIME_FOR_SERVICE, TIME_EXCEEDED), 
serviceWarningTrigger (CALENDAR_TIME, DISTANCE, ENGINE_HOURS, ECM)
```

Mercedes: https://developer.mercedes-benz.com/products/connect_your_fleet/details 

```
Service.Interval.Distance (Residual distance to service, km), 
Service.Interval.Time (Residual time to service, days)
```

erikbosch:
Fixes #244

This is a proposal for discussion on coming VSS meetings. 

Current proposal is to add a boolean signal (basically equal to the the service lamp) and two signals to indicate remaining time/distance to service. That would give support for both vehicles that want to communicate remaining time/distance as well as for vehicle that only want to communicate a boolean value. 

More discussions in the issue

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -58,10 +58,12 @@
   type: attribute
   description: Engine compression ratio.
 
+# In addition to this a signal a vehicle can report remaining time to service (including e.g. oil change)
+# by Vehicle.Service.TimeToService
 - OilLifeRemaining:
-  datatype: uint32
-  type: attribute
-  description: Remaining engine oil life in seconds.
+  datatype: int32
+  type: sensor
+  description: Remaining engine oil life in seconds. Negative values can be used to indicate that lifetime has been exceeded.
   unit: s
 
 - EngineOilCapacity:
--- a/spec/Vehicle/Service.vspec
+++ b/spec/Vehicle/Service.vspec
@@ -1,0 +1,52 @@
+#
+# (C) 2021 Robert Bosch GmbH
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# Service data
+#
+# This file contains signals relevant for service/maintenance of the vehicle,
+# giving information on when service is needed.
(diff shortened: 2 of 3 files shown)


SebastianSchildt:
I like putting the general service-related stuff stuff in a services branch.

danielwilms (approved):

PR #263, opened 2021-04-07
Temporary change of Makefile

erikbosch:
This is to make it possible to move vspec2c folder in vss-tools
When moved the Makefile shall be changed to refer to the new location


danielwilms:
not needed

PR #269, opened 2021-04-22
Improving documentation on instances

### LINKED ISSUE #267: Instance-level Information vs. Vehicle-level Information
Have we discussed or stated how a vehicle preferably shall handle instance-signals if it does not have multiple instances of that particular feature?

Example: A vehicle can currently report the fluid level for brake fluid for each individual wheel. But many vehicles do not have individual fluid containers for each wheel. Passenger cars often have only one container, motorbikes often have two containers, heavy trucks might have air-driven breaks and thus no container at all.

With the current VSS, I can see two alternatives for a vehicle with only one fluid container:

- The vehicle reports the same fluid level for all 4 wheels
- The vehicle only report fluid level for one wheel (e.g. front left), and does not report fluid level for the other. 

A similar problem occurs for actuators like `FanSpeed`. Assume a vehicle where you can select different temperature for left and right front seat, but they share the same fan-speed. Shall the vehicle then allow the fan-speed to be changed both by `Vehicle.Cabin.HVAC.Station.Row1.Left.FanSpeed` and `Vehicle.Cabin.HVAC.Station.Row1.Right.FanSpeed`, giving the impression that they can be controlled individually, or should the vehicle only expose one of them?

erikbosch:
This is an idea on how we could improve documentation/guidelines on how e.g. an OEM about to support VSS for a vehicle shall think and act concerning instance declarations in VSS. It is not straightforward, as we in e.g. VSS does not have a way to state whether e.g. rear windshield has a washer at all. In the future we may possibly refer to VSSo and state something like:
`
That a signal is present in a generated vss tree does not mean that it needs to be supported by a vehicle. Use e.g. VSSo to statically or dynamically describe which signals that the vehicle supports!`

For now, consider the changes in `instances.md` as a proposal and let us discuss at a meeting if we want to have guidelines like this in our documentation.

Fixes #267


danielwilms (approved):

danielwilms:
2021/05/11

Issue #180, opened 2020-08-25
Support ACEA Level A1 Proposed Data List


SebastianSchildt:
Industry is discussing, what kind of data a vehicle should be able to provide, and who should be able to access waht and where.

In VSS we want to create technology to support use cases from different players. [ACEA](https://www.acea.be) has a list of 42 basic data items, which should be able to be expressed in VSS, if within scope

Here is the list for A1 level access:

[acea42Listb.pdf](https://github.com/GENIVI/vehicle_signal_specification/files/5123874/acea42Listb.pdf)

For reference, this is ACEAs view on data different tiers of data access

- **Level A** data in the vehicle is available on the OEM backend via the Extended Vehicle web services interface. This can be accessed by a 3rd party either via an OEM server or via a Neutral Server. The customer has provided consent to the OEM for the transfer of level A data from their vehicle to the OEM server. Further consent is required from the customer for the OEM to transfer level A data from the OEM server to their chosen 3rd party.
-  **Level A** data is partitioned into a predefined data set **(A1)** and additional data used by an OEM **(A2)** for the generation of their own remote services. The predefined data set is data provided by all vehicle manufacturers provided that the vehicle in question is equipped to deliver the data and the data is available on the OEM server. Level A data can be accessed at the sampling frequency implemented and used by the OEM.
- **Level B** data is data available on a vehicle interface but has not yet been transferred to an OEM server. OEMs can only transfer data for specific purposes, with the consent of the customer. For a third party to access level B data, a B2B agreement will be required with the OEM. The data will be transferred specifically for their usage. The data can then be channeled to the third party via the OEM web interface or a Neutral Server, provided the vehicle owner has provided consent for the data transfer.
- **Level C** data is data that is generated by routines hosted in a vehicle’s electronic control unit (ECU). The routines generated by third parties can be used to aggregate and process data on-board. A B2B contract will be required between the OEM and third parties wishing to develop, integrate, test and deploy routines. Access to vehicle resources and the transfer of data generated by third party routines will be governed within the B2B contract.

klotzbenjamin:
Regarding the A1 list you provided, I quickly looked into what VSS can describe:

1. Vehicle.Speed
2. Vehicle.Acceleration.Longitudinal
3. Vehicle.TravelledDistance
4. Multiple OBD signals referring to DTC (e.g. Vehicle.OBD.Status.DTCCount), but not DTC branch
[...]

PR #271, opened 2021-04-27
Adding signal for vehicle engine oil level

erikbosch:
This is a follow up to #180 to cover the need for a signal on oil level

Having oil level seems to be quite common in APIs, see examples below. Many use enums, some use percent plus boolean indicator. The proposed change here is reasonably aligned with the Volvo and Android representation.

In my view a "subjective" enum makes more sense than an "objective" percentage value, as percentage value might be difficult to interpret as it might be very vehicle dependent. Assume for example an engine with a nominal oil capacity (EngineOilCapacity) of 5 liters. Then oil level might be "critically low" already at 70% (if we consider 0% to mean a vehicle without any oil at all). Also, how sensitive a vehicle is for overfilling with oil might vary.

The intention with the proposed enum comments is to let "normal" indicate that oil is within expected range. I.e if you just have had an oil change and the garage has added EngineOilCapacity liters of oil, then the OilLevel shall be "normal", not "high". I.e. low and high shall indicate something that needs to be addressed, although not critically.

Volvo (https://developer.volvocars.com/volvo-api/extended-vehicle/#Oil_Level)

```
VERY_LOW
LOW
NORMAL
HIGH
VERY_HIGH
```

Android (https://android.googlesource.com/platform/hardware/interfaces/+/master/automotive/vehicle/2.0/types.hal)

```
enum VehicleOilLevel : int32_t {
    /**
     * Oil level values
     */
    CRITICALLY_LOW = 0,
    LOW = 1,
    NORMAL = 2,
    HIGH = 3,
    ERROR = 4,
};

```

Mercedes (https://developer.mercedes-benz.com/products/connect_your_fleet/details)

```
InternalCombustionEngine.Oil.Level | percent | Oil level | onInterval
-- | -- | -- | --
InternalCombustionEngine.Oil.Level.Warning |   | Indicates whether an oil level   warning is triggered true: ok false: oil level is overfilled or underfilled | onInterval and onChange
```

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -58,6 +58,18 @@
   type: attribute
   description: Engine compression ratio.
 
+- EngineOilLevel:
+  datatype: string
+  type: attribute
+  enum: [ 
+    "critically_low",  # Critically low, immediate action required
+    "low",             # Level below recommended range, but not critical
+    "normal",          # Within normal range, no need for driver action
+    "high",            # Level above recommended range, but not critical
+    "critically_high"  # Critically high, immediate action required
+    ]
+  description: Vehicle oil level.
+
 # In addition to this a signal a vehicle can report remaining time to service (including e.g. oil change)
 # by Vehicle.Service.TimeToService
 - OilLifeRemaining:


erikbosch:
Updated according to discussion April 27th:

- "very" replaced with "critically"
- "error" removed, error handling is outside the scope of VSS

danielwilms (approved):

gunnarx (approved):

PR #282, opened 2021-05-26
Updating specification for wheels and tires

### LINKED ISSUE #41: Need to support fraction of inches?
Is there a need to support fraction of inches? At least for e.g. buses it is not uncommon with wheel sizes like 22.5" x 8.25".  If needed one could consider a representation like "milli-Inches", e.g. "22500" for a 22.5" wheel

Is the standardisation organisation not called ETRTO (ETRO in current spec)
http://www.etrto.org/Home

https://github.com/GENIVI/vehicle_signal_specification/blob/222bbc6d1900e2c82dc15138da2266dcdae3c53e/vss_rel_1.0.fidl#L139

erikbosch:
Fixes #41

Intentions with this change:

- Allow fraction of inches for rim/wheel size. It is e.g. quite common that trucks has 22.5 inches as rim size
- For most modern trucks and passenger cars the tire width is given in millimeters (e.g. 225 mm) so proposing change of tire width from inches to millimeters. 
- (Outer) TireDiameter is typically not specified for modern trucks and passenger cars. Instead aspect ratio is used (e.g. 195/60R15 where 60 is the aspect ratio). Even if aspect ratio can be calculated from other attributes I suggest adding it. Suggesting to keep TireDiameter however, as that can be handy for vehicles that use flotation/numeric tire sizes.

### DIFF
--- a/spec/Chassis/Chassis.vspec
+++ b/spec/Chassis/Chassis.vspec
@@ -100,34 +100,44 @@
   description: Number of wheels on the axle
 
 - Axle.WheelDiameter:
-  datatype: uint8
+  datatype: float
   type: attribute
   unit: inch
-  description: Diameter of wheels (without tires), in inches, as per ETRTO / TRA standard.
+  description: Diameter of wheels (rims without tires), in inches, as per ETRTO / TRA standard.
 
 - Axle.WheelWidth:
-  datatype: uint8
+  datatype: float
   type: attribute
   unit: inch
-  description: Width of wheels (without tires), in inches, as per ETRTO / TRA standard.
+  description: Width of wheels (rims without tires), in inches, as per ETRTO / TRA standard.
 
 
 #
 # Tire attributes
 #
+# Tire size can be specified by different systems
+# The VSS signals are intended to support both ISO metric tire code and flotation/numeric sizes.
+# Note that tires typically specify either tire diameter or aspect ratio, but both are included in VSS
+# for convenience, and it is possible for a vehicle to present both attributes as they can be calculated from each other.
+# Axle.TireDiameter = Axle.WheelDiameter + ((2 * Axle.TireWidth * Axle.TireAspectRatio) /(100*25.4))
 
 - Axle.TireDiameter:
-  datatype: uint8
+  datatype: float
   type: attribute
   unit: inch
-  description: Diameter of tires, in inches, as per ETRTO / TRA standard.
+  description: Outer diameter of tires, in inches, as per ETRTO / TRA standard.
 
 - Axle.TireWidth:
+  datatype: uint16
+  type: attribute
+  unit: mm
+  description: Nominal section width of tires, in mm, as per ETRTO / TRA standard.
+
+- Axle.TireAspectRatio:
   datatype: uint8
   type: attribute
-  unit: inch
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
Better than before 👍   Two questions/comments:

 - I like the more extensive documentation, regarding the "ETRTO / TRA standard", is there also any "official" link we can give? 
 - I know it is convention, and that in some areas legacy units sticking around, but it always hurts me dealing with non-SI units. But I guess VSS is not the place where we can change that...

erikbosch:
The problem with stating "ETRTO/TRA standard" is that they are not freely available. They might also in turn refer to an ISO standard (also not freely available). But we could if we like to add the links where you could order the publications: 

http://www.us-tra.org/publications.html 
https://www.etrto.org/Publications/Order
https://www.iso.org/ics/83.160.10/x/

But to just understand what measurements that are commonly used and printed on tires free resources could be used, and at least the wikipedia one should be quite stable:

https://en.wikipedia.org/wiki/Tire_code
https://generaltire.com/our-world/latest-news/making-some-sense-of-metric-and-american-standard-tire-sizing
https://www.goodyearautoservice.com/en-US/tire-basics/tire-size
https://www.bridgestoneamericas.com/en/corporate-social-responsibility/safety/choosing-tires/determining-tire-size

And concerning SI units - BMW and Michelin tried to introduce metric sized rims long time ago (with very limited success, see https://www.hemmings.com/stories/article/metric-masterpieces).

SebastianSchildt (approved):

erikbosch:
At the meeting I was proposed to give some more reasoning for the proposal. We can look at specifications for some different vehicles, both American and European.

If you want to buy new tires for your Lincoln Navigator 2021, the standard information given at https://www.wheel-size.com/size/lincoln/navigator/2021/ is (for the default tire):

- Tire 275/55R20 113H
- Rim 8.5Jx20 ET44

For a Volvo XC90 2020 we get this information (from https://www.volvocars.com/mt/support/manuals/xc90/2020w17/specifications/specifications-for-wheels-and-tyres/approved-wheel-and-tyre-sizes) for the smallest wheel:

- Tire 235/60R18
- Rim 8x18x42.5

Similar for a BMW R1250GS motorcycle (https://www.bmwmotorcycles.com/en/models/adventure/r1250gs/technicaldata.html#/section-technical-data)

- Tire Front: 120/70R19, Rear: 170/60R17
- Rim Front: 3.00x19, Rear 4.5x17

From this it can be seen that is industry standard to specify size of rim in inches. and width in millimetres. If we take the front wheel of the BMW R1250GS and translate to VSS signals according to this proposal the contents would be:

- Axle.WheelDiameter: 19 inches
- Axle.WheelWidth: 3 inches
- Axle.TireDiameter: Not specified for any of the vehicles above
- Axle.TireWidth: 120 millimeters
- Axle.TireAspectRatio: 70%

Even in e.g. agriculture metric numbers for width seems to become standard. 

http://salesmanual.deere.com/sales/salesmanual/en_NA/tractors/2012/feature/wheels_and_tires/7/understanding_basic_tire_info.html?sbu=ag&link=prodcat

`Axle.TireDiameter` is typically never specified, possibly with the exception of some light truck flotation tires.

PR #286, opened 2021-05-28
Correcting color comment

erikbosch:
Previous comment must be a copy paste error.
Now using comment from https://schema.org/Car

It can by the way be discussed to change `interior` to `Interior` (capital I) to be consistent with capitalization in schema.org where the name is e.g. `vehicleInteriorColor`


danielwilms:
meeting 01/06/21

PR #290, opened 2021-06-01
Makefile: Add target for graphql

gunnarx:
Note that this also updates the vss-tools submodule to include the
contrib/vspec2graphql.py, so that the new target can also be tested by CI.

If the graphql tool gets merged first (fast-forward), then the submodule hash will match and this can be merged directly.  If not, the submodule needs update in this PR before merging it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,9 @@
 # Makefile to generate specifications
 #
 
-.PHONY: clean all travis_targets json franca csv tests binary protobuf ttl ocf c install deploy
+.PHONY: clean all travis_targets json franca csv tests binary protobuf ttl graphql ocf c install deploy
 
-all: clean json franca csv binary tests protobuf
+all: clean json franca csv binary tests protobuf graphql
 
 # All mandatory targets that shall be built and pass on each pull request for
 # vehicle-signal-specification or vss-tools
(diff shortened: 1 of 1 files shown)


erikbosch (approved):
Cool, did not knew that it was sufficient to have a PR for the include submodule version for travis to find it. Looks good.

danielwilms:
fix build, then merge > meeting 08/06/21

gunnarx:
> Cool, did not knew that it was sufficient to have a PR for the include submodule version for travis to find it. Looks good.

It's somewhat tricky, but yes.   A submodule is like a pointer to a repo and to a specific commit hash.  Travis makes a clean checkout (and recursive, so including the submodules).  

It is possible to include a change of submodule in the single commit as done here.   Thus a PR, or commit, can usually make things work even when changes are made in repos that are dependent on each other. 

NOTE:  When working with forks it's easy to make the mistake that you let the submodule refer to a hash that was never pushed to the repo address that the submodule also points to (it only exists locally, or on another fork).

gunnarx:
Rebased on master to fix conflicts.

gunnarx:
> fix build, then merge > meeting 08/06/21

Done, merging.

PR #301, opened 2021-06-18
Improving attribute documentation

### LINKED ISSUE #295: How often are attributes allowed to change?
A vehicle might have some attributes that may change during lifetime. For example, the owner might change tire size, or gross weight might be changed for tax or driving license reasons. This is typically changes that happens at most a few times a year.

Does this fit well with our current view on attributes, that they e.g. can be specified when the vspec file is installed (see e.g. https://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/attributes/)? Or should they rather be sensors if they might change during the lifetime of the vehicle.

erikbosch:
Also making sure example is aligned with current vpsec definition
and remove double blanks in vspec for related signals.

Fixes #295

### DIFF
--- a/docs-gen/content/rule_set/data_entry/attributes.md
+++ b/docs-gen/content/rule_set/data_entry/attributes.md
@@ -7,17 +7,18 @@ weight: 5
 An attribute is a signal with a default value, specified by
 its ```default``` member.
 
-The value set for an attribute by a vspec file can be read by a
-consumer without the need of having the value sent out by a
-publisher. The attribute, in effect, is configuration data that
-can be specified when the vspec file is installed.
+Attribute values can also change, similar to sensor values.
+The latter can be useful for attribute values that are likely to change during the lifetime of the vehicle.
+However, attribute values should typically not change more than once per ignition cycle,
+or else it should be defined as a sensor instead.
 
 Below is an example of a complete attribute describing engine power
 
 ```YAML
 MaxPower:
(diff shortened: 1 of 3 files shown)


docs-gen/content/rule_set/data_entry/attributes.md
```diff
@@ -7,17 +7,20 @@ weight: 5
...
-The value set for an attribute by a vspec file can be read by a
-consumer without the need of having the value sent out by a
-publisher. The attribute, in effect, is configuration data that
-can be specified when the vspec file is installed.
+The value set for an attribute by a vspec file can be read by a consumer
+without the need of having the value sent out by a publisher.
+Attribute values can also be provided by publishers, similar to sensors.
+The latter can be useful for attribute values that are likely to change during the lifetime of the vehicle.
+However, this change should typically not be more than once per ignition cycle,
+or else it should be defined as a sensor instead.
```
danielwilms, line 16: stable in power-cycle

danielwilms (changes requested):

docs-gen/content/rule_set/data_entry/attributes.md
gunnarx: Possibly drive-cycle? Let's investigate and agree on the right term.   Keeping it open one more week.

danielwilms (approved):

PR #303, opened 2021-06-22
Build test using Github Actions workflow (vehicle_signal_specification)

gunnarx:
Still having some issues with Travis builds so here is a stopgap using the built in runners for GitHub Actions / Workflows.

I did not remove .travis.yml in this commit since we probably want to test this first and we might sort out Travis.

This includes make docs and deploy docs just like the travis file, and gh-pages are deployed only if it was a push to master.

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -1,0 +1,51 @@
+name: Standard Build Check
+
+on: [push, pull_request]
+
+jobs:
+  buildtest:
+    name: Build Test
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


gunnarx:
Also here I pushed to a [branch](https://github.com/GENIVI/vehicle_signal_specification/tree/github-actions) so we can see the build check in action --> results below (it did not trigger otherwise)

(!) Remember to remove [the branch](https://github.com/GENIVI/vehicle_signal_specification/tree/github-actions) after merging.

gunnarx:
Needed rebase on master.

PR #314, opened 2021-07-14
Fixing imports

oliveirarleo:
In these two files the relative import is incorrect, since:

1. There is no Identifier/ folder inside Cabin, where resides
   Occupant.vspec
2. There is no Cabin/ folder inside Driver/, where resides
   Driver.vspec

### DIFF
--- a/spec/Cabin/Occupant.vspec
+++ b/spec/Cabin/Occupant.vspec
@@ -11,4 +11,4 @@
 Identifier:
   type: branch
   description: Identifier attributes based on OAuth 2.0.
-#include Identifier/Identifier.vspec Identifier
+#include ../Identifier/Identifier.vspec Identifier
--- a/spec/Driver/Driver.vspec
+++ b/spec/Driver/Driver.vspec
@@ -9,7 +9,7 @@
 #
 
 
-#include Cabin/Occupant.vspec
+#include ../Cabin/Occupant.vspec
 
 DistractionLevel:
   datatype: float


barbieri (approved):

erikbosch:
To me the changes seems to be OK, based on the discussion in https://github.com/GENIVI/vehicle_signal_specification/issues/195.

Note that the change does not imply any change to generated csv, as of now the tooling seems to search in many places, so it find the files both with old and new location.

erikbosch:
@gunnarx @danielwilms - does anyone of you need to do a "workflow approval" to trigger a build?

danielwilms:
> @gunnarx @danielwilms - does anyone of you need to do a "workflow approval" to trigger a build?

just for first-time contributors. triggered the build.

About the change I don't have a strong opinion, as it's working as is for now, isn't it?

danielwilms:
please add documentation @oliveirarleo

oliveirarleo:
Taking a look at the [documentation](https://genivi.github.io/vehicle_signal_specification/rule_set/includes) we already have described the behaviour of imports.

> The <filename> part specifies the path, relative to the file with the #include directive, to the vspec file to replace the directive with.

In that way, the behaviour of the order of the `include_paths` should be documented at [GENIVI/vss-tools](github.com/GENIVI/vss-tools), right?

gunnarx (approved):

danielwilms:
meeting 26/06/2021

PR #316, opened 2021-07-20
Changing seat signals from sensors to actuators

### LINKED ISSUE #315: Adding Seat Position Actuators
In https://github.com/GENIVI/vehicle_signal_specification/blob/master/spec/Cabin/SingleSeat.vspec we have numerous signals for controlling seats. Typically each feature consist of one sensor and two boolean actuator switches to change the value. Below is an example for position. 

This works quite well if you have e.g. an app with buttons to move a seat forward/backward (with speed defined by vehicle), but it might not work that well if you want to move the seat to a previously set position, then it would be easier to be able to request an absolute position. This could be easily achieved by changing Position from sensor to actuator. 

_Does anyone have any objection changing Position and similar seat signals from sensors to actuators?_

(I.e. keeping the Forward/Backward actuators as an alternative method to move the seat)

```
Position:
  datatype: uint16
  type: sensor
  min: 0
  max: 1000
  unit: mm
  description: Seat horizontal position. 0 = Frontmost. 1000 = Rearmost

Switch.Forward:
  datatype: boolean
  type: actuator
  description: Seat forward switch engaged

Switch.Backward:
  datatype: boolean
  type: actuator
  description: Seat forward switch engaged
```

erikbosch:
As of today seat settings can only be set by switches, making it
difficult to e.g. restore an old setting. By changing sensors to
actuators it will be easy for e.g. a VSS client to e.g. restore a seat
to a previously used setting.

Also added a signal for seat height, previously only switches existed
for that setting, not any sensor/actuator.

Fixes #315

### DIFF
--- a/spec/Cabin/SingleSeat.vspec
+++ b/spec/Cabin/SingleSeat.vspec
@@ -9,6 +9,13 @@
 #
 # Seat signals
 #
+# VSS offer two methods to control a seat.
+# As most seat signals are actuators it is possible to request a specific setting,
+# e.g. to request that seat decline shall be 23 degrees.
+# It is also possible to control seat position by switch buttons (boolean actuators)
+# If a switch is engaged (== has the value true) the seat is expected to move according to
+# the selected switch. The movement is normally supposed to continue until either the switch is released,
+# (== has the value false), or until the maximum/minimum value supported by the vehicle has been reached.
 
 HasPassenger:
   datatype: boolean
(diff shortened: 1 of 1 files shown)


danielwilms (approved):

dwhiteddsoft:
Only question I have is if we have a actuator (ie. headrest position) why do I also have a switch? Is it simply to represent when the switch is being pressed?

erikbosch:
@dwhiteddsoft - I do not know the background for the switch-signals, but depending on use-case I can see benefits in supporting both an actuator and up/down switches. Switches might e.g. be useful if you want to have an app acting as remote control for the seat. You press e.g. the "up botton" and the app sends "up_switch_pressed==true" to the vehicle. When you are happy you release the "up button" and the app sends "up_switch_pressed==false" to the vehicle and the vehicle will stop the headrest motor. Here the switches gives an easier implementation in the app compared to if only the actuator existed, and the app would need to read current height, request height to be increased a few millimetres, wait until that height is reported by the vehicle and if button still is pressed request some more millimetres.

gunnarx (approved):
I think this change is a significant quality improvement.

danielwilms:
meeting 26/07/2021

PR #319, opened 2021-08-09
Improving enum documentation

erikbosch:
Fixing two flaws in enum documentation

1. Enums can (and are) also used for attributes, not only sensors and actuators
2. Enums can be used for both strings and ints.
   Current example only showed int-example, while VSS so far just use
   strings.

Also moving down enum to after attributes


erikbosch:
Discussed at meeting 2021-08-10. Remove old example with int, as we do not use int enums in VSS. To be further discussed in subsequent meetings

SebastianSchildt (approved):

SebastianSchildt:
Meeting 08/28 merge this, and keep discussing the enum issue

PR #324, opened 2021-08-11
Refactoring OBD oxygen sensors

### LINKED ISSUE #304: OBD - Oxygen sensors and banks - refactoring needed?
**Summary**

In the last meeting we discussed the empty O2Alt branches in OBD. I am by no way an OBD expert but with my understanding our current representation is wrong. Below is a quite lengthy analysis. Is someone actively using the VSS OBD signals as of today? Any feedback from them on the Oxygen signals would be much appreciated.

Is this is an area where we want to do some refactoring now, or do we want to wait until we have a "real user" of the signals?

**Analysis**

Based on https://en.wikipedia.org/wiki/OBD-II_PIDs my understanding is that PID 0x13 and 0x1D is 1 byte bitmasks to represent how oxygen sensors are arranged in a vehicle. Most vehicles have at most 2 banks and use 0x13 to represent which banks/sensors that are present. Some supercars might have more, e.g. Audi S8 with a V10 engine, and then 0x1D is used to represent which banks/sensors that are present. Regardless, OBD supports at most 8 sensors, values can be received in 0x14-0x1B, 0x24-0x2B and 0x34-0x3B. 

References:
https://mechanicbase.com/troubleshooting/bank1-vs-bank2/
https://www.obd-2.de/sensoren.html
https://jhmotorsports.com/o2-or-oxygen-sensor-upstream-bank-4-oem-bosch-for-s6-late-08-and-s8-v10.html
https://github.com/oesmith/obdgpslogger/blob/master/src/obdinfo/obdservicecommands.h

Based on this I think it is wrong to have branches like `Vehicle.OBD.O2.Bank1.Sensor1` and `Vehicle.OBD.O2Alt.Bank1.Sensor1`. If we want to use branches for O2 sensors we should likely instead use something like `Vehicle.OBD.O2WR.Sensor1`, i.e. without specifying bank as that might differ depending on how 0x13 and 0x1D are used.

But then we might have another problem - in OBD the O2 PIDs are 2/4 bytes long, and contain two values

```
0x14-0x1B A: Voltage B: Short term fuel trim 
0x24-0x2B AB: Air-Fuel Equivalence Ratio (lambda,λ) CD: Voltage 
0x34-0x3B AB: Air-Fuel Equivalence Ratio (lambda,λ) CD: Current 
```
... but in our current VSS we just represent the voltage/current. If we want to represent both values then we need to split the values. That could then give us something like:

```
Vehicle.OBD.O2.SensorsPresent2Banks (uint8, 0x13)
Vehicle.OBD.O2.Sensor1.Voltage (float, Volts, 0x14 byte A)
Vehicle.OBD.O2.Sensor1.ShortTermFuelTrim (float, Percent, 0x14 byte B)
Vehicle.OBD.O2.Sensor2.Voltage (float, Volts, 0x15 byte A)
Vehicle.OBD.O2.Sensor2.ShortTermFuelTrim (float, Percent, 0x15 byte B)
...
Vehicle.OBD.O2.SensorsPresent4Banks (uint8, 0x1D)
```

... and then the `Vehicle.OBD.O2Alt` branch could be removed. If you want to know which bank sensor 3 belongs to you must check 0x13 and 0x1D. Note however that we also better should review the  `Vehicle.OBD.O2WR` branch. as the lambda value is not currently represented. That could easily be added.  The lambda is however represented in both 0x24 and 0x34. I do not know if vehicles typically report both current and voltage, or if they typically just report either voltage or current.

```
Vehicle.OBD.O2WR.Sensor1.Lambda (float, 0x24 AB, same as 0x34 AB?)
Vehicle.OBD.O2WR.Sensor1.Voltage (float, 0x24 CD)
Vehicle.OBD.O2WR.Sensor1.Current (float, 0x34 CS)
...
```

erikbosch:
Fixes #304

This is to my understanding a better representation compared to current one.
I am by no means an OBD-expert so input from anyone with good knowledge on OBD or experience from using the current VSS OBD specification would be great.

### DIFF
--- a/spec/OBD/OBD.vspec
+++ b/spec/OBD/OBD.vspec
@@ -161,141 +161,44 @@ AirStatus:
   type: sensor
   description: PID 12 - Secondary air status
 
+# OBD supports reporting of up to 8 oxygen sensors
+# PID 13 or PID 1D can be used as bitmasks to define logical location of the sensors
+# Example 1: PID 13 = 0x11 means that PID 14 and PID 18 shall be present
+# PID 14 is then first sensor on bank 1, PID 18 is the first sensor on bank2
+# Vehicles with more than 2 banks report sensor mapping in PID 1D instead of PID 13
+#
+OxygenSensorsIn2Banks:
+  datatype: uint8
+  type: sensor
+  description: PID 13 - Presence of oxygen sensors in 2 banks. [A0..A3] == Bank 1, Sensors 1-4. [A4..A7] == Bank 2, Sensors 1-4
+  
 O2:
+  instances:
+    - Sensor[1,8]
   type: branch
-  description: PID 13 - Presence of oxygen sensors for the banks
+  description: Oxygen sensors (PID 14 - PID 1B)
 
-O2.Bank1:
-  type: branch
-  description: Oxygen sensors on bank 1, at most 4 sensors per bank
-
-O2.Bank1.Sensor1:
-  type: branch
-  description: Oxygen sensor
-
-O2.Bank1.Sensor1.Voltage:
+O2.Voltage:
   datatype: float
   type: sensor
   unit: V
-  description: PID 14 - Sensor voltage
-
-O2.Bank1.Sensor2:
-  type: branch
-  description: Oxygen sensor
-
-O2.Bank1.Sensor2.Voltage:
+  description: PID 1x (byte A) - Sensor voltage
+  
+O2.ShortTermFuelTrim:
   datatype: float
   type: sensor
-  unit: V
-  description: PID 15 - Sensor voltage
+  unit: percent
+  description: PID 1x (byte B) - Short term fuel trim
 
-O2.Bank1.Sensor3:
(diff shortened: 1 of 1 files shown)


gunnarx:
@danielwilms to ask internally among people knowledgable in OBD

SebastianSchildt:
I am by no means an OBD expert so I  would like to vote for a "weak accept" here on the grounds of

 - it does make more sense to me, than the previous version
 - the old version was definitely broken
 - semantics aside, this should be a better/more consistent VSS model (no empty branches)

SebastianSchildt (approved):

gunnarx (commented):
Last I heard we were waiting for stakeholder companies to reach out internally to potential diagnostic experts.
I don't know the details well enough but I agree it seems to be an improvement compared to the old tree that was broken because it had empty branches, for example.

danielwilms:
Good for now. Will try to get internal feedback later.

PR #334, opened 2021-09-01
Changing seat instantiation and improving default value documentation

### LINKED ISSUE #329: Default value behavior for arrays
If datatype is specified as an array of some types, why default value is not an array, even if it's a single element?  Maybe this is an typo, but documentation isn't clear in this case too. 

example from generated json:

```json
"SeatPosCount": {
            "datatype": "uint8[]",
            "default": 0,
            "description": "Number of seats across each row from the front to the rear",
            "type": "attribute",
            "uuid": "8dd40ecd47ab51c79ed9c74ae4296d7e"
          }
```

erikbosch:
Content of this change:

- Improved documentation on how to set default values for array
- Changing default values for seats/doors. Now default corresponds to a classical passenger car with 2 rows of seats, 2 seats in front row and 3 in back row. That reduces the size of generated "standard VSS" significantly.

Fixes #329

Example on how generated JSON with new default looks like:
```
 "SeatPosCount": {
            "datatype": "uint8[]",
            "default": [
              2,
              3
            ],
            "description": "Number of seats across each row from the front to the rear",
            "type": "attribute",
            "uuid": "8dd40ecd47ab51c79ed9c74ae4296d7e"
          },
```

### DIFF
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -94,64 +94,83 @@ Lights:
 
 ##
 # Door signals and attributes
+# The default VSS (i.e. this file) assumes a vehicle with two rows of doors
+# with two doors in front row and two doors in second row.
+# For real deployments this needs to be overridden with actual values
+# Row1 is the frontmost row.
+# Row2-RowX are subsequent rows of doors.
+#
+# Note that rear door (for hatchback and station wagons) shall typically not be represented
+# as a door, but rather be represented as a trunk (Vehicle.Body.Trunk) and
+# rear shade (Vehicle.Cabin.RearShade)
 ##
 Door:
   type: branch
   instances:
-    - Row[1,4]
+    - Row[1,2]
     - ["Left","Right"]
   description: All doors, including windows and switches
 #include SingleDoor.vspec Door
 
+# Default value based on instance declaration above with 2 rows and 2 doors in each row.
 DoorCount:
   datatype: uint8
   type: attribute
-  default: 0
+  default: 4
   description: Number of doors in vehicle
 
 
 
 ##
 # Seat signals and attributes
+# The default VSS (i.e. this file) assumes a vehicle with two rows of seats
+# with two seats in front row and three seats in second row.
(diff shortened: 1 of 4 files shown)


spec/Cabin/Cabin.vspec
```diff
@@ -98,61 +98,71 @@ Lights:
 Door:
   type: branch
   instances:
-    - Row[1,4]
+    - Row[1,2]
```
erikbosch, line 101: Adapting number of rows to changed number of seats.

But then I started to think about hatchback and station wagons. At least in some countries they are sometimes referred to as 3/5-door vehicles. 

I think we would benefit from making it clear in a comment in VSS if the rear door for hatchbacks and station wagons shall be considered as a door (and represented here) or being considered as a trunk (`Vehicle.Body.Trunk`). If we want to consider them as a door we should preferably give some guidance on what literal to use for position. E.g. by either having a station wagon as "default vehicle" and include the position "center", or just giving a comment that rear doot shall preferably have the position "center"


```
Door:
  type: branch
  instances:
    - Row[1,3]
    - ["Left", "Center", "Right"]
  description: All doors in the vehicle.

# Default value based on instance declaration above with 2 rows and 2 doors in each row.
DoorCount:
  datatype: uint8
  type: attribute
  default: 5 # (i.e. 2 in first row, 2 in second row, and then a rear door as part of third row)
  description: Number of doors in vehicle

```

But that would give us a specification with quite a lot of holes, e.g. there will be a `Row1.Center` in generated JSON/CSV even if not needed. The alternative is of course that we add a comment that rear door (if any) is to be handled

danielwilms: At the end it could be a combination of the original instances in door and a second branch Door.Hatchback with instances `[left, center, right]`

erikbosch: I think it boils down to what we need for the rear door. Is the open/locked-info part of Vehicle.Body.Trunk sufficient, or do we need all window/shade-information that you typically have for a door as well.

danielwilms:
2021/09/7: decision to merge

PR #335, opened 2021-09-03
Issue#328-Gear split into CurrentGear and SelectedGear.

UlfBj:


spec/Powertrain/Transmission.vspec
```diff
@@ -70,12 +70,19 @@ TravelledDistance:
 #
 # Current gear
 #
-Gear:
+CurrentGear:
```
erikbosch, line 73: Here we touch the "deprecated" topic. If we know or can assume that no-one is using the old signal, then I think it is fine that we just remove/rename the existing signal. If not we should keep the old signal as well and mark it as deprecated, to give users time to change to the new signal name

UlfBj:
GearSelect might be a better name than SelectedGear as it is an actuator.

spec/Powertrain/Transmission.vspec
```diff
@@ -70,12 +70,19 @@ TravelledDistance:
...
+  datatype: int8
+  type: sensor
+  description: The current gear. 0=Neutral, 1/2/..=Forward, -1/..=Reverse
+
+
+#
+# Selected gear
+#
+SelectedGear:
   datatype: int8
```
danielwilms, line 83: here we could work with an enum, maybe?

SebastianSchildt:
Generally, I approve the split

spec/Powertrain/Transmission.vspec
danielwilms: meeting conclusion: We can leave it as is, but use deprecation.

PR #336, opened 2021-09-13
Fixes range for uint64

erikbosch:
Max should be 2^64 - 1, right?


SebastianSchildt (approved):
True. 2^64-1 or 18446744073709551615 if feeling verbose 👍

danielwilms (approved):

PR #344, opened 2021-10-13
Makefile: Add csv2pdf converter (optional group)

gunnarx:
Relates to PR https://github.com/GENIVI/vss-tools/pull/112

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -34,9 +34,6 @@ jobs:
       - name: Test optional targets. NOTE - always succeeds
         run: make -k travis_optional || true
 
-      - name: Test csv2pdf
-        run: vss-tools/contrib/csv2pdf.py vss_rel*.csv
-
       - name: Install hugo
         env:
           HUGO_VER : 0.78.1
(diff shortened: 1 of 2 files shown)


.github/workflows/buildcheck.yml
```diff
@@ -28,7 +29,10 @@ jobs:
         run: make travis_targets
 
       - name: Test optional targets. NOTE - always succeeds
-        run: make -k travis_optional || true
+        run: |
+          sudo apt-get install -y pandoc  # For csv2pdf
+          make -k travis_optional || true
+          ls -al vss-tools/contrib *.pdf
```
erikbosch, line 35: Noticed that the last check failed due to the "frozen" error that hopefully is fixed with the last commit in vss-tools.

But the reason for the failed check was that `ls` did not find any pdf file. If we want that part to always pass we should add `true`like below

`ls -al vss-tools/contrib *.pdf || true`

gunnarx:
I have decided to close this PR because the pandoc tools is very different between versions.  The version installed in the GitHub Actions (which is based on Ubuntu 20.04) seems to be v2.5 and this one does not have the CSV input filter?  Without the csv input filter, this script does not make much sense.  Probably better to return to this with a new approach.  I will keep the (messy) draft commits for future reference.

PR #355, opened 2021-10-27
README.md: Change GENIVI to COVESA

SebastianSchildt:
The link to the documentation in the Readme gave a 404 as the genivi subdomain for github pages has been deleted. This PR fixes that. Also replaced other occurences of Genivi with COVESA


gunnarx:
I think there might be even more places :)
I also have a PR coming.  But we an start with this, or I can build on top of it.

danielwilms (approved):

Issue #343, opened 2021-10-13
Use of VersionVSS: branch in tagged releases


slawr:
Hi,

Looking recently into the code I noticed that the values of the [VersionVSS](https://github.com/GENIVI/vehicle_signal_specification/blob/bcb53893dcff5990d4735168b8966b62d927d2ac/spec/Vehicle/Vehicle.vspec#L18) branch attributes were not set in tagged releases. With the exception of the end user focused 'VersionVSS.Label' attribute, I thought the other attributes were intended to reflect the VSS version - possibly even be fixed.

This is probably a learning moment for me, but wouldn't you expect the tagged releases to have those values set? In the Vehicle.vspec or the serialisation rendering of the data description. That's not the case in the VSS v2.x releases I looked at.

e.g. VSS v2.1 using 'default':
```
VersionVSS.Major:
  datatype: uint32
  type: attribute
  description: Supported Version of VSS - Major version
  default: 2

VersionVSS.Minor:
  datatype: uint32
  type: attribute
  description: Supported Version of VSS - Minor version
  default: 1
```

or 'min':
```
VersionVSS.Major:
  datatype: uint32
  type: attribute
  description: Supported Version of VSS - Major version
  min: 2

VersionVSS.Minor:
  datatype: uint32
  type: attribute
  description: Supported Version of VSS - Minor version
  min: 1
```

slawr:
@gunnarx, @danielwilms , @erikbosch (as a random group from W3C calls) bump :)

Background to this query was some work I am planning with VSS data in CCS/CVII and wanted to set this appropriately. From CSS call today my working assumption for now is that the standards body would be setting the major and minor number in the releases.

danielwilms:
Close issue, as PR is merged.

PR #358, opened 2021-11-08
Vehicle.vspec: add missing version attribute values

slawr:
Currently no values are set for the VSS Major, Minor and Patch
attributes in the Vehicle.vspec that communicate the VSS version.
Address this by setting a default value for each attribute starting
with the last released version 2.1.0.

Possible solution to issue #343

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -22,16 +22,19 @@ VersionVSS:
 VersionVSS.Major:
   datatype: uint32
   type: attribute
+  default: 2
   description: Supported Version of VSS - Major version
 
 VersionVSS.Minor:
   datatype: uint32
   type: attribute
+  default: 1
   description: Supported Version of VSS - Minor version
 
 VersionVSS.Patch:
   datatype: uint32
   type: attribute
+  default: 0
   description: Supported Version of VSS - Patch version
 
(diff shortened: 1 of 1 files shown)


erikbosch:
We have now and then discussed if it makes sense to have `default` in standard VSS catalog, as in most cases like for `MaxPower` and other attributes the default value anyway needs to be overwritten (by layers?) But for the 3 attributes in this PR I think a default value makes sense.

Biut it is important that we add it to our "release checklist" that the values must be checked and updated as needed,

erikbosch (approved):

slawr:
Completely agree on the release process. 

I also considered some subtleties about the handling of the values as they would be important in interoperability. Whether it might be a read only value for instance and who would set it. The standards body on release or whether some flexibility was required. So I started with the issue intending to discuss before creating a PR, but based on the discussion in CCS today decided to post something concrete to discuss.

erikbosch:
@danielwilms  - do we currently have any other document for the release process than https://github.com/COVESA/vehicle_signal_specification/blob/master/RELEASE_PROCESS.md

I am thinking of the minor steps required, like updating the main README file, updating the VERSION file, and if this PR is merged also updating default values

danielwilms (approved):

SebastianSchildt (approved):
Makes sense

PR #363, opened 2021-11-15
Improving battery signals

### LINKED ISSUE #354: Issues in Battery.vspec
Some ambiguities/errors have been identified for Battery.vspec. Originally one error was detected as part of https://github.com/COVESA/vehicle_signal_specification/pull/351 but during review additional issues was identified. By that reason closing the original PR and keep the discussion in this issues instead.

Initially identified issues:

- Charging.TimeToComplete - description says that negative values can be use but datatype is unsigned. No definition on what negative number to use. Is a negative number needed? One could think of using 0 for both "charging completed" and "no charging active", indicated by e.g. Charging.Status
- Charging.Status - Is boolean - unclear what true/false means. Shall true mean "charger connected, charging ongoing or completed" 
- Charging.Timer.Time - using unsigned int - Shall 0 be used to indicate that no timer/time has been set.

I could prepare a  PR but would like input on what changes you see as needed/wanted to avoid errors and ambiguities

erikbosch:
This PR tries to remove some ambiguities I see. Fixes #354 

I am not saying that the proposed changes are "right". I am happy to receive counter-proposals if someone has a better proposal, or if we in some cases shall keep the ambiguity as an OEM variation point.

### DIFF
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -12,23 +12,25 @@
 # For some vehicles the traction battery is referred to as the high voltage battery.
 #
 
-# It might be possible to disconnect the traction battery used by an electric powertrain.
-# This is achieved by connectors, typically one for plus and one for minus.
-PowerConnected:
+IsPowerConnected:
   datatype: boolean
   type: sensor
   description: Indicating if the power (positive terminator) of the traction battery is connected to the powertrain.
+  comment: It might be possible to disconnect the traction battery used by an electric powertrain.
+           This is achieved by connectors, typically one for plus and one for minus.
 
-GroundConnected:
+IsGroundConnected:
   datatype: boolean
   type: sensor
   description: Indicating if the ground (negative terminator) of the traction battery is connected to the powertrain.
+  comment: It might be possible to disconnect the traction battery used by an electric powertrain.
+           This is achieved by connectors, typically one for plus and one for minus.
 
 Temperature:
   datatype: float
   type: sensor
   unit: celsius
-  description: Temperature of the battery pack
+  description: Temperature of the battery pack.
 
 StateOfCharge:
   type: branch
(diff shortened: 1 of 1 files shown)


spec/Powertrain/Battery.vspec
```diff
@@ -127,7 +127,7 @@ Charging.ChargePortFlap:
 Charging.ChargePlugStatus:
   datatype: boolean
   type: sensor
-  description: Signal indicating if charge plug is connected or not.
+  description: Signal indicating if a charging cable is connected to the charge plug or not.
```
erikbosch, line 130: I am not 100% sure on the intention of this signal. If the intention is to focus on the cable or on the connection to a charger? I.e. if you connect a cable to a vehicle, but do not connect the other end to a charger, shall this signal then anyway shall true? Or shall the focus rather be on if the vehicle is connected to a functional/powered charger?

spec/Powertrain/Battery.vspec
```diff
@@ -185,12 +187,12 @@ Charging.Timer.Time:
...
   type: actuator
   unit: s
-  description: Time value for next charging-related action (Unix timestamp, seconds).
+  description: Time value for next charging-related action (Unix timestamp, seconds). 0 if no timer has been set.
 
 Charging.Timer.Mode:
   datatype: string
   type: actuator
-  enum: [ 'starttime', 'endtime', 'depaturetime' ]
+  enum: [ 'inactive', 'starttime', 'endtime', 'departuretime' ]
```
erikbosch, line 195: Here I assume that a specific element to indicate that timer-based charging is deactivated could make sense if you want to cancel planned charging, rather than setting Charging.Timer.Time to 0.

I corrected spelling for departuretime as well, but do not know if it makes sense to keep it as long as we do not have any "departuretime" signal in VSS

spec/Powertrain/Battery.vspec
```diff
@@ -145,7 +145,7 @@ Charging.Mode:
 Charging.Status:
```
danielwilms, line 145: * @danielwilms will check internally (@adobekan, or you?) 
* other proposal: `isCharging` instead of `Status`

spec/Powertrain/Battery.vspec
erikbosch: Meeting: Change name to "IsChargingCableConnected"

spec/Powertrain/Battery.vspec
```diff
@@ -185,12 +187,12 @@ Charging.Timer.Time:
   datatype: uint32
```
erikbosch, line 187: Meeting: use string, format according to ISO standard 8601

erikbosch:
Reworked based on comments from last meeting, changes performed:

- Renamed boolean signals to start with "Is"
- Align descriptions so they alway end with a period (`.`)
- Changed timer to use string value. Also proposing to remove the `departuretime` alternative as we do not have any signal for that in VSS

adobekan:
@danielwilms If needed, I can check next week. Somehow i missed the notification.

erikbosch:
Updated `Charging.Timer.Time` to clarify that time shall be given in UTC.

An ISO 8601 timestamp can exclude time zone information, i.e. referring to "local time zone" which would be ambiguous.  I also do not really see any value in specifying time with offset from UTC. This is aligned with VISS that also require UTC (Z) time.

danielwilms (approved):

danielwilms:
meeting 2021-12-14 approved

Issue #364, opened 2021-11-15
Adding support of "comment"?


erikbosch:
In VSS we have today `description`, and we can give unstructured comments by using `#`. I am considering if it would be beneficial to extend our language definition so that also structured comments could be given. Then we can have the recommendation that description shall be rather short, more like a title, additional information shall be put in comments.

If we think this would be a good idea the first step would be to change documentation and generic tooling so that `comment` is accepted and consumed. Then I would consider it optional for tools (csv, json, ocf, protobuf, ...) if they want to include comments in their output or not.

Example of a current signal:

```
DistractionLevel:
  datatype: float
  type: sensor
  unit: percent
  description: Distraction level of the driver will be the level how much the driver is distracted, by multiple factors. E.g. Driving situation, acustical or optical signales inside the cockpit, phone calls
```

Possible usage if comment is used:

```
DistractionLevel:
  datatype: float
  type: sensor
  unit: percent
  description: Distraction level of the driver.
  comment: Can be calculated by multiple factors E.g. Driving situation, acustical or optical signales inside the cockpit, phone calls
```

danielwilms:
Thanks @erikbosch for the suggestion. I absolutely think we should find a solution for it, as otherwise the information in the unstructured comments is 'lost'. I like `comment` as keyword. It's used as well in rdfs.

erikbosch:
All PRs merged, closing it

PR #367, opened 2021-11-17
Proposed documentation of "description" and "comment"

erikbosch:
In last VSS meeting we discussed #364. There seems to be an agreement that `comment`could be useful, but it must be clear that all information required to understand a signal or to send a signal with the correct value must be present in description together with other. I tried to document this. I also tried to add some style-guidelines, think that would be good to have.

If we take the `DistractionLevel` signal as an example, then maybe something like below could fulfill our requirements on how `description` and `comments` shall be used. This is a typical case where description is important as there is to my knowledge no standard on how to measure distraction. I assume 100 means that the driver is unconscious, dead or absent and does not react to anything. A sleeping driver is possibly not 100% distracted as the driver might wake up on strong noise.  But shall it be up to the OEM to decide if a driver that has been sleeping for 5 minutes shall be considered to be 95 or 99 percent distracted?

Comments appreaciated. And this PR shall not be merged until we have added basic support for comments in tooling.

```
DistractionLevel:
  datatype: float
  type: sensor
  unit: percent
  min: 0
  max: 100
  description: Distraction level of the driver. 0 = Fully Concentrated on Driving, 100 = Unconscious.  Exact meaning of intermediate values not standardized.
  comment: Can be deduced by multiple factors. E.g. Driving situation, acustical or optical signals inside the cockpit, phone calls.
```

### DIFF
--- a/docs-gen/content/rule_set/data_entry/sensor_actuator.md
+++ b/docs-gen/content/rule_set/data_entry/sensor_actuator.md
@@ -9,8 +9,9 @@ entry example is given below:
 ```YAML
 Speed:
   type: sensor
-  description: The vehicle speed, as measured by the drivetrain.
-  datatype: uint16
+  description: The vehicle speed.
+  comment: For engine speed see Vehicle.Powertrain.CombustionEngine.Engine.Speed.
+  datatype: float
   unit: km/h
   min: 0
   max: 300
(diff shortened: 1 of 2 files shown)


docs-gen/content/rule_set/data_entry/sensor_actuator.md
```diff
@@ -9,8 +9,9 @@ entry example is given below:
 ```YAML
 Speed:
   type: sensor
-  description: The vehicle speed, as measured by the drivetrain.
-  datatype: uint16
+  description: The vehicle speed.
+  comment: For engine speed see Vehicle.Powertrain.CombustionEngine.Engine.Speed.
```
erikbosch, line 13: Here I added comment after description. It has no semantic meaning, but I believe our examples should use our preferred order, and I do not know if this is the preferred order.

danielwilms:
I like the documentation and I have nothing to add. 

One thing I would add is since which version it's used. Here, I would suggest 3.0, like [here](https://covesa.github.io/vehicle_signal_specification/rule_set/basics/#deprecation-since-version-21)

danielwilms (approved):

docs-gen/content/rule_set/data_entry/sensor_actuator.md
```diff
@@ -24,14 +25,22 @@ all parental branches included in the name must be defined as well.
...
+Describes the meaning and content of the signal.
+The `description`shall together with other mandatory members like `datatype` and `unit` provide sufficient information
+to understand what the signal contains and how signal values shall be constructed or interpreted.
+Recommended to start with a capital letter and end with a dot (`.`).
+
+**```comment ```**  *[optional]* `since version 3.0`
+A comment can be used to provide additional informal information on a signal.
+This could include background information on the rationale for the signal design,
+references to related signals, standards and similar.
+Recommended to start with a capital letter and end with a dot (`.`).
```
erikbosch, line 42: @gunnarx - you wanted me to add something like "The information shall not contradict the information given in the description", right?

SebastianSchildt:
Lgtm @gunnarx

danielwilms:
@gunnarx, please create another PR if description is not sufficient.

meeting 2021-12-14 approved

PR #373, opened 2021-11-25
        Cleanup of CombustionEngine file

erikbosch:
Following changes included:

- Order rearranged so that attributes comes first
- Changing EngineOilLevel from attribute to sensor
- Improved/corrected descriptions
- Boolean signal Running renamed to IsRunning
- Limits that seem to be arbitrary removed
- Signals with previously had 0 as lower limit has been changed to unsigned

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -17,6 +17,7 @@
 #
 # Engine attributes
 #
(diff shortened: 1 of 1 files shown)


SebastianSchildt (approved):

danielwilms (approved):
meeting 2021-12-14 approved

PR #374, opened 2021-11-25
Refactoring engine time signals

### LINKED ISSUE #356: Ambiguous *Time signals in Vehicle.vspec
The `Vehicle.vspec` file contains four signals that are not well defined. It should better be discussed if they are useful. if they are found to be not useful they could be removed. If they are useful they should preferably get better descriptions, possibly related to the change in #348

- IgnitionOnTime
- IgnitionOffTime
- DriveTime
- IdleTime

Problems/Ambiguities identified so far:

- The description talks about "accumulated" but it is not clear when the value shall be reset. Shall they never be reset, i.e. represent total time during the lifetime of the vehicle, or should they be reset at some time, e.g. when the engine is replaced, by request of the driver or by some other event?
- What is the definition of IgnitionOn?` LowVoltageSystemState in ["ON", "START"]`?
- What is the definition of IgnitionOff?` LowVoltageSystemState in ["LOCK", "OFF","ACC"]`?
- What is the definition of DriveTime?` (LowVoltageSystemState in ["ON", "START"]) AND (Transmission.CurrentGear != 0)`?
- What is the definition of IdleTime?` (LowVoltageSystemState in ["ON", "START"]) AND (Transmission.CurrentGear == 0)`?

erikbosch:
The old *Time signals were not well defined. The need to have signals form accumulated engine hours
have been verified, and idle hours also seem to be relatively normal,
but the others does not seem to be needed.

This commit propose two new signals to replace the old *Time signals

Fixes #356

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -136,6 +136,22 @@ Speed:
   max: 20000
   description: Engine speed measured as rotations per minute.
 
+
+EngineHours:
+  datatype: float
+  type: sensor
+  description: Accumulated time during engine lifetime with 'engine speed (rpm) > 0'.
+  unit: h
+
+IdleHours:
+  datatype: float
+  type: sensor
+  description: Accumulated idling time during engine lifetime.
+               Definition of idling is not standardized.
+  comment: Vehicles may calculate accumulated idle time for an engine.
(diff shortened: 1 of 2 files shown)


gunnarx:
LGTM!

danielwilms (approved):
meeting 2021-12-14 approved

PR #379, opened 2021-12-06
Adding Time signal to VSS

erikbosch:
As of today we do not have a time signal in VSS. This PR proposes to add one. 

A possible use case could be internal applications that want to keep some form of log. Like all VSS-signals the delay, accuracy and update-frequency varies, so will likely be less useful for time-critical operations.

For now proposing it to be a sensor. Could alternatively be an an actuator if we want to allow Apps to set the time as well

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -385,6 +385,15 @@ Trailer.Connected:
   description: Signal indicating if trailer is connected or not.
 
 #
+# Time
+#
+
+Time:
+  datatype: string
+  type: sensor
+  description: Current date and time in the vehicle, formatted according to ISO 8601.
+
+#
 # Location
 #
 


SebastianSchildt:
Personally I doubt the usefulness of a generic "time" signal in VSS. Whatever language or framework is used for writing apps and services with VSS I can not imagine somebody will asks VSS for time.

 *  Use case "in-vehicle" App that is processing/logging time: Will use whatever API call the environment/language provides
 * Use case working with time-stamped data: A matter of implementation/protocol, i.e, VISS has timestamps in its protocol, and it is expected the VISS server puts reasonable ones
 * "Backend"/Cloud Apps: Would store - if so desired - whatever times come from the used protocol e.g. VISS; or timestamp data themselves as it comes in. Will probably never explicitly ask a vehicle about its opinion on time.

erikbosch:
Put on hold for now. May be reopened later.

PR #380, opened 2021-12-07
Adding signals for Diesel Exhaust Fluid

erikbosch:
This is related to number 17 adn 17 in #180 

In reail the fluid is often called AdBlue but that as I understand it a Brand/Trademark, and maybe something we should not mention in VSS

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -231,6 +231,43 @@ Torque:
            This negative torque shall be ignored, instead 0 shall be reported.
 
 #
+# Diesel Exhaust Fluid
+#
+
+DieselExhaustFluid:
+  type: branch
+  description: Signals related to Diesel Exhaust Fluid (DEF).
+               DEF is called AUS32 in ISO 22241.
+  comment: In retail and marketing other names are typically used for the fluid.
+
+DieselExhaustFluid.Capacity:
+  datatype: float
+  type: attribute
+  unit: l
+  description: Capacity in liters of the Diesel Exhaust Fluid Tank.
+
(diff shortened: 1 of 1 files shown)


spec/Powertrain/CombustionEngine.vspec
```diff
@@ -230,6 +230,42 @@ Torque:
...
+DieselExhaustFluid.Range:
+  datatype: uint32
+  type: sensor
+  unit: m
+  description: Remaining range in meters of the Diesel Exhaust Fluid present in the vehicle.
+
+DieselExhaustFluid.IsLevelLow:
+  datatype: boolean
+  type: sensor
+  description: Indicates if the Diesel Exhaust Fluid level is low.
```
SebastianSchildt, line 266: Nitpicking mode: Any specific reason why this description is multi-line, while the others are not?

erikbosch: Not really, Three sentences so would be quite long if put on the same line. But we have not really defined any guidelines on recommended line length, or? Might possibly be an idea to add?

SebastianSchildt: Let's not overdo explicitly spelling out best practices :) I see your argument, but the description would still fit into two reasonable sized lines... 🙈 
Anyway, approving it now, because the signals are what we are missing 👍

SebastianSchildt (approved):

danielwilms (approved):
meeting 2021-12-14 approved

danielwilms:
@erikbosch could you please fix the merge conflict

erikbosch:
Merge conflict fixed @danielwilms

Issue #323, opened 2021-08-11
Use of string constants in VSS


erikbosch:
When looking at enum documentation I realized that we are quite inconsistent concerning strings in VSS, if we use single quotes, no quotes or double quotes. This might possibly make life difficult for tools, and I think it would be good to align and document it. As part of this it would be good to document what special characters that are allowed for enums and/or string constants. If we allow special characters like line breaks, blank space, commas or quotes, we also need a defined escape-mechanism, possibly like `"hi\"there!"` or similar.

A few examples from VSS:

```
enum: [ "Inactive", "Close", "Open", "OneShotClose", "OneShotOpen" ]
enum: [ 'type 1', 'type 2', 'ccs', 'chademo' ]
enum: [ good, flat spot, low, uneven, separated, other damage ]

default: "unknown"
default: ccs
```

Some ideas:

- For string signals defined with enum allow only "uppercase and lowercase letters ( A-Z , a-z ), digits ( 0-9 ), and the underscore character ( _ )". That would require some changes to existing VSS signals, but would simplify for tools that want to represent them as true enums in target languages which often have similar restrictions. 
- Support all characters for string signals (that not use enum), i.e. "smörgåstårta is great!!!" should be allowed. Possibly with exception of line breaks and/or tabs, if we think that gives problems when parsing vspec files
- Always require quotes for string values in vspec files. (Single or double does not really matter for me). 
- Use backslash to escape quote

oliveirarleo:
I think this is a valid discussion and ends up showing that some information is missing comparing to enums in almost every programming language. 

Usually, enums come as ways to describe categories, or limited amounts of options to values. Here follows an example in C#:
```C#
enum ErrorCode : ushort
{
    None = 0,
    Unknown = 1,
    ConnectionLost = 100,
    OutlierReading = 200
}
```

We have described here an association between _symbolic values_ and _values_, like in the line `ConnectionLost = 100,`. As discussed in our meeting, I believe now `enum: ` follows with the _values_, not the _symbolic values_. We were using the strings received as the symbolic values, internally converting `flat spot` to `FLAT_SPOT` for example, which is not ideal.

Maybe a way to solve this would be having a structure similar to the following one:
```YAML
enum: [ "None", "Unknown", "ConnectionLost", "OutlierReading" ]
enumvalues: [ 0, 1, 100, 200]
```
[...]

PR #381, opened 2021-12-07
[SPEC] Changed `enum` to `allowed`

danielwilms:
Currently `enum` is used for restrictions on allowed values
for a specific data entry. This causes confusions about
usage and specification  as #323 shows.

This commit changes `enum` to `allowedValues` for spec files
and documentation.

### DIFF
--- a/docs-gen/content/rule_set/data_entry/enum.md
+++ b/docs-gen/content/rule_set/data_entry/enum.md
@@ -1,14 +1,13 @@
 ---
-title: "Enumerated Data Entries"
+title: "Defined Value restrictions"
 date: 2019-08-04T12:37:12+02:00
 weight: 5
 ---
 
-
-A data entry can optionally be enumerated, allowing it to be assigned a
-value from a specified set of values. An examples of a signal using enum is
-given below:
-
+Optionally it is possible to define an array of `allowed`, which will restrict the usage of the data entry in the implementation of the specification.
+It is expected, that any value not mentioned in the array is considered an error and the implementation of the specification shall react accordingly.
+The datatype of the array elements is the `datatype` defined for the data entry itself.
+For `attributes` it is possible to set optionally a default value.
 
 ```YAML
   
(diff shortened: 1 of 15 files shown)


erikbosch:
If we are to start changing enums for VSS 3.0 I would at the same time consider aligning values. Today we use 3 forms, `xxx`, `'xxx'` and `"xxx"`. At least it would be good to have a recommendation on how they should look like in VSS. Possibly something like:

- Start with [A-Z], then [A-Za-z0-9_]*
- Write without single/double quotes

I would myself be quite glad if we could get rid of the blanks in the existing VSS enums.

danielwilms:
> If we are to start changing enums for VSS 3.0 I would at the same time consider aligning values. Today we use 3 forms, `xxx`, `'xxx'` and `"xxx"`. At least it would be good to have a recommendation on how they should look like in VSS. Possibly something like:
> 
> * Start with [A-Z], then [A-Za-z0-9_]*
> * Write without single/double quotes
> 
> I would myself be quite glad if we could get rid of the blanks in the existing VSS enums.

Completely agree. Just went for the bare minimum here.

SebastianSchildt:
I agree with this change.

I also agree we should use the chance to clean up as @erikbosch  mentioned. It will never be "as easy" as on the verge of a new major release, but suggest that is done in a separate PR.

Expecting the camel/snake/lowercase discussion: I am fine with anything 😄

danielwilms:
possible alternatives to `allowedValues`:
* `values`
* `allowed`
please comment and then let's vote in the next meeting.

SebastianSchildt:
I would be fine with any choice, but if asked I would prefer `allowed`

danielwilms:
Decision 22/01/04: go with `allowed`

@danielwilms: will modify PR and rebase

danielwilms:
Meeting 11/01/2022: approved for merging

danielwilms:
Rebased and ready to merge, according to meeting decision and build successful.

PR #385, opened 2021-12-15
Clarifying use of GNSS and Timestamp

erikbosch:
This PR tries to improve two areas of VSS

**GNSS**

Previously we did not explicitly state that coordinates shall be based on WGS84, and the description on altitude was a bit ambiguous as it mentioned elevation. Now it clearly states that it is altitude over WGS84 reference ellipsoid.

(As I understand it altitude in WGS84 context refer to height over WGS84 reference ellipsoid, but elevation typically refers to height over WGS84 geoid (which varies depending on WGS84 revision used, e.g. EGM96, EGM2008)

**Timestamp**

We have recently started to specify that string formatted according to ISO 8601 shall be used for some signals. But ISO 8601 gives you a lot of alternatives and you do not even need to specify time zone, and then the string shall be interpreted as local time. We see no real reason to use anything else than UTC in signals, and we think it would be good to give additional details/recommendations in documentation.

The documentation proposal is reasonable aligned with what is specified by VISS and W3C:

https://www.w3.org/TR/NOTE-datetime
https://raw.githack.com/w3c/automotive/gh-pages/spec/VISSv2_Core.html#timestamp

### DIFF
--- a/docs-gen/content/rule_set/data_entry/data_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_types.md
@@ -41,6 +41,17 @@ arraysize: 5
 An example for the usage of `arrays` is `Vehicle.OBD.DTCList` which contains a list
 of Diagnostic Trouble Codes (DTCs) present in a vehicle.
 
+## Timestamps
+
+Timestamps are in VSS typically represented as strings, formatted according to ISO 8601.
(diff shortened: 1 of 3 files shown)


danielwilms (approved):
looks good, thanks

danielwilms:
Decision 22/01/04: approved

PR #386, opened 2021-12-15
Improving transmission signals

erikbosch:
- Adding missing signals related to e.g. diff lock
  and torque distribution
- Changing GearCount to int8 to be able to handle -1 = CVT as described in description
- Removing arbitrary limit on Temperature
- Make sure all descriptions end with period.


danielwilms:
agreed to merge 2022/01/18

PR #389, opened 2021-12-20
Adapting signal names to naming rules

### LINKED ISSUE #239: Field Naming convention
Hi,
I noticed that few fields in the spec violates the naming convention from other.  For example.

Vehicle.IgnitionOn --> Field Starts with Upper case.  - This is the majority
Vehicle.accelerationTime --> Field starts with lower case. ( very few fields are like this)

Wondering if this is intended or an issue ?. 

Thanks,
Jins

### LINKED ISSUE #293: Naming of schema.org signals
VSS has some signals from https://schema.org/Car. The naming (e.g. capitalization of letters) is not consistent with other VSS signals, and not always consistent with the capitalization used in schema.org. The expanded names also gets very long, like `Vehicle.VehicleIdentification.ACRISSCode` and `Vehicle.VehicleIdentification.vehicleinteriorColor` . Do we want to do any change? Are the signals actively used by someone today?

```
VSS:
VehicleIdentification.ACRISSCode
VehicleIdentification.vehicleinteriorColor

schema.org:
acrissCode
vehicleInteriorColor
```

### LINKED ISSUE #384: Check/change data catalog according to naming conventions
* check data catalog according to #371
* adapt tooling as well

erikbosch:
Making sure first character is upper case and that boolean signals
starts with "Is" as described by naming conventions in https://covesa.github.io/vehicle_signal_specification/rule_set/basics/

Also some other minor changes, like adding period at end of description when missing and removing branch if branch only has one member in a few situations.

Solves the warnings reported by https://github.com/COVESA/vss-tools/pull/128

Fixes #384 
Fixes #293 
Fixes #239

### DIFF
--- a/spec/OBD/OBD.vspec
+++ b/spec/OBD/OBD.vspec
@@ -41,7 +41,7 @@ Status:
   type: branch
   description: PID 01 - OBD status
 
-Status.MIL:
+Status.IsMILOn:
   datatype: boolean
   type: sensor
   description: Malfunction Indicator Light (MIL) False = Off, True = On
@@ -200,7 +200,7 @@ OxygenSensorsIn4Banks:
   type: sensor
   description: PID 1D - Presence of oxygen sensors in 4 banks. Similar to PID 13, but [A0..A7] == [B1S1, B1S2, B2S1, B2S2, B3S1, B3S2, B4S1, B4S2]
 
-AuxInputStatus:
+IsPTOActive:
   datatype: boolean
   type: sensor
   description: PID 1E - Auxiliary input status (power take off)
@@ -349,7 +349,7 @@ DriveCycleStatus:
   type: branch
   description: PID 41 - OBD status for the current drive cycle
 
-DriveCycleStatus.MIL:
+DriveCycleStatus.IsMILOn:
   datatype: boolean
   type: sensor
   description: Malfunction Indicator Light (MIL) - False = Off, True = On
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -17,31 +17,31 @@
 
 VersionVSS:
   type: branch
-  description: Supported Version of VSS
+  description: Supported Version of VSS.
 
 VersionVSS.Major:
   datatype: uint32
   type: attribute
   default: 3
-  description: Supported Version of VSS - Major version
+  description: Supported Version of VSS - Major version.
 
 VersionVSS.Minor:
   datatype: uint32
   type: attribute
   default: 0  
-  description: Supported Version of VSS - Minor version
+  description: Supported Version of VSS - Minor version.
 
 VersionVSS.Patch:
   datatype: uint32
   type: attribute
   default: 0
-  description: Supported Version of VSS - Patch version
+  description: Supported Version of VSS - Patch version.
 
 VersionVSS.Label:
   datatype: string
   type: attribute
   default: "develop"
-  description: Label to further describe the version
+  description: Label to further describe the version.
 
 #
 # Vehicle identification attributes.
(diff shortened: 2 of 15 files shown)


spec/OBD/OBD.vspec
```diff
@@ -200,7 +200,7 @@ OxygenSensorsIn4Banks:
   type: sensor
   description: PID 1D - Presence of oxygen sensors in 4 banks. Similar to PID 13, but [A0..A7] == [B1S1, B1S2, B2S1, B2S2, B3S1, B3S2, B4S1, B4S2]
 
-AuxInputStatus:
+IsPTOActive:
```
SebastianSchildt, line 203: PTO?

spec/Vehicle/Vehicle.vspec
```diff
@@ -17,131 +17,134 @@
...
 
-VehicleIdentification.dateVehicleFirstRegistered:
+VehicleIdentification.DateVehicleFirstRegistered:
   datatype: string
   type: attribute
-  description: The date of the first registration of the vehicle with the respective public authorities.
+  description: The date in ISO 8601 format of the first registration of the vehicle with the respective public authorities.
 
-VehicleIdentification.meetsEmissionStandard:
+VehicleIdentification.MeetsEmissionStandard:
```
SebastianSchildt, line 102: This may need rework in another PR. Waht is meant with this? THe string should be EuroV/Euro5 oder something like that? Maybe needs to be clarified more. But won't fix in this PR

spec/Vehicle/Vehicle.vspec
```diff
@@ -261,14 +264,14 @@ RoofLoad:
...
   type: attribute
   unit: l
   description: The available volume for cargo or luggage. For automobiles, this is usually the trunk volume.
   min: 0
 
-emissionsCO2:
+EmissionsCO2:
   datatype: int16
   type: attribute
   description: The CO2 emissions.
```
SebastianSchildt, line 277: For later: Needs better description

spec/OBD/OBD.vspec
erikbosch: Looking at https://en.wikipedia.org/wiki/OBD-II_PIDs the description of  `#1E` is Auxiliary input status, but  looking at the data only 1 bit is used and it is used for Power Take Off (PTO).  (see https://en.wikipedia.org/wiki/Power_take-off)  

```
A0 == Power Take Off (PTO) status (1 == active)
[A1..A7] not used
```

To make a good boolean name I think the name shall be based on PTO in some way, either like this or expanded like IsPowerTakeOffActtive. The current description gives some hints that PTO ="power take off" but that could of course be improved.

`PID 1E - Auxiliary input status (power take off)`

spec/Vehicle/Vehicle.vspec
erikbosch: Like all signals here it comes from schema.org (https://schema.org/docs/auto.home.html) and they do not specify it any further (https://schema.org/meetsEmissionStandard), seems to be a free text field that could contain text, URLs or anything.

But in general it seems like not much is happening in this area, I do not know if there are anyone actively maintaining the auto-related signals in schema.org. There is a W3C group but it seems to be quite dead. If no one use or maintain it we can consider removing the signals.

http://www.automotive-ontology.org/
https://www.w3.org/community/gao/

spec/Vehicle/Vehicle.vspec
erikbosch: Agree, I assume it should be linked to some standard. 

But is there possibly multiple standards? (e.g. EU vs US)  And do we need to handle different versions of them? E.g. an older car might report emissions according to an older standard, possibly not directly comparable to the value from a newer standard. Do we need/want multiple signals, one for each standard, so that a vehicle possibly can report values according to multiple standards?

SebastianSchildt: I just _assume_ this might be CO2/km, which will be given by car manufactuers according to some standard, and it also is used for taxation purposes in Germany.
That would be a useful attribute, but then we should describe it accordingly (and uint16 might be more appropriate).

We just need to make sure this isn't anything else :)

spec/Vehicle/Vehicle.vspec
SebastianSchildt: Having an attribute describing emission standard might still make sense. I just fear it would be quite some work to collect all possible allowed values. P19 https://www.continental-automotive.com/getattachment/8f2dedad-b510-4672-a005-3156f77d1f85/EMISSIONBOOKLET_2019.pdf/  gives some hint towards just the more recent ones...

danielwilms:
Meeting 11/01/2022: approved for merging

erikbosch:
@danielwilms @gunnarx - now rebased

danielwilms:
Thanks @erikbosch. Will merge then, as agreed in the meeting.

PR #397, opened 2022-01-12
[BUILD] Changed `buildcheck.yaml` to `pyenv` and `Pipfile`

danielwilms:
We changed in vss-tools from `requirements.txt` to
`Pipfile` and in order to update the submodule
we have to do the change here in VSS as well.

Updated `buildcheck.yaml`:
* changed to pyenv instead of python action
* Use `Pipfile` instead of `requirements.txt`

Related to: https://github.com/COVESA/vss-tools/pull/131

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -12,23 +12,40 @@ jobs:
         with:
           submodules: true
 
-      - name: Setup Python
-        uses: actions/setup-python@v1
-        with:
-          python-version: 3.8
+      - name: Install pyenv
+        run: |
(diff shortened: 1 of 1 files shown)


erikbosch:
The run give some complains on Pyton versions. I assume at least one explanation is that we still use the old vss-tools submodule, or?

https://github.com/COVESA/vehicle_signal_specification/runs/4788041057?check_suite_focus=true 
Warning: Your Pipfile requires python_version 3.8.5, but you are using 3.8.10 (/home/runner/work/v/v/v/.venv/bin/python).

danielwilms:
agreed to merge 2022/01/18

PR #401, opened 2022-01-17
Restructuring unit documentation

erikbosch:
This PR is related to https://github.com/COVESA/vss-tools/pull/134

It adapts the VSS documentation so that:

- The unit definitions in config.yaml and documentation matches, use the same domain name and are defined in the same order


danielwilms:
agreed to merge 2022/01/18

PR #402, opened 2022-01-21
Removing arbitrary min/max limits

erikbosch:
Min/Max values in VSS can serve two purposes:
- Define a logical range, like -180 to 180 for longitude
  or 0 to 100 for percentage
- Define constraints for a particular vehicle,
  e.g. supported range for seat position

This PR removes constraints that are vehicle specific.
Even if they might be useful they shall not be specified in the
standard VSS file, but rather be specified in a vehicle specific
VSS-file , possibly using VSS layering mechanism.

### DIFF
--- a/spec/Cabin/SingleSeat.vspec
+++ b/spec/Cabin/SingleSeat.vspec
@@ -64,17 +64,15 @@ Position:
   datatype: uint16
   type: actuator
   min: 0
-  max: 1000
   unit: mm
-  description: Seat horizontal position. 0 = Frontmost. 1000 = Rearmost.
+  description: Seat horizontal position. 0 = Frontmost.
 
 Height:
   datatype: uint16
   type: actuator
   min: 0
-  max: 1000
   unit: mm
-  description: Seat vertical position. 0 = Lowermost. 1000 = Uppermost.
+  description: Seat vertical position. 0 = Lowermost.
 
 Cushion:
   type: branch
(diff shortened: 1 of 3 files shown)


spec/Cabin/SingleSeat.vspec
```diff
@@ -111,8 +107,8 @@ Lumbar.Height:
   datatype: uint8
   type: actuator
   min: 0
-  max: 255
-  description: Lumbar support position. 0 = Lowermost. 255 = Uppermost.
+  unit: mm
```
erikbosch, line 110: For all other "Height" signals for seats we use absolute position in millimeters, so I assume we want that here as well.

(I see it as a different generic topic if we shall prefer absolute (mm) or relative (percent) values for seat positions - I do not know how positions typically are communicated inside vehicles, by relative or absolute values)

spec/Cabin/SingleSeat.vspec
```diff
@@ -133,9 +129,8 @@ HeadRestraint.Height:
   datatype: uint8
   type: actuator
   min: 0
-  max: 255
   unit: mm
-  description: Height of head restraint. 0 = Bottommost. 255 = Uppermost.
+  description: Height of head restraint. 0 = Bottommost.
```
erikbosch, line 133: Another observation, we sometimes use "lowermost" and sometimes "bottommost". For me the terms seems to be semantically the same - so which one do we prefer?

danielwilms:
approved in meeting 2022/01/25

PR #404, opened 2022-01-26
Renaming "allowed" literals according to guidelines

erikbosch:
This is an attempt to harmonize the use of literals for "allowed" values in the VSS standard, so that they as default use only uppercase, number and underscore. That has in this PR been consistently changed, but there might be some cases where it can be discussed if the recommendation shall be followed or not.

In most cases the change is straight forward, but for some a bit more advanced "translation" has been proposed, e.g. replacing `kWh/100mi` with `KILOWATT_HOURS_PER_100_MILES`.

This PR has not analyzed if the values/signals actually make sense, i.e. no values have been added/removed, the semantic meaning of each is supposed to be the same after the rename. Also `UNKNOWN` is kept as is, it need to be discussed on signal by signal basis whether it makes sense to keep it or not.

Notes:

Private-branch (CARFIT) not updated - private branch so not covered by recommendations. Branches reflecting other standards/solutions might also be a typical case where it might be a reason not to follow the recommendation. If e.g. CARFIT use `abc`, then it might be less useful to require that it shall be `ABC` in VSS.

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -16,7 +16,7 @@ Media:
 Media.Action:
   datatype: string
   type: actuator
-  allowed: [ "unknown", "Stop", "Play", "FastForward", "FastBackward", "SkipForward",  "SkipBackward" ]
+  allowed: ['UNKNOWN', 'STOP', 'PLAY', 'FAST_FORWARD', 'FAST_BACKWARD', 'SKIP_FORWARD', 'SKIP_BACKWARD']
   description: Tells if the media was
 
 Media.Played:
@@ -26,7 +26,7 @@ Media.Played:
 Media.Played.Source:
   datatype: string
   type: actuator
-  allowed: [ "unknown", "SiriusXM", "AM", "FM", "DAB", "TV",  "CD", "DVD", "AUX", "USB", "Disk", "Bluetooth", "Internet", "Voice", "Beep" ]
+  allowed: ['UNKNOWN', 'SIRIUS_XM', 'AM', 'FM', 'DAB', 'TV', 'CD', 'DVD', 'AUX', 'USB', 'DISK', 'BLUETOOTH', 'INTERNET', 'VOICE', 'BEEP']
   description: Media selected for playback
 
 Media.Played.Artist:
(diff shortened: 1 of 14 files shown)


SebastianSchildt (approved):
Looks good.
Found some instances where we might need to discuss semantics, but as you mentioned, let's first just convert the format

spec/Cabin/Infotainment.vspec
```diff
@@ -102,41 +102,41 @@ HMI.CurrentLanguage:
...
   type: actuator
-  allowed: [ "YYYY MM DD", "DD MM YYYY", "MM DD YYYY", "YY MM DD", "DD MM YY", "MM DD YY"]
+  allowed: [YYYY_MM_DD, DD_MM_YYYY, MM_DD_YYYY, YY_MM_DD, DD_MM_YY, MM_DD_YY]
   description: Date format used in the current HMI
 
 HMI.TimeFormat:
   datatype: string
   type: actuator
-  allowed: [ "12HR", "24HR" ]
+  allowed: [FORMAT_12_HR, FORMAT_24_HR]
```
gunnarx, line 111: My preference is that it does not need to be more verbose.  (Do not add FORMAT_)

Also, not sure if we are 100% through the discussion of allowed/enum, but my interpretation here is that normally the _string_ will be transferred, so in that case this is just increasing the bandwidth usage for string-transfer for no reason?

gunnarx (dismissed):

spec/Cabin/Infotainment.vspec
danielwilms: @erikbosch changes it to HR_x and then it can be merged.

erikbosch:
@gunnarx - now PR is updated

erikbosch:
Maybe we shall wait a little with merging this one. I noticed that our tooling do some unexpected conversion of string literals not quoted, well at least unexpected for me.

See https://github.com/COVESA/vss-tools/issues/143

FYI: @gunnarx @danielwilms @SebastianSchildt

erikbosch:
Meeting notes:

- Keep PR on hold until linked issue have been investigated

danielwilms:
meeting approved 15/02/22

PR #405, opened 2022-01-26
Refactoring charge plug signals

erikbosch:
Fixes #376

Merging the two existing signals to a common signal
Inspired by EvConnectorType in
https://android.googlesource.com/platform/hardware/interfaces/+/master/automotive/vehicle/2.0/types.hal
See https://www.evexpert.eu/eshop1/knowledge-center/connector-types-for-ev-charging-around-the-world
for general information on connector types

Based on previous discussions in VSS I do not include any "unknown" literal in allowed values.

PR #406, opened 2022-02-07
Adding attributes for engine code

erikbosch:
Knowing what type of engine(s) the vehicle is equipped with is
important information for garages. This PR introduces two signals
for engine codes, one for combustion engines and one for electric
engines.

For hybrid vehicles a manufacturer might use a common designation
for the whole drivetrain rather than separate engine codes.
I.e. the engine code for combustion engine may include
information also on the electric motors used.

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -1,7 +1,7 @@
 #
+# (C) 2020 Robert Bosch GmbH
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
-# (C) 2020 Robert Bosch GmbH
 #
 # All files and artifacts in this repository are licensed under the
 # provisions of the license provided by the LICENSE file in this repository.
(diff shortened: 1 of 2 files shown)


spec/Powertrain/CombustionEngine.vspec
```diff
@@ -18,6 +18,13 @@
 # Engine attributes
 #
 
+EngineCode:
+  datatype: string
+  type: attribute
+  description: Engine code designation, as specified by vehicle manufacturer.
+  comment: For hybrid vehicles the engine code may refer to the combination of
+           combustion and electric engine.
```
gunnarx, line 26: Multiline.  Pending outcome of #405, but you could also just join the line into one IMHO.

erikbosch:
Meeting notes:

- Consider merging on next meeting if no remarks received

spec/Powertrain/CombustionEngine.vspec
erikbosch: As I understand correct Yaml, see #405. 

Concerning if this shall be 1 or 2 lines it depends on what style guide we want to follow. Merging comment to 1 line gives a line that is about 110 characters long. With 2 lines we are within traditional 80 characters. Maybe we shall specify a recommended maximum line length in our documentation.

PR #408, opened 2022-02-09
Changing documentation so attribute is considered a signal.

### LINKED ISSUE #398: Is an attribute a signal?
This repository/standard is called the COVESA Vehicle Signal Specification. But in the documentation we here and there states that only actuators and sensors are signals, not attributes. I see some problems with this definition - often when we talk about signals we actually mean all three entities (actuators, sensors, attributes). 

Would there be a problem changing the definition of signals so they also include attributes? Of course some documentation like the snippet below will need to be updated. As a side note, standards like VISS does not specifiy any difference between attributes and sensors - there are e.g. nothing that restricts you from subscribing to an attribute as I understand it.

Current description:

https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/

_Leaf nodes of the tree contain metadata describing the data associated to the node. In order to help application developers, who are using the specification, it makes a distinction between signals - in the following as sensor and actuator - and attribute. The difference between a signal and an attribute is that the signal has a publisher (or producer) that continuously updates the signal value while an attribute has a set value that should typically not change more than once per ignition cycle._

erikbosch:
Previously we stated that only sensors and actuators were signals.

Fixes #398

### DIFF
--- a/docs-gen/content/ecosystem/software.md
+++ b/docs-gen/content/ecosystem/software.md
@@ -7,7 +7,7 @@ weight: 3
 Various software frameworks exist to work with VSS data. If something is missing here, please [open an issue](https://github.com/covesa/vehicle_signal_specification/issues) and tell us.
 
 ### W3C VISS2 reference server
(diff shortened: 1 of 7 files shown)


danielwilms (approved):

danielwilms:

PR #409, opened 2022-02-09
Adding line length recommendation.

erikbosch:
Originates from a discussion in #406 

As far as I know we have not specified recommend max line length for e.g.

- *.vspec files
- *.md files

I think it would be good to specify a recommendation. The original version of this PR propose 120 but I am open to any number for both types of files.


danielwilms (approved):

danielwilms:

PR #417, opened 2022-02-25
Refactoring Accuracy signals

erikbosch:
We have identified a need for vertical accuracy.
Also proposing changing name of current Accuracy signals to avoid ambiguity

Also improving description for some other signals, including clarifying for compass that it concerns heading versus geographic north, not magnetic north


danielwilms (approved):

danielwilms:

PR #418, opened 2022-02-25
Adding support for EBD and EBA

erikbosch:
Also reactoring signal names - "IsActive" is ambiguous for some systems. 
VSS has so far used "IsActive" to describe that the system is enabled, i.e. that the system is monitoring conditions and is ready to take actions and "isEngaged" to describe that the system is currently taking action (e.g. regulating brake pressure). But in other contexts (outside VSS) "active" may refer to that the system is taking action. To avoid this it is proposed to instead use the terms "IsEnabled" and "IsEngaged" for systems that typically are in standby for most of the time

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -10,18 +10,16 @@
 #
 # All Advanced Driver Assist System signals
 #
-
-#  ABS
-#  CruiseControl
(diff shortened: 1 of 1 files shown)


danielwilms:
cruise control + isEnabled

danielwilms:
meeting 8/3/22 decided to merge

PR #419, opened 2022-02-28
Updating vss-tools submodule

erikbosch:
To include instantiation fix in mainsteam builds


SebastianSchildt (approved):

danielwilms:

PR #425, opened 2022-03-09
Clarifying combination of instances and aggregate for branches

erikbosch:
Theoretically a branch can have both instances and aggregate defined.
This opens up two interpretations - that aggregate concerns the union of all instances or that aggregate concern individual instances. This PR intends to clarify interpretation

Example:

```
Wunderbaum:
  type: branch
  aggregate: true
  instances:
    - Number[1,100]
  description: All wunderbaums in the vehicle

Wunderbaum.Color:
  datatype: string
  type: sensor
  description: Color of Wunderbaum

Wunderbaum.Fragrance:
  datatype: string
  type: sensor
  description: Fragrance of Wunderbaum
```
The proposed interpretation is that aggregate here means that it makes sense to update color and fragrance for individual Wunderbaums together,  but nothing prevents Wunderbaum1 from being updated separately from e.g. Wunderbaum3.

### DIFF
--- a/docs-gen/content/rule_set/branches.md
+++ b/docs-gen/content/rule_set/branches.md
@@ -37,6 +37,10 @@ This could include background information on the rationale for the branch,
 references to related branches, standards and similar.
 Recommended to start with a capital letter and end with a dot (`.`).
 
+**```instances```** *[optional]*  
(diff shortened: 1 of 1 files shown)


danielwilms (approved):

danielwilms:
decision to merge 2022-03-15

PR #433, opened 2022-04-01
Integrating Ampere hour support

erikbosch:
- Updating vss-tools to get Ampere hours support
- Updating VSS documentation

### DIFF
--- a/docs-gen/content/rule_set/data_entry/data_unit_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_unit_types.md
@@ -33,6 +33,7 @@ kg            | Weight                          | Mass measured in kilograms
 lbs           | Weight                          | Mass measured in pounds
 V             | Electric Potential              | Electric potential measured in volts
 A             | Electric Current                | Electric current measured in amperes
+Ah            | Electric Charge                 | Electric charge measured in ampere hours
 ms            | Time                            | Time measured in milliseconds
 s             | Time                            | Time measured in seconds
 min           | Time                            | Time measured in minutes


docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -33,6 +33,7 @@ kg            | Weight                          | Mass measured in kilograms
 lbs           | Weight                          | Mass measured in pounds
 V             | Electric Potential              | Electric potential measured in volts
 A             | Electric Current                | Electric current measured in amperes
+Ah            | Electric Charge                 | Electric charge measured in ampere hours
```
ppb2020, line 36: ampere hours -> ampere-hours

To clearly distinguish from ampere/hour.

ppb2020 (commented):
Suggested enhancement aligned with standards recommendations.

docs-gen/content/rule_set/data_entry/data_unit_types.md
ppb2020: Interesting: I had started writing a suggestion to convert to ampere-hours (as per my comment above), then noticed that SI seems to forego the hyphen, except in the unit. I updated my comment to instead suggest the unit be changed to "A h" or "A-h", leaving the choice to use "ampere hour" or "ampere-hour" as both seem to be commonly used (but see the discussions on not using the hyphen, below). Not sure why my original comment was posted instead of my updated comment. :-(

https://en.wikipedia.org/wiki/Ampere_hour

Perhaps use "ampere hour" in the text if the decision is to use "A h" as the unit, and use "ampere-hour" in the text if "A-h" is selected as the unit.

As to "kWh" ("kW h" or "kW-h"), I would think so, to be consistent:

https://en.wikipedia.org/wiki/Kilowatt-hour

However, this is a mere suggestion. Other opinions are welcome. Does VSS require use of a standard, such as SI? My read of the set of VSS units is that they seem to follow the SI standard.

Other references:

https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication811e2008.pdf

https://en.wikipedia.org/wiki/International_System_of_Units

https://en.wikipedia.org/wiki/Talk:Ampere_hour#Hyphen

erikbosch: I think it better should be discussed on an upcoming VSS meeting (Tuesdays 16.00 CET/CEST).

But if want to change `Ah` to `A-h` or `A h` it must be as well changed (first) in https://github.com/COVESA/vss-tools/blob/master/vspec/config.yaml, as that is the file used by the tools when parsing vspec files to check that valid units are used. The tooling accept blanks (e.g. `A h`) and hyphens (e.g. `A-h`) so that is no blocker for changing.

VSS does not require SI units, also others if commonly used in the vehicle industry. That is the reason why e.g. `inch` exists, as it is quite common even in mainland Europe to express e.g. wheel/rim-sizes in inches, e.g. a new car in Europe often has 17-inch rims.

SebastianSchildt:
This might be relevant input https://jkorpela.fi/chars/si.html

docs-gen/content/rule_set/data_entry/data_unit_types.md
danielwilms: Discussion on 22/05/03: Even though informal, decision made to keep it as is without ` ` or `-` for practical reasons and broad usage.

danielwilms:
decision to merge: 22/05/03

PR #435, opened 2022-04-06
Adding signal for vehicle broken down status.

erikbosch:
Intention is to cover scenarios where vehicle is "broken down", e.g. it is not possible
to drive the car, it is not safe to drive the car or the car can be further damaged.

Signal description based on Sensoris Hazard.TypeAndConfidence.Type

See https://sensoris.org/wp-content/uploads/sites/21/2021/09/sensoris-specification-v1.2.2-public-1.zip

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -185,6 +185,15 @@ TripMeterReading:
   unit: km
   description: Current trip meter reading.
 
+IsBrokenDown:
+  datatype: boolean
+  type: sensor
+  description: Vehicle breakdown or any similar event causing vehicle to stop on the road,
+               that might pose a risk to other road users.
+               True = Vehicle broken down on the road, due to e.g. engine problems, flat tire, out of gas, brake problems.
+               False = Vehicle not broken down.
+  comment: Actual criteria and method used to decide if a vehicle is broken down is implementation specific.
+
 IsMoving:
   datatype: boolean
   type: sensor


ppb2020 (commented):
Not clear what "ego" means in this context. Should this signal not be applicable to all types of vehicles? Why restrict it to the "ego vehicles"?

Would something like a Vehicle.OperationalMode, with values such as ["factory", "transport", "normal", "service", "incident"] as set by the OEM not be more representative of the state of operation of the vehicle? (Still in factory, in transport to dealership, in normal operation, being serviced, and broken down/in accident. Perhaps adding another entry for "car in sales lot"?)

End comment sentence with a period?

erikbosch:
> Not clear what "ego" means in this context. Should this signal not be applicable to all types of vehicles? Why restrict it to the "ego vehicles"?
> 
> Would something like a Vehicle.OperationalMode, with values such as ["factory", "transport", "normal", "service", "incident"] as set by the OEM not be more representative of the state of operation of the vehicle? (Still in factory, in transport to dealership, in normal operation, being serviced, and broken down/in accident. Perhaps adding another entry for "car in sales lot"?)
> 
> End comment sentence with a period?

Added a period.

"Ego vehicle" here intends to mean the vehicle that VSS concerns, i.e. the current vehicle. One can argue that "ego" here is redundant, as all signals in VSS refer to the "current" vehicle unless otherwise stated.

The proposed description of the signal comes from `Hazard.TypeAndConfidence.Type` in Sensoris that use the description "Vehicle breakdown of ego vehicle or any similar event of ego vehicle stopped on the road, that might pose a risk to other road users." 

The term ego vehicle is also used in other publications like:

https://www.etsi.org/deliver/etsi_ts/101500_101599/10153901/01.01.01_60/ts_10153901v010101p.pdf
https://www.gpsworld.com/intelligent-transportation-systems-require-the-ego-vehicle/
https://www.bsigroup.com/en-GB/CAV/cav-vocabulary/ego-vehicle/

... but I have no problem removing "ego" if we see it as redundant or that it can be interpreted in other ways. Concerning having a generic "OperationMode" instead - I actually do not know if there are any industry de-facto standard on what states a vehicle has, e.g. if the list of states proposed by you would make sense for most OEMs. For example - I do not know if vehicles typically have a generic service mode, or if this rather is something that is controlled on individual subsystem level. I suggest this to be discussed on an upcoming VSS meeting (Tuesday at 19.00 CET), feel free to join.

danielwilms:
Discussion 22/05/03: remove ego as all signals are meant to be related to a specific vehicle.

SebastianSchildt (approved):

erikbosch:
Decision 220510: Merge!

PR #437, opened 2022-04-06
Adding additional information for ADAS systems

erikbosch:


spec/ADAS/ADAS.vspec
```diff
@@ -200,6 +200,45 @@ ESC.IsEngaged:
   type: sensor
   description: Indicates if ESC is currently regulating vehicle stability. True = Engaged. False = Not Engaged.
 
+ESC.IsStrongCrossWindDetected:
+  datatype: boolean
+  type: sensor
+  description: Indicates if ESC system is detecting strong cross winds.
+               True = Strong Cross winds detected.
```
ppb2020, line 207: Other descriptions provide information for False as well. Granted, this is a bit redundant, but not providing it seems inconsistent (see line 201 above).

spec/ADAS/ADAS.vspec
```diff
@@ -200,6 +200,45 @@ ESC.IsEngaged:
...
+               
+ESC.RoadFriction:
+  type: branch
+  description: Road friction values reported by ESC system.
```
ppb2020, line 211: by ESC -> by the ESC

spec/ADAS/ADAS.vspec
```diff
@@ -200,6 +200,45 @@ ESC.IsEngaged:
...
+
+ESC.RoadFriction.MostProbable:
+  datatype: float
```
ppb2020, line 214: Is this level of precision useful given this is a percentage? That is, would 99.111% be much different from 99%.

spec/ADAS/ADAS.vspec
```diff
@@ -200,6 +200,45 @@ ESC.IsEngaged:
...
+  type: sensor
+  unit: percent
+  min: 0
+  max: 100
+  description: Most probable road friction, as calculated by ESC system.
```
ppb2020, line 219: by ESC -> by the ESC

Here an throughout.

spec/ADAS/ADAS.vspec
erikbosch: As always difficult to know what resolution that is needed for future use-cases, float proposed to allow for higher precision than int. Of course, the difference between 99.1% and 99% is not that big, but the relative difference between 0.4% and 0% is bigger. And as with all VSS-signals, the actual accuracy/precision used in transmission can vary between implementations.

danielwilms:
decision to merge: 22/05/03

PR #438, opened 2022-04-06
Adding weather environment signals

erikbosch:


spec/Vehicle/Vehicle.vspec
```diff
@@ -191,6 +191,24 @@ AmbientAirTemperature:
   unit: celsius
   description: Ambient air temperature outside the vehicle.
 
+AmbientHumidity:
+  datatype: float
```
ppb2020, line 195: Is the level of precision of float required here?

Would a uint8 not be sufficient?

Apologies if these kinds of questions have already been raised and addressed in your design review process.

spec/Vehicle/Vehicle.vspec
```diff
@@ -191,6 +191,24 @@ AmbientAirTemperature:
   unit: celsius
   description: Ambient air temperature outside the vehicle.
 
+AmbientHumidity:
```
ppb2020, line 194: The next signal is named OutsideLightIntensity. Should this signal be OutsideAmbientHumidity?

How about Exterior as a name instead of Outside? Or even making Exterior a branch? As in:

Vehicle.Exterior.AmbientHumidity
Vehicel.Exterior.LightIntensity

spec/Vehicle/Vehicle.vspec
```diff
@@ -191,6 +191,24 @@ AmbientAirTemperature:
...
+               0 = Dry, 100 = Air fully saturated.
+
+OutsideLightIntensity:
+  datatype: float
+  type: sensor
+  unit: percent
+  min: 0
+  max: 100
+  description: Light intensity outside the vehicle.
+               0 = No light detected, 100 = Fully lit.
```
ppb2020, line 210: "Fully lit" is interesting as I'm not clear on how this can be assessed. (Relative to the sun's intensity?)

spec/Vehicle/Vehicle.vspec
erikbosch: In some standards they use better precision than int. As an example, in J1939 PGN 65164 SPN 354 "Relative Humidity, Measures humidity of combustion air prior to entry into turbocharger", humidity is represented as 1 byte, with 0.4% per bit. If we would use this J1939 signal as source for this VSS signal (or in the other direction) both 249 (99.6%) and 250 (100%) would be represented as 100% in VSS, i.e. some precision would be lost.

Using float minimizes loss of precision when converting between e.g. VSS-representation and J1939-representation

spec/Vehicle/Vehicle.vspec
erikbosch: I understand your point that the names proposed are not that aligned. We already have the signal `AmbientAirTemperature`. I can see a benefit in using the same "prefix" for all 3 signals (temperature, humidity, light intensity), i.e. that all of them start with "Ambient", "Outside" or "Exterior". I have no strong opinion on what to use, or if it shall be a branch or not. One (minor) advantage sticking to "Ambient" is that we avoid the need to rename `AmbientAirTemperature`.

I would have no problems changing `OutsideLightIntensity` to `AmbientLightIntensity`, or to use "Outside" for all, e.g. `OutsideAirTemperature` and `OutsideHumidity`. But `OutsideAmbientHumidity` is a bit redundant, or?

spec/Vehicle/Vehicle.vspec
erikbosch: That is the reason for the comment below. We see it as useful to have light intensity represented, as that can be used to control e.g. Daytime Running Lights and brightness of displays. A vehicle may e.g. decide to put on low beam if light intensity falls below 50%, or shift navigation to night mode on navigation if intensity falls below 30%, but how light intensity is calculated is likely vehicle dependent, i.e. there is as far as I know no de-facto industry standard saying that e.g. any Lux value above 20000 Lux shall be considered 100% and 1000 Lux be considered 50%. 

So in short, this leaves it to the OEM to decide on how to calculate brightness. This is similar to other signals like `Brake.IsBrakesWorn` where we also leave the logic and detailed meaning to the OEM.

spec/Vehicle/Vehicle.vspec
danielwilms: I sort of like the idea of an exterior branch to group the signals, even though it would mean breaking the existing temperature signal.

ppb2020: Vehicle.Exterior.AmbientHumidity
Vehicle.Exterior.AmbientAirTemperature
Vehicle.Exterior.AmbientLightIntensity

or

Vehicle.Exterior.Humidity
Vehicle.Exterior.AirTemperature
Vehicle.Exterior.LightIntensity

erikbosch:
Adapted PR to the meeting discussion to have a separate branch. Then I also took the chance to put the signals in a separate file.

danielwilms:
decision to merge: 22/05/03

PR #439, opened 2022-04-06
Extending GNSS data

erikbosch:
Vehicles typically report position related to actual position of GNSS receiver.
In some cases it is necessary to know where the GNSS receiver is located.

Augmentation data inspired by specification in Sensoris VehiclePositionAndOrientation.NavigationSatelliteSystemStatus.FixType from https://sensoris.org/wp-content/uploads/sites/21/2021/09/sensoris-specification-v1.2.2-public-1.zip

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -370,7 +370,7 @@ CurrentLocation.Latitude:
   min: -90
   max: 90
   unit: degrees
-  description: Current latitude of vehicle according to WGS 84.
+  description: Current latitude of vehicle in WGS 84 geodetic coordinates, as measured at the position of GNSS receiver antenna.
 
 CurrentLocation.Longitude:
   datatype: double
@@ -378,7 +378,7 @@ CurrentLocation.Longitude:
   min: -180
   max: 180
   unit: degrees
-  description: Current longitude of vehicle according to WGS 84.
+  description: Current longitude of vehicle in WGS 84 geodetic coordinates, as measured at the position of GNSS receiver antenna.
 
 CurrentLocation.Heading:
   datatype: double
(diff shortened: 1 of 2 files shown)


spec/Vehicle/Vehicle.vspec
```diff
@@ -370,15 +370,15 @@ CurrentLocation.Latitude:
   min: -90
   max: 90
   unit: degrees
-  description: Current latitude of vehicle according to WGS 84.
+  description: Current latitude of vehicle according to WGS 84, as measured at position of GNSS receiver antenna.
```
ppb2020, line 373: Perhaps:

as measured at the position of the GNSS receiver antenna

or

as measured by the GNSS receiver antenna

or

as measured at the GNSS receiver antenna

erikbosch: Taking first variant for now, thanks!

spec/Vehicle/Vehicle.vspec
```diff
@@ -370,15 +370,15 @@ CurrentLocation.Latitude:
   min: -90
   max: 90
   unit: degrees
-  description: Current latitude of vehicle according to WGS 84.
+  description: Current latitude of vehicle according to WGS 84, as measured at the position of GNSS receiver antenna.
```
danielwilms, line 373: > according to WGS 84

maybe better: "in WGS84 geodetic coordinates"

danielwilms (approved):
just minor change to be more clear.

spec/Vehicle/Vehicle.vspec
erikbosch: Works fine for me, but it seems that they use `WGS 84` (with space) in the standard rather than `WGS84` (without space) so I will keep the space

danielwilms:
decision to merge: 22/05/03

PR #442, opened 2022-05-09
Improving charging and battery signals

erikbosch:
This PR contains various improvement suggestions to battery and charging signals.

Major changes:
- Battery identifiers added
- Extension of temperature signals
- State of health signal added
- NetCapacity changed to sensor
- Many Charging signals refactored to allow separate measurements for each phase.
- ChargePlugType now use an array, as vehicles mioght have multiple charge plugs
- Some signals for power loss (to heat) added

### DIFF
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -1,7 +1,7 @@
 #
+# (C) 2020 Robert Bosch GmbH
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
-# (C) 2020 Robert Bosch GmbH
 #
 # All files and artifacts in this repository are licensed under the
 # provisions of the license provided by the LICENSE file in this repository.
@@ -11,6 +11,18 @@
 # Signals and attributes related to the traction battery in vehicles with electrical powertrain.
 # For some vehicles the traction battery is referred to as the high voltage battery.
 #
+
+Id:
+  datatype: string
+  type: attribute
+  description: Battery Identification Number as assigned by OEM.
+  comment: This could be serial number, part number plus serial number, UUID, 
+           or any other identifier that the OEM want to use to uniquely identify the battery individual. 
+
+ProductionDate:
+  datatype: string
+  type: attribute
+  description: Production date of battery in ISO8601 format, e.g. YYYY-MM-DD.
 
 IsPowerConnected:
   datatype: boolean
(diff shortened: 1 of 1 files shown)


erikbosch:
At the meeting yesterday there were a question on backward incompatible changes, here I try to summarize them including a rationale:

- `Vehicle.Powertrain.TractionBattery.Temperature` was previously a sensor-signal, now a branch to contain 3 temperature signals `Min`, `Max` and `Average` as a battery pack often have multiple temperature sensors, and it was ambiguous what the old sensor value actually meant, i.e. did it refer to avergae temperature or the maximum temperature?
- `Vehicle.Powertrain.TractionBattery.StateOfCharge.Target` removed , considered to be the same as `Vehicle.Powertrain.TractionBattery.Charging.ChargeLimit`, i.e. we could not identify any use-case where both are needed/used
- `Vehicle.Powertrain.TractionBattery.ReferentVoltage` removed. We had problems finding a definition for what is meant with referent voltage and how it differs from nominal voltage. We already have `Vehicle.Powertrain.TractionBattery.NominalVoltage`, and propose a new signal  `Vehicle.Powertrain.TractionBattery.MaxVoltage`
- `Vehicle.Powertrain.TractionBattery.Charging.ChargePlugType` has type changed to `string[]`. Reason is that a vehicle may have multiple plugs/inlets.
- `Vehicle.Powertrain.TractionBattery.Charging.ChargeCurrent` and ` Vehicle.Powertrain.TractionBattery.Charging.ChargeVoltage` converted from sensors to branches. The old signals were ambiguous as they did not state if the value was as measured at the charge plug/inlet, or as measured at the battery. Now separate signals has been created for values at charge inlet and at battery, like `Vehicle.Powertrain.TractionBattery.CurrentCurrent` and `Vehicle.Powertrain.TractionBattery.Charging.ChargeCurrent.DC`. For charge current separate signals are introduced for DC and AC phases.

spec/Powertrain/Battery.vspec
```diff
@@ -12,6 +12,16 @@
 # For some vehicles the traction battery is referred to as the high voltage battery.
 #
 
+Id:
+  datatype: string
+  type: attribute
+  description: Battery Identification Number as assigned by OEM.
```
ppb2020, line 18: What is this ID meant to represent? Some serial number? A location in the vehicle? A differentiator in case there are more than one batteries in the vehicle?

erikbosch: Description could be improved, the intention is not to specify that this is battery "2" in the vehicle, but rather a (more or less) unique identifier of the battery instance mounted in the vehicle. E.g. serial number, UUID, part_number+serial_number or similar.

As far as I know there are no standard similar to Vehicle Identification Number (VIN) that assures that battery identifiers are globally unique, but I assume that the id (possibly combined with part number) makes it posssible to identify exactly which battery which is mounted.

I will double check with my colleagues and propose an update.

erikbosch: Added a comment to clarify that it concerns a "unique" identifier. But the level of uniqueness is up to the OEM.

danielwilms:
Decision to merge 2022/05/24

PR #444, opened 2022-05-10
Introducing signal for connectivity available

erikbosch:
A possible use-case for this signal is onboard services that relies on connectivity, like FOTA and possibly navigation. A service may check the `IsConnectivityAvailable` signal before e.g. enabling the button offering the user to download new maps or check for updates.

I assume implementation will vary. Some vehicles may only check if WiFi or Cellular connection is connected or even only available, others may more or less frequently check that also important destinations can be reached (e.g. OEM-cloud, maps-server, FOTA-server). Possibly additional signals might need to be added in the future, if e.g. a service only wants to offer certain services when WiFi (or other high speed or low cost connection) is available.

### DIFF
--- a/spec/Vehicle/Connectivity.vspec
+++ b/spec/Vehicle/Connectivity.vspec
@@ -1,0 +1,23 @@
+#
+# (C) 2022 Robert Bosch GmbH
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# Connectivity data
+#
+# This file contains signals related to connectivity between Vehicle and X (V2X).
+# Initially it contains only a single signal specifying if connectivity available.
+# May later be extended with information on type of connectivity available (e.g., Wi-Fi, Cellular), signal strength,
+# and throughput.
+#
+
+IsConnectivityAvailable:
+  datatype: boolean
+  type: sensor
+  description: Indicates if connectivity between vehicle and cloud is available.
+               True = Connectivity is available. False = Connectivity is not available.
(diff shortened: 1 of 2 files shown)


spec/VehicleSignalSpecification.vspec
```diff
@@ -184,10 +184,10 @@ Vehicle.Exterior:
 #include Vehicle/Exterior.vspec Vehicle.Exterior
 
 #
-# Service information
+# Connectivity information
```
ppb2020, line 187: The description for this commit is that Vehicle.Connectivity is being added. However, this change seems to be replacing Vehicle.Service with Vehicle.Connectivity. Will Vehicle.Service and its components (such as DistanceToService and TimeToService) no longer be supported?

As well, Connectivity seems to be very specific to ability to connect. Will this still be applicable once further signals are added, such as Wi-Fi SSID, Cellular IMEI, etc.?

spec/Vehicle/Connectivity.vspec
```diff
@@ -0,0 +1,23 @@
...
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# Connectivity data
+#
+# This file contains signals related to connectivity between Vehicle and X (V2X).
+# Initially it contains only a single signal specifying if connectivity available.
+# May later be extended with information on type of connectivity available (e.g., WiFi, Cellular), signal strength,
```
ppb2020, line 12: Wifi -> Wi-Fi

Reference: https://www.wi-fi.org/

danielwilms:
Decision to merge 2022/05/24

PR #446, opened 2022-05-11
Aligning signals using percent

erikbosch:
- Added max:100 (if not already where) for signals where I assume logical max value shall be 100
- Added min:0 (if not already where) for signals where I assume logical min value shall be 0
  (Exception is unsigned data types, where minimum value 0 is given implicitly)
- Added "percent" as data unit type in one case where it was missing
- (Exception: Did not update the OBD signals, bigger effort to check actual limits specified in OBD standard)


danielwilms:
Decision to merge 2022/05/24

Issue #447, opened 2022-05-17
Instance handling of axles/wheels


erikbosch:
The wheel-representation in VSS as of today (shown below) does not work that well for certain vehicles.

- Two-wheelers (motorbikes) typically do not have `left` and `right` wheel but just a single wheel on each axle.
- Trucks often have twin wheels on some axles, i.e. 4 wheels on some axles. 

```
Axle:
  instances:
    - Row[1,2]
  type: branch
  description: Axle signals

Axle.Wheel:
  instances: ["Left","Right"]
  type: branch
  description: Wheel signals for axle
```

A possible change could be to address wheels similar to seats, i.e. instead of using left/right use positions:

```
Wheel:
  instances:
    - Row[1,2]
    - Pos[1,4]
  type: branch
  description: Wheel signals
```

But if we would do that we would open up for another question - and that is to review of what data we need for each wheel. If we look at https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Chassis/Wheel.vspec one can identify some areas where a change could be useful:

- `Brake.FluidLevel` is today given on wheel level, but how frequent is it that vehicle have separate brake fluid containers for individual wheels? Could likely be moved to "axle"-level, or a new `BrakeSystem` branch that can have one or more instance, e.g. typically 1 instance for a passenger car and 2 systems for a motorbike.
- `Brake.PadWear` - is today on wheel level, which will not work well if we start supporting twin wheels as they typically share the the same brake disc/drum/pads. I do not know how many vehicles that actually measure individual pad wear, or if they mostly just estimate padwear based on driving pattern. If they estimate padwear then the signal could likely be on axle-level. If measured, do we rather need a branch `BrakeUnit` with row/pos instantiations, i.e an instantiation that does not necessarily needs to be the same as wheel instantiation? One brake unit could then server multiple wheels (typical truck use case), or multiple brake units could serve a single wheel (typical motorbike front wheel use case).
- `Tire.Pressure`  and `Tire.Temperature` are likely to be measured for each wheel individually.

Thoughts?

danielwilms:
```
Wheel:
  instances:
    - Row[1,2]
    - Pos[1,4]

As for the proposal above, I'm not quite sure, if we should generalise more in the standard catalogue with regard to the instantiation. I guess there will be more things, which have to be adopted for a motor cycle. As we are now so close, I would rather see a specific layer file for motor cycles as an example or even as `official` release.

[...]

PR #450, opened 2022-05-31
Motorbike signals as additional layer

danielwilms:
As discussed previously (#447), mixing different kinds vehicles
in one vspec structure might lead to too generic structures
and confusion. One way to achieve such variety is through
layers. This is a first proposal on how it could look like,
open for discussion.

### DIFF
--- a/overlays/profiles/motorbike.vspec
+++ b/overlays/profiles/motorbike.vspec
@@ -1,0 +1,45 @@
+#
+# (C) 2022, Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# This file contains adoptions to the main spec concering motorbikes.
+# It mainly addresses istantiation issues (e.g. wheels) and 
+# additional signals, which are motorbike specific (e.g. handlebars).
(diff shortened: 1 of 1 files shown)


layers/motorbike/motorbike_layer.vspec
```diff
@@ -0,0 +1,41 @@
...
+Vehicle.Chassis.Axle:
+    type: branch
+
+# Changed instantiation
+Vehicle.Chassis.Axle.Wheel:
+    type: branch
+    instances: ["Center"]
+
+# Branches used for the structure, not adopted in any way
+Vehicle.Cabin:
```
erikbosch, line 29: Maybe we should discuss how useful it is to keep the "cabin" branch for motorbikes. I believe only a very small subset of the signals (like seat heating) would be reusable. The term "cabin" is not intuitive for most bikes, but relevant for some like https://en.wikipedia.org/wiki/BMW_C1

danielwilms:
- add comments about stability
- better to merge fast and evolve than do it on the initial PR

danielwilms:
@erikbosch: could you do a quick sanity check please? I you gice thumbs up, I'd merge

erikbosch (approved):

erikbosch:
Looks good @danielwilms

PR #465, opened 2022-07-06
Restructured exterior lights to make use of instances. Enables to add…

grg-ffb:
Restructured the lights-branch so it follows more the tree nature of vss. Also it makes it easier to add additional leafs for lights. See the defect flag I added as an example. Other possible values could be an intensity value in percent.


grg-ffb:
> I like it!
> 
> I am just not sure whether indicator lights (at least I interpreted the "Left" and "Right" there always as turn signals?) might need a special treatment.
> 
> (Because I expect semantic to be different, I think VSS needs a signal that says whether a turn singal is.... well "signaling", while just modelling it as a "generic light" might indicate it will regularly switch between on/off (maybe need both?)

Good point. I could adjust the description to 'on or signaling' or add a new type of lights beside genericLights like flashingLights or something like this. Do you have a recommendation how to proceed?

jdacoello:
I think that further definition of instances might be needed to avoid redundant concepts. For example, I see repetition of specific components/concepts in:

```
instances: ["HighBeam", "LowBeam", "Running", "Backup", "Parking", "Brake", "RearFog", "FrontFog", "Hazard", "LeftIndicator", "RightIndicator"]
```

- High**Beam**
- Low**Beam**

- Rear**Fog**
- Front**Fog**

- Left**Indicator**
- Right**Indicator**

Does it make sense to have `Beam`, `Fog`, `Indicator` as instances of `Lights`, as follows:
```
Lights:
    instances: ["Beam", "Running", "Backup", "Parking", "Fog", "Hazard", "Indicator"] 
   ...
```

And then defining further instances for each of them, as follows:
```
Lights.Beam:
    instances: ["High", "Low"]
    ...
Lights.Fog:
    instances: ["Rear", "Front"]
    ...
Lights.Indicator:
    instances: ["Left", "Right"]
   ...
```
?


That way, you keep the model flexible with the option to add more instances to each type of light. The cost would be a longer name with extra branches.

spec/Body/BackupLights.vspec
```diff
@@ -0,0 +1,12 @@
...
+
+IsOn:
+  datatype: boolean
+  type: actuator
+  description: Indicates if light is on or off. True = On. False = Off.
+
+HasDefect:
+  datatype: boolean
+  type: sensor
+  description: Indicates if light is defect. True = Light is defect. False = Light has no defect.
```
ppb2020, line 12: is defect -> is defective

spec/Body/BackupLights.vspec
```diff
@@ -0,0 +1,12 @@
+# (C) 2022 BMW Group - All rights reserved.
+# AUTHOR: Gerrit Grotewold BMW Group;
+
+IsOn:
+  datatype: boolean
+  type: actuator
+  description: Indicates if light is on or off. True = On. False = Off.
```
ppb2020, line 7: This uses the singular, but the file name uses the plural. Consider renaming the file.

For some types of lights that are blinking, this terminology can be confusing. Hazrds lights, for example, blink on and off. Is this signal intended to be sent every time the light comes on and goes off? Probably not.

Hence, perhaps this should be renamed to IsActivated?

spec/Body/Body.vspec
```diff
@@ -161,11 +161,53 @@ Windshield.WasherFluid.Level:
...
-  description: All lights.
+  instances: ["Beam", "Running", "Backup", "Parking", "Brake", "Fog", "Hazard", "Indicator"]
+  description: Exterior lights.
 
-# Include lights specification and attach it to the Lights branch
-#include ExteriorLights.vspec Lights
+Lights.Beam:
+  type: branch
+  instances: ["Low","High"]
+  description: Beam lights
```
ppb2020, line 170: Perhaps provide more description of what is a "beam" light? Does this include high and low beams?

Add a period at end of sentence.

spec/Body/Body.vspec
```diff
@@ -161,11 +161,53 @@ Windshield.WasherFluid.Level:
...
+# Include lights specification and attach it to the Lights.Beam branch
```
ppb2020, line 171: Comment seems redundant a it describes the obvious from the code on the next line?

spec/Body/Body.vspec
```diff
@@ -161,11 +161,53 @@ Windshield.WasherFluid.Level:
...
+#include BeamLights.vspec Lights.Beam
+
+Lights.Running:
+  type: branch
+  description: Running lights
```
ppb2020, line 176: Provide a better description?

spec/Body/Body.vspec
```diff
@@ -161,11 +161,53 @@ Windshield.WasherFluid.Level:
...
+  instances: ["Rear","Front"]
+  description: Fog lights
+# Include lights specification and attach it to the Lights.Fog branch
+#include FogLights.vspec Lights.Fog
 
+Lights.Hazard:
+  type: branch
+  description: Hazard lights
+# Include lights specification and attach it to the Lights.Hazard branch
+#include HazardLights.vspec Lights.Hazard
```
ppb2020, line 203: Hazard lights blink. Need to be clear what is meant by hazard lights being on. That is, do we mean "activated" or do we expect to receive a cycling signal telling us on/off/on/off/...

spec/Body/Body.vspec
```diff
@@ -161,11 +161,53 @@ Windshield.WasherFluid.Level:
...
+
+Lights.Indicator:
```
ppb2020, line 205: Another type of light that cycles between on and off.

spec/Body/FogLights.vspec
```diff
@@ -0,0 +1,12 @@
+# (C) 2022 BMW Group - All rights reserved.
+# AUTHOR: Gerrit Grotewold BMW Group;
+
+IsOn:
```
ppb2020, line 4: All of these are repetitions. Use instances instead?

spec/Body/BackupLights.vspec
grg-ffb: thank you. Is part of new commit.

spec/Body/BackupLights.vspec
grg-ffb: thank you. your feedback is part of latest commit.

spec/Body/Body.vspec
grg-ffb: Thank you for your feedback. See instances of beam-lights for low and high beams.

spec/Body/SignalingLights.vspec
```diff
@@ -0,0 +1,12 @@
+# (C) 2022 BMW Group - All rights reserved.
+# AUTHOR: Gerrit Grotewold BMW Group;
+
+IsOn:
+  datatype: boolean
+  type: actuator
+  description: Indicates if light is on or off. True = On. False = Off.
```
SebastianSchildt, line 7: In case of singaling lights, should the description rather be something like inidicates wheter light is "signaling" ?

danielwilms:
looking good. two small things:
1. In the documentation we say that boolean should start with an `Is` and for `Has` a warning is raised. @erikbosch, @SebastianSchildt, IMHO `HasDefect` makes sense here. We should discuss, if we recommend that officially.
2. @UlfBj, the binary exporter fails and the others don't. Could you check, what the problem is?

SebastianSchildt:
I remember we had the Is<Something> and Has<Something> discussion and at that point decided to not add has, because we did not have an example in the core catalogue. That might have changed now.

However, I can not find hard evidence of that discussion and the outcome right now. Maybe it is just in my mind :)  Related were 
https://github.com/COVESA/vss-tools/pull/128
and
https://github.com/COVESA/vehicle_signal_specification/pull/389

Generally speaking I would be fine including "HasXXX" as alternate best practices for boolean signals (and is trivial to fix in tools). (Although in this case "IsDefect" would also be fine I guess).

spec/Body/StaticLights.vspec
```diff
@@ -0,0 +1,12 @@
+# (C) 2022 BMW Group - All rights reserved.
```
erikbosch, line 1: In other parts of code BMW is referred to as "Bayerische Motoren Werke Aktiengesellschaft (BMW AG)". Is it intentional to use a new name, if so, shall the old copyright be updated?

spec/Body/StaticLights.vspec
```diff
@@ -0,0 +1,12 @@
...
+# AUTHOR: Gerrit Grotewold BMW Group;
```
erikbosch, line 2: We have previously not included "Author" in VSS files. I do not really have a strong opinion but I think a short discussion could be useful. Do we want "author" to be included, if so shall it be optional, recommended or mandatory going forward. If we want to use "author" I assume it will be handled like copyright, over time additional names will be added based on additions to the file.

spec/Body/Body.vspec
```diff
@@ -161,11 +161,57 @@ Windshield.WasherFluid.Level:
...
+
+Lights.Fog:
+  type: branch
+  instances: ["Rear","Front"]
+  description: Fog lights.
+#include StaticLights.vspec Lights.Fog
+
+Lights.LicensePlate:
+  type: branch
+  instances: ["Rear","Front"]
```
erikbosch, line 196: Is there any country that mandates that front license plate is illuminated? And/Or any vehicles that support it?

spec/Body/Body.vspec
```diff
@@ -161,11 +161,57 @@ Windshield.WasherFluid.Level:
...
+  description: License plate lights.
+#include StaticLights.vspec Lights.LicensePlate
+
+Lights.Brake:
+  type: branch
+  instance: ["Left","Right","Top"]
```
erikbosch, line 202: What is the use case of having separate instances here? In which cases are they controlled separately?

spec/Body/Body.vspec
```diff
@@ -161,11 +161,57 @@ Windshield.WasherFluid.Level:
...
+
+Lights.Hazard:
+  type: branch
+  description: Hazard lights.
+#include SignalingLights.vspec Lights.Hazard
 
+Lights.Indicator:
+  type: branch
+  instances: ["Left","Right"]
+  description: Indicator lights.
```
erikbosch, line 213: To avoid misunderstandings, would it be an idea to include alternative terms in the description like:

`description: Indicator (turn signal) lights.`

 Or possibly use the term DirectionIndicator for the signal as used in e.g. https://unece.org/fileadmin/DAM/trans/main/wp29/wp29regs/2015/R048r12e.pdf

I actaully do not know if the terms we are using for the signal names are based on any standard, to me they seem to be a bit random.

spec/Body/Body.vspec
jdacoello: @erikbosch Does it really matter if the license plate light is mandatory or not in a particular country? I would say not. As far as I understand, the idea is that VSS provides the means to name and access the data point of interest. In this case, it would be sufficient that there exist one car that has this type of light to model it.

spec/Body/Body.vspec
jdacoello: Several instances of the same type of light might have different statuses at some point. For example, if one of the brake lights is defective, it would differ from the others.

spec/Body/Body.vspec
erikbosch: Agree, but do we know of any vehicle supporting it today. Just want to make sure that we do not add an instance that likely not will be used.

spec/Body/Body.vspec
erikbosch: But that is true also for others like `Lights.Fog`, the left front fog light might be ok, but the right front fog light might be broken. For the others (except turning indicators) we do not separate between left and right.

erikbosch: By the way, a syntax error, it shall be `instances:`, not `instance:`

erikbosch: And the lack of `description:` on this node causes the binary exporter to crash

spec/Body/StaticLights.vspec
erikbosch: Meeting notes: remove copyright

spec/Body/Body.vspec
erikbosch: Meeting notes:

- Possible use case example: Left brake light not used if left turn indicator is active at the same time. If so, there is a business case to control them separately.
- Possibly start without separation, then add separation later if needed

Adnan to check PR status with author

adobekan:
Check changes:
@adobekan

erikbosch (approved):

spec/Body/Body.vspec
```diff
@@ -161,11 +161,55 @@ Windshield.WasherFluid.Level:
 ##
 Lights:
   type: branch
-  description: All lights.
+  instances: ["Beam", "Running", "Backup", "Parking", "Brake", "Fog", "Hazard", "Indicator"]
```
erikbosch, line 164: This row shall be removed, right?

erikbosch (changes requested):
See comment on line 164

spec/Body/Body.vspec
```diff
@@ -161,11 +161,55 @@ Windshield.WasherFluid.Level:
...
+Lights.Hazard:
+  type: branch
+  description: Hazard lights.
+#include SignalingLights.vspec Lights.Hazard
 
+Lights.DirectionIndicator:
+  type: branch
+  instances: ["Left","Right"]
+  description: Indicator lights.
+#include SignalingLights.vspec Lights.Indicator
```
erikbosch, line 212: Change `Lights.Indicator`to `Lights.DirectionIndicator`

grg-ffb: Thank you for that input!

adobekan (approved):

erikbosch (approved):
Ready to merge as I see it.

erikbosch:
@adobekan - we can merge this one now

PR #469, opened 2022-08-04
Add tar-file command to Makefile

erikbosch:
Also remove obsolete deploy command and update gitignore

Background for this PR is to simplify our release process by having a "make tar" command that creates the *.tar.gz file attached to our releases. The intention (for now) is that the maintainer preparing the release manually will use the command as part of the release process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,4 @@
 *.id
 
 # Generated files
-vss_rel_*.cnative
-vss_rel_*.csv
-vss_rel_*.fidl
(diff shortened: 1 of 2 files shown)


erikbosch:
@danielwilms - added a tar target to the Makefile as an idea to standardize what we include in releases. Also adapting .gitignore (it was a bit outdated) and removing "deploy"-target (I do not think it is used any longer as we do not create any snapshot releases). Could you check if you have any comments

PR #470, opened 2022-08-04
Updated logo and color theme

danielwilms:
* added the new logo
* updated color theme to blue

This is how it looks right now:

Screen Shot 2022-08-04 at 16 51 04


SebastianSchildt (approved):
Still think, that was the worst possible logo choice (except keeping the old one), but hey democracy at work 😁

Thanks for making this PR. Approve 👍

adobekan (approved):
Hello darkness, my old friend. 
I like the icon, but not color scheme that much. We should keep other options as backup, as @SebastianSchildt  it is democracy. You never know! ;)

danielwilms:
I like it 😂

erikbosch:
Maybe we can use the color scheme as code names for new releases - 3.0 will be the "blue" release, for 4.0 we can select a different color scheme. I think this is a great step forward, as always it can be changed in the future.

PR #480, opened 2022-08-22
Improve trunk/hood documentation

erikbosch:
I got some internal questions on the difference between "front trunk" and "hood". This PR tries to improve the documentation based on the discussion we had when introducing front trunk.

Example: Is the luggage compartment at the front of a Porsche 911 considered to be a "hood" or "front trunk"? This comment tries to specify that it preferably shall be considered as "front trunk" rather than "hood" (as it gives access to a luggage compartment and not an engine compartment).


erikbosch:
Meeting decision to merge

PR #483, opened 2022-08-24
Buildcheck update

erikbosch:
This seems at least for now to solve the build problems - have run it a few times without experiencing any problems. As the error occurred quite recently I expect that we have succeeded to find a spot where pipenv is not as deterministic as one would like. The use of the `--sequential flag` seems to help, but I am no pipenv expert, so there might be other reasons.

https://buildmedia.readthedocs.org/media/pdf/pipenv/stable/pipenv.pdf

_Installation is intended to be as deterministic as possible — use the `--sequential flag` to increase this, if experiencing issues._

Also improving verbosity on the part where the build have failed in the past.


danielwilms (approved):
good for me. let's give it a try.

Issue #490, opened 2022-09-13
Status and Handling of overlays/profiles


erikbosch:
As of know the status of the overlays in VSS is unclear. We need to clarify how to communicate whether overlays are considered to be part of official VSS, and/or how extensions of them shall be handled concerning e.g. testing. Also needs to be aligned with #473. This should better be discussed before merging #487.

Target:  Give comments until next meeting 2022-09-20.

adobekan:
Notes 04.10.22
- Introduce guidelines and governance on how to handle overlays.
- Define ownership of the overlays and structure e.g. contrib and deprecation.

erikbosch:
Closed now when #491 is merged

Issue #488, opened 2022-09-07
Contribution guidelines?


erikbosch:
We do not have (as far as i know) a hands-on guide on how you contribute something to VSS, how the processes looks like including what you should think of to increase the possibility that your PR will get merged.

I think that this could be a good addition, either in the Github wiki or possibly in a `CONTRIBUTING.md` file. Any thoughts?

In e.g. Eclipse it is quite common to have a `CONTRIBUTING.md` file, see e.g [Kuksa.val](https://github.com/eclipse/kuksa.val/blob/master/CONTRIBUTING.md)

erikbosch:
Meeting notes:
Stephen - likes CONTRIBUTING .md files. Possibly refer to general COVESA contribution guidelines.

erikbosch:
#491  has been merged, this one can be closed

PR #487, opened 2022-08-30
Extend motorbike profile and include profiles in builds

erikbosch:
This PR contains some signals relevant to motorbikes.
Some of them might be of general interest and may be moved to
"standard vss" later.

Signed-off-by: Erik Jaegervall <erik.jaegervall@se.bosch.com>

Issue #473, opened 2022-08-05
Intended content of a release


erikbosch:
The [latest release](https://github.com/COVESA/vehicle_signal_specification/releases/tag/v3.0) contain the following assets

- [vss_rel_3.0.json](https://github.com/COVESA/vehicle_signal_specification/releases/download/v3.0/vss_rel_3.0.json)
- [vss_rel_3.0.tar.gz](https://github.com/COVESA/vehicle_signal_specification/releases/download/v3.0/vss_rel_3.0.tar.gz)
- Source code (zip) 
- Source code (tar.gz) 

The first two are manually added. That tar.gz file contain output from all "supported" tools in vss-tools except binary, as binary output might be target specific). The json file (generated by vss-tools json exporter) exist in the tar.gz file as well but is also added separately to simplify showing VSS in online JSON tools.

@slawr has identified in a slack discussion that the content may be confusing and might need to be changed or better explained.

slawr:
### Input on Release Assets
I have a feeling I may have said this before, so apologies if I have and we discussed it..

Personally, I think the 'assets' of the release are presented a little strangely for people not very familiar with the project. Especially for newcomers to understand.

Looking at it from that perspective:
- The yaml is the single source of truth for the standard catalog but if I open v2.2.tar.gz I don't see the yaml, but generated serialisations with no doc or comment inside them to tell me that. Of course the newbie will not know this.
- The assets also includes vss_rel_2.2.json in the root and I might wonder is this .json the catalog, which is confusing as the doc talks about .yaml.. or maybe I'm just moving very fast and assume that the json is the source of truth. Side note: I see Daniel's explanation on slack that the json is there to drive the json viewer embedded in the doc.
- If I finally open the source code archive then I may then find a yaml file if I hunt around.

After a while you may figure out that the source code archive contains the catalog and that the v2.2.tar.gz contains generated output from vsc-tools, etc. but its not obvious early on.

How to improve that is a matter of details, but could be easily fixed in various straight forward ways. You could have a few lines in a readme.md/release.md released as an asset, which summarises what the files are using vX.Y.tag.gz filenames that don't need to be maintained. Another would be more descriptive archive names.

erikbosch:
Meeting comments:

- Shall individual files be added separately? (i.e. not in *.tar.gz)

slawr:
### Deprecation
Should the review consider to what extent we document deprecation in the rule set or standard catalog?

[...]

PR #491, opened 2022-09-16
Extend governance and changelog information

erikbosch:
This is a proposal on how to handle some recently discussed issues related to governance.

- Proposing a CHANGELOG.md file to highlight import changes, with a special focus on changes that affect backward compatibility. (Related to discussion in #473). This could also include agreed upcoming changes, e.g. if something is planned to be removed in VSS 4.0
- Introducing a CONTRIBUTING.md file, as per discussion in #488 
- Introducing the concept of "normative content". I.e. content that we consider to be part of the "VSS standard", and highlighting that overlays not (yet) are considered as normative content. (Related to #490 and #487) 

I see that something similar could be added in VSS-Tools, but possibly refering to VSS when equal.

Comments welcome!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,110 @@
+# Introduction to VSS Contribution Process
+
+The COVESA VSS project has public two GitHub repositories
+
+- [Vehicle Signal Specification (VSS)](https://github.com/COVESA/vehicle_signal_specification), containing signal specifications and documentation.
+- [VSS Tools](https://github.com/COVESA/vss-tools), containing tools for validating and transforming VSS specifications.
+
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+5. A VSS maintainer will perform the merge.
+
+## Guidelines and Recommendations
+
+This section includes general guidelines and recommendations for anyone interested in contributing to VSS.
+
+### VSS Signals shall be generic
+
+Signals added to standard VSS shall be generic, i.e. it shall be possible that other manufacturers can reuse the signal.
+Manufacturer-specific signals shall preferably eb part of private overlays, and not part of standard VSS.
```
slawr, line 57: spell check the text, e.g. eb

PR #492, opened 2022-09-16
Update vss-tools to latest version

erikbosch:
To remove stringcase dependency also in VSS


erikbosch:
Meeting decision: merge

PR #498, opened 2022-10-05
Adapt overlay documentation to implicit branches

### LINKED ISSUE #462: Support for implicit branches
In VSS, including overlays, we always request branches to be explicitly defined. Especially for overlays this does not provide much value and makes the overlays more complex than they need to. The overlay below would be easier to write if you did not had to (re)define the Vehicle-branch.

Today:

```yaml

Vehicle:
  type: branch

Vehicle.Speed:
    quality: 100
    source: "ecu0xAA"
```
Would it be an idea to support implicit branches? I.e. if a signal is defined as `Vehicle.B.C` and branch `Vehicle.B` has not yet been defined, then a branch `Vehicle.B` is implicitly created. Tooling could potentially have an option to give warnings/errors if implicit branches are used. For example - we may not want to use implicit branches in "standard VSS" (i.e. only for overlays), and warning/errors could then be useful to detect spelling errors, like if someone tries to define `Vehikle.Speed`

erikbosch:
Adapting VSS documentation to tooling support in https://github.com/COVESA/vss-tools/pull/190
Fixes #462 

Open topic: Currently this PR only describes use of implicit branches for overlays. Do we also want to mention it in "general branch documentation", i.e. state that you technically can use it everywhere, not only for overlays?

### DIFF
--- a/docs-gen/content/rule_set/overlay.md
+++ b/docs-gen/content/rule_set/overlay.md
@@ -33,8 +33,8 @@ specification file, which can overwrite or extend data in the VSS tree defined b
 the original specification.Before you start you should know:
 - **Overlay-files have to be valid specification files by themselves.**
   In practice that means, that the path to a node has to be well defined.
-  Just to specify the changes of a certain node isn't enough and needs the
-  nodes, which form the path to the node in question.
+- **You can omit parent branches if there is no need to change them**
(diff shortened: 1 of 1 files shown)


adobekan (approved):

adobekan:
11.10.22 Agreed to merge.

PR #499, opened 2022-10-18
Add PlaybackSpeed into Media.Played

wonsuk73:
- It needs to add Playback Speed into Media.Played branch because it's important feature to change playback speed on some media services as audio book, youtube video

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -49,6 +49,12 @@ Media.Played.URI:
   type: sensor
   description: User Resource associated with the media
 
+Media.Played.PlaybackSpeed:
+  datatype: float
+  type: actuator
+  allowed: [0.25, 0.5, 0.75, 1, 1.25, 1.5, 1.75, 2]
+  description: Current playback speed of media being played
+
 Media.DeclinedURI:
   datatype: string
   type: sensor


spec/Cabin/Infotainment.vspec
```diff
@@ -49,6 +49,12 @@ Media.Played.URI:
   type: sensor
   description: User Resource associated with the media
 
+Media.Played.PlaybackSpeed:
+  datatype: float
+  type: actuator
+  allowed: [0.25, 0.5, 0.75, 1, 1.25, 1.5, 1.75, 2]
```
erikbosch, line 55: Is the restrictions to only allow those particular values common, or even stated in a specification somewhere? I noticed that e.g. Android also use a float, but does not state any limitations, see eg., https://developer.android.com/reference/android/media/PlaybackParams#setSpeed(float)

wonsuk73: @erikbosch Thanks a lot for quick review! I agreed to remove the restrictions of PlaybackSpeed. I think the restriction probably can be covered by applications side. So I will remove this.

spec/Cabin/Infotainment.vspec
```diff
@@ -49,6 +49,12 @@ Media.Played.URI:
...
+  description: Current playback speed of media being played
```
erikbosch, line 55: Is it easy to understand that 1 means normal speed, <1 slower and >1 faste, or do we need to write that down in some way?

wonsuk73: @erikbosch That's good point. I will do that.

wonsuk73:
@erikbosch How about revised one? Any comment on this?

erikbosch:
> @erikbosch How about revised one? Any comment on this?

Looks OK to me, lets present it at the next meeting, and then let it be open until the following meeting for allow for additional comments

erikbosch:
Meeting decision: Valid proposal, to be merged at next meeting unless any objections received.

adobekan (commented):
Looks good to me.

spec/Cabin/Infotainment.vspec
adobekan: If needed we could find a way as well how to describe these rates in allowed tag. But as well current proposal should be good enough.

erikbosch (approved):
Looks good to me

erikbosch:
Meeting notes:

- Change name to PlaybackRate to be aligned with https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/playbackRate
- Include relationship to real speed in description, something similar to the text in the link above (" The normal playback rate is multiplied by this value to obtain the current rate, so a value of 1.0 indicates normal speed.")

wonsuk73:
@erikbosch Thanks for comments. Done the work aligned with Mozilla spec.

wonsuk73:
@erikbosch Is it okay to you?

PR #500, opened 2022-10-18
Add SmartphoneMirroring into HMI branch

spshin3:
It is required to support SmartphoneMirroring into HMI branch.  Mirroring service such as AndroidAuto, AppleCarplay, and MirrorLink are broadly used in recent HMI system in the vehicle.  The connectivity information for smart device such as android or iOS based phone or tablet is required for VSS.

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -157,7 +157,8 @@ SmartphoneProjection.Source:
   allowed: ['USB', 'BLUETOOTH', 'WIFI']
   description: Connectivity source selected for projection.
 
-SmartphoneProjection.List:
+SmartphoneProjection.SupportedMode:
   datatype: string[]
   type: sensor
+  allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -150,6 +150,7 @@ SmartphoneProjection.Active:
   type: actuator
   allowed: [ 'NONE', 'ACTIVE',  'INACTIVE' ]
   description: Projection activation info.
+  comment: NONE indicates that projection is not supported.
   
 SmartphoneProjection.Source:
   datatype: string
@@ -159,6 +160,6 @@ SmartphoneProjection.Source:
 
 SmartphoneProjection.SupportedMode:
   datatype: string[]
-  type: sensor
+  type: attribute
   allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -142,7 +142,7 @@ HMI.DayNightMode:
   description: Current display theme
   
 HMI.SmartphoneMirroring:
-	datatype: string
-	type: sensor
-	allowed: [ 'AndoridAuto', 'AppleCarplay', 'MirrorLink', 'UNKNOWN' ]
-	description: Mirroring info for connectivity of user's smartphone
+  datatype: string
+  type: sensor
+  allowed: [ 'AndoridAuto', 'AppleCarplay', 'MirrorLink', 'UNKNOWN' ]
+  description: Mirroring info for connectivity of user's smartphone
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -140,3 +140,9 @@ HMI.DayNightMode:
   type: actuator
   allowed: ['DAY', 'NIGHT']
   description: Current display theme
+  
+HMI.SmartphoneMirroring:
+	datatype: string
+	type: sensor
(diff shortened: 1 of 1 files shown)


spec/Cabin/Infotainment.vspec
```diff
@@ -140,3 +140,9 @@ HMI.DayNightMode:
   type: actuator
   allowed: ['DAY', 'NIGHT']
   description: Current display theme
+  
+HMI.SmartphoneMirroring:
+	datatype: string
```
erikbosch, line 145: Please replace tabs with two blanks, tabs cause build errors

spec/Cabin/Infotainment.vspec
```diff
@@ -140,3 +140,9 @@ HMI.DayNightMode:
   type: actuator
   allowed: ['DAY', 'NIGHT']
   description: Current display theme
+  
+HMI.SmartphoneMirroring:
+  datatype: string
+  type: sensor
+  allowed: [ 'AndoridAuto', 'AppleCarplay', 'MirrorLink', 'UNKNOWN' ]
```
SebastianSchildt, line 147: Typo: Andorid->Android

I am however not sure, this modelling is enough, as Systems might support several Mirroring Systems,. How do you express support for AppleCarPlay AND Android Auto.

Maybe better expressed by a branch and then childs like `HasMirroLink`, `HasCarPlay`, ... ?

(e.g. this https://www.dacia.co.uk/multimedia/media-nav-media-display.html supports both)

erikbosch: It needs to be clarified in the description if this signal should be used to describe supported mirroring technologies and the one currently used. As Sebastian write, if it is supposed to indicate available alternatives then we may either have multiple boolean signals in a branch, or use an array here (i.e. `string[]`) to be able to select multiple ones. If it is supposed to indicate the mirroring technology currently used then I think a signal like this would work, but then I would consider replacing `UNKNOWN` with `OTHER` (as the vehicle likely knows what technology is used, but it is not one of the standard ones). (In COVESA we have so far avoided to have special values for "unknown", "undefined" or similar, based on the assumption that if a vehicle cannot report speed, mirroring-status, supported mirroring alternatives or similar due to e.g. limitations or internal errors, then it is assumed that vehicle just ignore to report/publish that signal)

But if it is to reflect current mirroring then we might need `NONE`, `INACTIVE` or similar as well to indicate that no mirror is inactive.

I would also suggest sticking to the upper case style as stated in [documentation](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/)

i.e. something like:

`allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'UNKNOWN' ]`

spshin3:
@SebastianSchildt , @erikbosch , 
Sorry for the late response. 
I carefully reviewed the comments from you, and proposed new update.
First of all, I proposed new branch for resolving all comments for smartphone projection. 
And followings are considered to resolve the comments:
- The name of this signal changed to projection. (I agree that projection is appropriate for the name. Mirroring is used to reflect the screen of smart phone.)
- To indicate available alternatives, just use an array for the supportable list without designating specific standards or technologies.
- The activation and source signal is added.

spec/Cabin/Infotainment.vspec
```diff
@@ -140,3 +140,24 @@ HMI.DayNightMode:
...
+  allowed: [ 'NONE', 'ACTIVE',  'INACTIVE' ]
+  description: Projection activation info.
+  
+SmartphoneProjection.Source:
+  datatype: string
+  type: actuator
+  allowed: ['USB', 'BLUETOOTH', 'WIFI']
+  description: Connectivity source selected for projection.
+
+SmartphoneProjection.List:
```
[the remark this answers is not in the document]
spshin3: The naming of you sounds good for me.
And I added the allowable values with some comment.
Thanks

spec/Cabin/Infotainment.vspec
```diff
@@ -140,3 +140,25 @@ HMI.DayNightMode:
...
+  
+SmartphoneProjection:
+  type: branch
+  description: All smartphone projection actions.
+  
+SmartphoneProjection.Active:
+  datatype: string
+  type: actuator
+  allowed: [ 'NONE', 'ACTIVE',  'INACTIVE' ]
+  description: Projection activation info.
```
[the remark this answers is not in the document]
spshin3: The former one is right. My intention for NONE means that the vehicle does not support projection. In that case, now it looks more appropriate to just allow ACTIVE and INACTIVE only. Do we have any reason to keep NONE for the error? If not, I will remove the NONE. Thanks.

spec/Cabin/Infotainment.vspec
```diff
@@ -140,3 +140,25 @@ HMI.DayNightMode:
...
+  
+SmartphoneProjection.Source:
+  datatype: string
+  type: actuator
+  allowed: ['USB', 'BLUETOOTH', 'WIFI']
+  description: Connectivity source selected for projection.
+
+SmartphoneProjection.SupportedMode:
+  datatype: string[]
+  type: sensor
```
erikbosch, line 162: Should better be an attribute, will not change

spec/Cabin/Infotainment.vspec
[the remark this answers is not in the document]
spshin3: Thanks for the good comment. I refleted the comment.

erikbosch (approved):

PR #501, opened 2022-10-25
Add tire pressure unit to HMI

wonsuk73:
- It needs to add tire pressure unit to HMI branch because there are three different units related with tire pressure in usual.

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -138,7 +138,7 @@ HMI.TemperatureUnit:
 HMI.TirePressureUnit:
   datatype: string
   type: actuator
-  allowed: [‘psi’, ‘kPa’, ’bar’]
+  allowed: [‘PSI’, ‘KPA’, ’BAR’]
   description: Tire pressure unit used in the current HMI 
 
 HMI.DayNightMode:
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -135,7 +135,7 @@ HMI.TemperatureUnit:
   allowed: ['C', 'F']
   description: Temperature unit used in the current HMI
  
-HMI.TirePressureUnit
+HMI.TirePressureUnit:
   datatype: string
   type: actuator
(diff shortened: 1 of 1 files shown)


wonsuk73:
@erikbosch could you review this? Does it make sense for HMI?

spec/Cabin/Infotainment.vspec
```diff
@@ -134,6 +134,12 @@ HMI.TemperatureUnit:
   type: actuator
   allowed: ['C', 'F']
   description: Temperature unit used in the current HMI
+ 
+HMI.TirePressureUnit:
+  datatype: string
+  type: actuator
+  allowed: [‘psi’, ‘kPa’, ’bar’]
```
erikbosch, line 141: I think the signal makes sense, it is similar to other signals in this file. What needs to be discussed is the values to use; shall we use "real abbreviation" like currently proposed, or something more enum-like as in existing signals. In https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/#recommendation-on-string-values we suggest to only use upper case (and numbers), but as we here are concerned with "real abbreviations" I see that this actually could be an OK exception. But I would like to get comments from others - do we want to keep it as is or do we prefer for example:

`allowed: [‘PSI’, ‘KPA’, ’BAR’]`

wonsuk73: @erikbosch Thanks for quick review and good comment! I prefer to use upper case only like other signals. But as you said let's see the comments from others.

erikbosch: Meeting decision: Valid proposal, using upper case preferred, please update PR using upper case.
If updated and no objections received ready to be merged on next meeting

wonsuk73: @erikbosch Thanks! I have done the upper case changes.

adobekan: @erikbosch We have mixture of abbreviation and full names and i guess that should be fine.

Cheers

erikbosch (approved):
Looks good to me

PR #503, opened 2022-10-31
Add mute property into navigation branch

wonsuk73:
It is required to add date property for navigation mute.

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -89,6 +89,12 @@ Navigation.DestinationSet.Longitude:
   max: 180
   unit: degrees
   description: Longitude of destination in WGS 84 geodetic coordinates.
+  
+Navigation.Mute:
+  datatype: string
+  type: actuator
+  allowed: ['Muted', 'Alerts only', 'Unmuted']
+  description: Navigation mute state that was selected.  
 
 HMI:
   type: branch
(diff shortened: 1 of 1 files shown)


spec/Cabin/Infotainment.vspec
```diff
@@ -89,6 +89,12 @@ Navigation.DestinationSet.Longitude:
   max: 180
   unit: degrees
   description: Longitude of destination in WGS 84 geodetic coordinates.
+  
+Navigation.Mute:
+  datatype: string
+  type: actuator
+  allowed: ['Muted', 'Alerts only', 'Unmuted']
```
erikbosch, line 96: I would prefer if the [style guide](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/#recommendation-on-string-values) is followed, i.e. use upper case and underscore. (Alternatively we can discuss the style guide at next meeting to see if we still consider the style guide as relevant)

i.e. something like:

`allowed: ['MUTED', 'ALERT_ONLY', 'UNMUTED']`

erikbosch:
A minor comment - according to https://www.covesa.global/contribute there should be a sign-off on each commit (e.g. `git commit -s`). We have not benn following that rule strictly in the past, but for new commits we shall try to follow it.

spec/Cabin/Infotainment.vspec
wonsuk73: @erikbosch Thanks for quick comment! I am sorry that It is the same comment with previous my PR :( I have applied your comment.

wonsuk73:
@erikbosch Thanks for this comment. I will follow up the guideline you mentioned.

erikbosch (approved):
Looks good to me. I will be on vacation next week, I suggest that @adobekan or @SebastianSchildt  initiates the discussion on the meeting next week if this PR can be merged

adobekan:
looks good

Issue #505, opened 2022-11-01
Definition of "current trip"


erikbosch:
We have some signals related to "current trip" in VSS, but they do not really give any definition on what is meant with "current trip". Do we consider "current trip" to be something that is manually controlled by the driver (or a vehicle app, like automatic reset when a new rental period begins), i.e. there is no definition that say that starting the engine means that a new trip automatically begins?

A side topic - how does OEMs typically calculate average speed for "current trip". Do they consider all time when "Vehicle.LowVoltageSystemState" is on, i.e. considering time when engine is idling? Or do they exclude time when transmission is in neutral? 

```

ConsumptionSinceStart:
  datatype: float
  type: sensor
  unit: l
  description: Fuel amount in liters consumed since start of current trip.

TimeSinceStart:
  datatype: uint32
  type: sensor
  unit: s
  description: Time in seconds elapsed since start of current trip.


TripMeterReading:
  datatype: float
  type: sensor
  unit: km
  description: Current trip meter reading.


AverageSpeed:
  datatype: float
  type: sensor
  unit: km/h
  description: Average speed for the current trip.

```

adobekan:
@jdacoello will have a look for BMW details.

erikbosch:
Meeting notes:

Daniel Alvarez - current trip is from "start of car" to "stop of car". (from pressing "start button" to pressing "stop button"). It shall better be described how calculation was done for AverageSpeed. E.g. only considering when vehicle is moving?
Erik - But is Trip meter referring to the trip meter that can be manually reset.  E.g. current value of trip meter, or value of current trip.
Erik - For AverageSpeed, can we in VSS define how average speed is calculated, or say that it is OEM specific
Erik to create PR for further discussion

erikbosch:
Can be closed as PR merged

PR #509, opened 2022-11-18
Refactor Current Trip signals

erikbosch:
Related to #505 

Comments are welcome as it is still a bit open what is actually needed by OEMs

### DIFF
--- a/spec/Powertrain/FuelSystem.vspec
+++ b/spec/Powertrain/FuelSystem.vspec
@@ -81,7 +81,7 @@ ConsumptionSinceStart:
            The signal may however keep the value of the last trip until a new trip is started.
 
 TimeSinceStart:
-  deprecation: V3.1 replaced by Vehicle.StartTime
+  deprecation: V3.1 replaced by Vehicle.StartTime and Vehicle.TripDuration
   datatype: uint32
   type: sensor
   unit: s
--- a/spec/Powertrain/FuelSystem.vspec
+++ b/spec/Powertrain/FuelSystem.vspec
@@ -76,8 +76,12 @@ ConsumptionSinceStart:
   type: sensor
   unit: l
   description: Fuel amount in liters consumed since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
+           The signal may however keep the value of the last trip until a new trip is started.
 
 TimeSinceStart:
+  deprecation: V3.1 replaced by Vehicle.StartTime
   datatype: uint32
   type: sensor
   unit: s
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -207,20 +207,23 @@ TraveledDistanceSinceStart:
 
 StartTime:
   datatype: string
-  type: sensor
+  type: attribute
+  default: '0000-01-01T00:00Z'
   description: Start time of current or latest trip, formatted according to ISO 8601 with UTC time zone.
   comment: This signal is supposed to be set whenever a new trip starts.
            A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
            A trip is considered to end when engine is no longer enabled.
+           The default value indicates that the vehicle never has been started, or that latest start time is unknown.
 
-EndTime:
-  datatype: string
+TripDuration:
+  datatype: float
   type: sensor
-  description: End time of latest trip, formatted according to ISO 8601 with UTC time zone.
-  comment: This signal is supposed to be set whenever a trip ends.
+  unit: s
+  description: Duration of latest trip.
+  comment: This signal is not assumed to be continuously updated, but instead set to 0 when a trip starts
+           and set to the the actual duration of the trip when a trip ends.
            A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
            A trip is considered to end when engine is no longer enabled.
-           If EndTime is later than StartTime that means no trip is ongoing and duration of latest trip can be calculated by EndTime - StartTime.
 
 TripMeterReading:
   datatype: float
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -184,16 +184,50 @@ Speed:
   description: Vehicle speed.
 
 TravelledDistance:
+  deprecation: V3.1 moved to Vehicle.TraveledDistance
   datatype: float
   type: sensor
   unit: km
-  description: Odometer reading, total distance travelled during the lifetime of the vehicle.
+  description: Odometer reading, total distance traveled during the lifetime of the vehicle.
+
+TraveledDistance:
+  datatype: float
+  type: sensor
+  unit: km
+  description: Odometer reading, total distance traveled during the lifetime of the vehicle.
+
+TraveledDistanceSinceStart:
+  datatype: float
+  type: sensor
+  unit: km
+  description: Distance traveled since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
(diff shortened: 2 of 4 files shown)


spec/Vehicle/Vehicle.vspec
```diff
@@ -184,16 +184,43 @@ Speed:
   description: Vehicle speed.
 
 TravelledDistance:
+  deprecation: V3.1 moved to Vehicle.TraveledDistance
```
erikbosch, line 187: Today we are mixing British and American English. While poking around in this area, do we think it is a good idea to rename TravelledDistance to TraveledDistance (using our deprecated-methodology, i.e. keeping old signal until v4.0)

spec/Vehicle/Vehicle.vspec
```diff
@@ -184,16 +184,43 @@ Speed:
...
+TraveledDistanceSinceStart:
+  datatype: float
+  type: sensor
+  unit: km
+  description: Distance traveled since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
+           The signal may however keep the value of the last trip until a new trip is started.
+
+TimeSinceStart:
```
erikbosch, line 208: Proposing to move this signal from FuelSystem to Vehicle as it might be relevant for any vehicle

spec/Vehicle/Vehicle.vspec
```diff
@@ -184,16 +184,43 @@ Speed:
...
+  deprecation: V3.1
+  datatype: uint32
+  type: sensor
+  unit: s
+  description: Time in seconds elapsed since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
+           The signal may however keep the value of the last trip until a new trip is started.
+
+TripMeterReading:
```
erikbosch, line 218: For traveled distance I see it relevant to have 3 values

- TraveledDistance - the value for the life time of the vehicle
- TraveledDistanceSinceStart - the value since start of current trip
- TripMeterReading - the value since manual reset (by driver)

Adapting description accordingly.

spec/Vehicle/Vehicle.vspec
```diff
@@ -214,7 +241,10 @@ AverageSpeed:
   type: sensor
   unit: km/h
   description: Average speed for the current trip.
-
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
```
erikbosch, line 244: Is it sufficient to have a single `AverageSpeed` - is so do we want to keep old name, or rename to `AverageSpeedSinceStart` to be better aligned with other signals referring to current trip.

Or do we need additional signals for average speed - e.g. by adding one related to average speed over lifetime or average speed since trip meter reset?

spec/Powertrain/FuelSystem.vspec
```diff
@@ -76,8 +76,12 @@ ConsumptionSinceStart:
   type: sensor
   unit: l
   description: Fuel amount in liters consumed since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
```
erikbosch, line 79: Is it sufficient to have consumption since start of current trip, or do we also need a signal for "consumption since trip meter reset". Similar for the signal `AverageConsumption` above. Today we state that period is a variation point, shall we better specify that it is for current trip or split it to two signals, one for "average consumption for current trip" and one for "average consumption since trip meter reset"

spec/Vehicle/Vehicle.vspec
SebastianSchildt: +1

SebastianSchildt:
In case we want to agree that - if in doubt - American English spelling "wins", should we write it down somewhere in the docs?

erikbosch:
> In case we want to agree that - if in doubt - American English spelling "wins", should we write it down somewhere in the docs?

Briefly mentioned in https://github.com/COVESA/vehicle_signal_specification/blob/master/CONTRIBUTING.md#use-existing-style

spec/Vehicle/Vehicle.vspec
erikbosch: Meeting notes:
Daniel: Would it be better to have starttime than time since start?
Ulf: Dont we not need a signal for "now" as well
Daniel: Assume apps can always get current time
Erik: Have no problems having start time instead

erikbosch:
Meeting notes:
- Erik to update with timestamp instead of "time since start"
- Others should review after update so it can be discussed/approved at next meeting

spec/Vehicle/Vehicle.vspec
```diff
@@ -184,16 +184,41 @@ Speed:
...
+TraveledDistanceSinceStart:
+  datatype: float
+  type: sensor
+  unit: km
+  description: Distance traveled since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
+           The signal may however keep the value of the last trip until a new trip is started.
+
+StartTime:
```
erikbosch, line 208: Changed from TimeSinceStart to StartTime based on discussion in last meeting.
Using string in UTC-format, as that is the format used in some other places in the tree to represent time - we currently do not use unix time stamps (i.e. seconds since 1970 represented as an int value anywhere)

(That is a side topic that comes up now and then, do we in VSS syntax need something that from a formal perspective tells that a signal contains a time stamp, either by using a predefined unit, or a datatype alias (for example "datatype:time" but specifying that it means string + iso 8601 syntax)

erikbosch:
Decision postponed

erikbosch:
A thought - if we change to `StartTime`, would it then make sense to add `EndTime`? I.e. so that duration of last trip is (indirectly) persisted in tree, at least until a new `StartTime` has been set.

spec/Vehicle/Vehicle.vspec
```diff
@@ -184,16 +184,50 @@ Speed:
...
+  datatype: string
+  type: sensor
+  description: Start time of current or latest trip, formatted according to ISO 8601 with UTC time zone.
+  comment: This signal is supposed to be set whenever a new trip starts.
+           A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
+
+EndTime:
+  datatype: string
+  type: sensor
```
erikbosch, line 218: Meeting notes:

- Daniel: Is rather an attribute, change only once per ignition cycle. 
- Sebastian: but could be not that intuitative to subscribe to an attribute to be informed when trip finishes.
- Daniel: Can be better to have both start time (attribute) and duration (sensor). 
- Decision: Erik to adapt to PR to use "both start time (attribute) and duration (sensor)". After update, others shall review. Merge or continued discussion on next meeting

spec/Vehicle/Vehicle.vspec
```diff
@@ -184,16 +184,53 @@ Speed:
...
+  unit: km
+  description: Distance traveled since start of current trip.
+  comment: A new trip is considered to start when engine gets enabled (e.g. LowVoltageSystemState in ON or START mode).
+           A trip is considered to end when engine is no longer enabled.
+           The signal may however keep the value of the last trip until a new trip is started.
+
+StartTime:
+  datatype: string
+  type: attribute
+  default: '0000-01-01T00:00Z'
```
erikbosch, line 211: At last meeting we discussed a bit on what an attribute is, and if this signal makes sense to be an attribute as it is only set once per ignition cycle. I changed the PR to use attribute as type, but then I found that we say in documentation that an attribute must have a default value. Although not explicitly said I assume it originates from an idea that an attribute always will have a value, i.e. if you do a Read you will always get a value (if the signal is defined), compared to sensors where you might get an answer that no value is available.

We can discuss if we want to keep this restriction for attributes - for now selecting a year 0 as default value, if used indicating that vehicle has never been started or that latest start time is unknown

erikbosch:
Meeting decision: merge

PR #514, opened 2022-11-29
Add light switches

erikbosch:
VSS as of today has individual signals controlling/reporting if lights are on, but it does not have any signals representing the controls typically available to the driver. This PR intends to address that.

Some comments in addition to what is in the file

**Lights.LightSwitch**

- The "main" light switch that almost all vehicles have. 
- Actual behavior typically depends on ignition state and vehicle configuration. I.e. 'OFF' does not necessarily means that all lights are off, in countries where driving with lights off is illegal some vehicles will light tail light and low beam in 'OFF' mode, or some other configuration.
- Having an explicit DRL state seems to be rare, but exists. In many vehicles the only way to use DRL is to select AUTO and drive during daytime.

**Lights.IsHighBeamSwitchOn**

- The switch to enable high beam, on many vehicles controlled by the same handle as the blinkers
- I assume we do not need a "single flash" state, as that is handled by the handle mechanics rather than a special state
- I assume we do not need a separate signal or value for "automatic high beam" - on the cars I have driven so far the same switch is used and whether you get manual or automatic high beam depends on Lights.LightSwitch

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -170,6 +170,26 @@ Lights:
   type: branch
   description: Exterior lights.
 
+Lights.LightSwitch:
+  datatype: string
+  type: actuator
+  allowed: ['OFF', 'POSITION', 'DAYTIME_RUNNING_LIGHTS', 'AUTO', 'BEAM']
+  description: Status of the vehicle main light switch.
+  comment: A vehicle typically does not support all alternatives.
+           Which lights that actually are lit may vary according to vehicle configuration and local legislation.
+           OFF is typically indicated by 0.
+           POSITION is typically indicated by ISO 7000 symbol 0456.
+           DAYTIME_RUNNING_LIGHTS (DRL) can be indicated by ISO 7000 symbol 2611.
+           AUTO indicates that vehicle automatically selects suitable lights.
+           BEAM is typically indicated by ISO 7000 symbol 0083.
+
+Lights.IsHighBeamSwitchOn:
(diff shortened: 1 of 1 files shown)


spec/Body/Body.vspec
```diff
@@ -170,6 +170,26 @@ Lights:
...
+Lights.LightSwitch:
+  datatype: string
+  type: actuator
+  allowed: ['OFF', 'POSITION', 'DRL', 'AUTO', 'BEAM']
+  description: Status of the vehicle main light switch.
+  comment: A vehicle typically does not support all alternatives.
+           Which lights that actually are lit may vary according to vehicle configuration and local legislation.
+           OFF is typically indicated by 0.
+           POSITION is typically indicated by ISO 7000 symbol 0456.
+           DRL (if present) indicates Daytime Running Lights.
```
SebastianSchildt, line 182: is "DRL" a common abbreviation? Otherwise VSS-wise having something more expressive like DAYLIGHT_RUNNING would not hurt us

erikbosch: Quite common (see e.g. https://en.wikipedia.org/wiki/Daytime_running_lamp), but for example ISO 7000 use the term "Daytime running lights" so let us use that.

erikbosch:
Presented at VSS meeting.
Please review until next meeting

SebastianSchildt (approved):
Looking good to me

erikbosch:
Meeting decision: merge

PR #520, opened 2022-12-20
Added Display Brightness Level for Vehicle Power Optimization

UmangSharmaMobis:
added power optimization to existing structure  as per the feedback of the meeting.

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -211,3 +211,20 @@ SmartphoneProjection.SupportedMode:
   type: attribute
   allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
+
+PowerOptimize:
+  type: branch
+  description: All PowerOptimize actions.
+
+PowerOptimize.DisplayBrightnessLevel:
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Display BrightnessLevel in the case of PowerOptimize
+               0 = Lowest brightness possible.
+               100 = Maximum Brightness possible.
+
+
+


UmangSharmaMobis (commented):
as per the feedback added signal to the existing structure, may be in future power optimization can have a separate branch.

spec/Cabin/Infotainment.vspec
```diff
@@ -211,3 +211,20 @@ SmartphoneProjection.SupportedMode:
   type: attribute
   allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
+
+PowerOptimize:
+  type: branch
+  description: All PowerOptimize actions.
+
+PowerOptimize.DisplayBrightnessLevel:
```
erikbosch, line 219: I see two points that needs to be discussed

**Intended logic**

If this signal is added we will have two brightness signals, this one and `HMI.Brightness`. How is an HMI/Display supposed to behave? Shall it use `HMI.Brightness` normally but `PowerOptimize.DisplayBrightnessLevel` if PowerOptimize mode is set? If so, then I think it would make sense to add the "global" PowerOptimize signal first so it is clear when this signal shall be used.

**Naming**

We already have a signal `HMI.Brightness`, this signal use the term `DisplayBrightnessLevel`. Assuming that they refer to the same display I think we should align the names, for example by making sure they both have the same leaf-name, e.g. either  `HMI.Brightness` and  `PowerOptimize.Brightness`, or `HMI.DisplayBrightnessLevel` and  `PowerOptimize.DisplayBrightnessLevel` or some other aligned naming alternative,

SebastianSchildt: Not followed all the discussion, but are we expecting Poweroptimize sub-branches in other areas as well? Would that not be a better fit for an overlay?

spec/Cabin/Infotainment.vspec
```diff
@@ -211,3 +211,20 @@ SmartphoneProjection.SupportedMode:
...
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Display BrightnessLevel in the case of PowerOptimize
```
erikbosch, line 225: Discussed in meeting. Intended usage is rather:

description: "Brightness level proposed by power optimization system. It is up to HMI to decide if it want to follow advice."

erikbosch:
This PR and the Power Optimize topic in general was discussed at VSS meeting 2022-12-20, see [notes](https://wiki.covesa.global/display/WIK4/VSS+Meeting+Topics+and+Meeting+Notes). A simple architecture description is requested that could help to understand how the "Power optimize" signals are supposed to be used. For example - if we add a `PowerOptimize.DisplayBrightnessLevel` signal, who is supposed to set it and who is supposed to read it. Shall the HMI/Display always use the brightness specified in `HMI.Brightness` or shall it in some cases instead use the brightness in `PowerOptimize.DisplayBrightnessLevel`? Discussion in VSS meetings to be continued when there is a description on how the signals are to be used.

(I have by the way created a separate ticket in #522 on how to document intended usage of VSS signals, where power optimization could be a typical example of a feature that needs to be documented)

erikbosch:
Meeting notes: Mobis to prepare architecture documents for next week

UmangSharmaMobis:
Please find architecture below:
Capture_1

erikbosch:
Trying to understand the architecture and how it is linked to this PR, base don the image above and what was communicated at the last VSS meeting. My understanding is somewhat like this; please comment if it is correctly understood

- The "EV Power Optimization Manager" (the yellow box) will select a power optimization scheme based on e.g. battery charging level
- A subsystem component (the green boxes) will react on the power optimization scheme selected by the Manager and do relevant actions in that sub system.

I.e the "EV Power Optimization Manager" will not directly request a specific HMI brightness, but rather inform "EV Power Optimization GUI" component that it e.g. shall use scheme number X or something similar? I.e. it is the "EV Power Optimization GUI" that will use both `PowerOptimize.DisplayBrightnessLevel` (in this PR) and the existing `HMI.Brightness`. I.e. use `PowerOptimize.DisplayBrightnessLevel` as some form of configuration parameter and when needed use it to change value for `HMI.Brightness`?

UmangSharmaMobis:
it fit to be fine and can be replaced by #528.

Issue #396, opened 2022-01-11
How to document and where to place unit


danielwilms:
* how to keep docs and config in sync
* where to keep config and make it user friendly

slawr:
Related to this is to what extent the changes introduced with config.yaml have any rule set implications to document? One example would be the 'domain' key for each unit. Is this a defined list of allowed values or free-form?

slawr:
**Current sync state**
Table of unit types from the vss-tools config.yaml and the table in the VSS doc

|config.yaml|VSS doc|
|---|---|
|A|A|
|  Nm|Hz|
|  PS|N|
|  Pa|Nm|
|  UNIX Timestamp|Pa|
|  V|V|
|  celsius|W|
|  cm|celsius|
|  cm^3|cm/s^2|
|  day|degree|
|  degrees/s|degrees/s|
|  degrees|g|
|  g/km|g/s|
|  g/s|h|
|  h|inch|
|  inch|kPa|
|  kPa|kW|
|  kW|kWh|
|  kWh/100km|kg|
|  kWh|km|
|  kg|km/h|
|  km/h|l|
|  km|l/100km|
|  l/100km|l/h|
|  l/h|lat|
|  l|lon|
|  lbs|m|
|  m/s^2|m/s|
|  m|m/s^2|
|  min|mbar|
|  mm|min|
|  months|ml|
|  mpg|ml/100km|
|  percent|mm|
|  ratio|ms|
|  rpm|percent|
|  s|ratio|
|  stars|rpm|
|  weeks|s|
|  years||

danielwilms:
Decision to move config.yaml for specification of supported unit in to the vss repository being a required input for vss-tools.

Issue #346, opened 2021-10-18
Units definitions both SI and common usage ?


petervolvowinz:
So, I am looking at signal sets and their sometimes specified unit. I was wondering a bit about how we define units in VSS - http://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/data_unit_types/-  and  for example km/h and not mph. Both are commonly used and not SI-units , so we sometimes use SI and sometimes not. Is there a rule based decision to include non SI units ? Such as mph is commonly used as unit for this particular senor...or ? I am trying to build a table of units and their relations.

oliveirarleo:
Discussions on vss-tools are running about having custom units (any unit you want to have basically). Currently, there is a [PR](https://github.com/GENIVI/vss-tools/pull/104) there, but there is the possibility of putting a file here with the units you want to support.

erikbosch:
Discussed at meeting, no we do not have any strict policy as of today. we prefer SI units unless there is a good reason to use other units due to industry standard (e.g. km/h for speed, inch for rim size)

@petervolvowinz - are there any particular unit you ahve problems with or would like to remove/add?

danielwilms:
@petervolvowinz: now there's as well a description in the [vss-tools FAQ](https://github.com/COVESA/vss-tools/blob/master/FAQ.md) with information on how to extend the list of units. Please have a look and check if there's something we can improve here, information you're missing or use cases, which aren't covered by now.

PR #521, opened 2022-12-21
Use explicit VSS units

erikbosch:
This PR copies the current unit file to VSS and use it in the Makefile.
This is part of the decided separation of tooling and actual VSS catalog.

Related to #346 and #396


erikbosch:
Meeting notes:

- Please review until next meeting

erikbosch:
Meeting decision: merge

PR #535, opened 2023-02-10
Add description for optional extras, or optional equipment of vehicle

adobekan:
Adding attribute for vehicle optional equipment

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -151,10 +151,10 @@ VehicleIdentification.KnownVehicleDamages:
   description: A textual description of known damages, both repaired and unrepaired.
 
 VehicleIdentification.OptionalExtras:
-  datatype:string[]
+  datatype: string[]
   type: attribute
   description: Optional extras refers to all car equipment options that are not installed as standard by the manufacturer.
-
+  comment: Allowed values are not standardized, each OEM can specify detail descriptions of array elements.
 #
 # Vehicle signals
 #
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -150,6 +150,11 @@ VehicleIdentification.KnownVehicleDamages:
   type: attribute
   description: A textual description of known damages, both repaired and unrepaired.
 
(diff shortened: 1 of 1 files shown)


spec/Vehicle/Vehicle.vspec
```diff
@@ -150,6 +150,11 @@ VehicleIdentification.KnownVehicleDamages:
   type: attribute
   description: A textual description of known damages, both repaired and unrepaired.
 
+VehicleIdentification.OptionalExtras:
+  datatype:string[]
```
erikbosch, line 154: I think the syntax require a blank after the colon, likely reason why build fails

spec/Vehicle/Vehicle.vspec
```diff
@@ -150,6 +150,11 @@ VehicleIdentification.KnownVehicleDamages:
...
+  type: attribute
+  description: Optional extras refers to all car equipment options that are not installed as standard by the manufacturer.
```
erikbosch, line 156: Do we need to add a comment that allowed values are not standardized, it is up to OEM to define what to use.

adobekan: Done.

SebastianSchildt (approved):
Fine for me, just a note, should this be called "Options" instead, because while probably the intnetion is to add "ZOMGBoomSoundSystem", "Optigrill Seatheating" ) , I would assume there are some "options/extras" that are mandatory/standard in some markets and optional/paid in others (i.e. maybe lane assist).
VSS side, as this is a configuration/technical info that may be helpful, I would assume the intention is to list all options available in a specific vehicle type and only the ones a specific customers in a specific market happened to need to pay extra for??

adobekan:
@SebastianSchildt  Proposal is OptionalExtras and description is saying "options that are not installed as standard ". What is standard I do not see described in this field. DefaultExtras and OptionalExtras you could have. Calling it Options then it will mix everything, plus name I would discuss. :)

erikbosch:
Meeting notes: Please review/comment. Possible merge decision Feb 21st.

erikbosch:
Meeing decision: merge and include in VSS 3.1

PR #536, opened 2023-02-10
Adding new sensors to powertrain battery and fuel system.

adobekan:
Adding few more sensor descriptions for battery, fuel system and charging.

### DIFF
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -189,6 +189,7 @@ Charging.ChargeLimit:
   type: actuator
   min: 0
   max: 100
+  default: 100
   unit: percent
   description: Target charge limit (state of charge) for battery.
 
@@ -382,13 +383,6 @@ Charging.TimeToComplete:
   comment: Shall consider time set by Charging.Timer.Time.
            E.g. if charging shall start in 3 hours and 2 hours of charging is needed,
            then Charging.TimeToComplete shall report 5 hours.
-
-Charging.Displayed:
-  type: sensor
-  datatype: uint32
-  unit: s
-  description: The time displayed to the customer which is needed for the current charging process to reach complete NetCapacity of the battery.
-  comment: Calculated time for reaching 100% of NetCapacity.
 
 Charging.Timer:
   type: branch
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -104,11 +104,12 @@ StateOfCharge.Displayed:
   max: 100.00
   description: State of charge displayed to the customer.
 
-StateOfCharge.CurrentNetCharge:
+StateOfCharge.CurrentEnergy:
   type: sensor
   datatype: float
   unit: kWh
-  description: Current net state of charge of the high voltage battery expressed in kWh.
+  description: Physical state of charge of high voltage battery expressed in kWh.
+  comment: Current energy could be calculated as .StateOfCharge.Current * .NetCapacity.
 
 NominalVoltage:
   datatype: uint16
@@ -386,7 +387,8 @@ Charging.Displayed:
   type: sensor
   datatype: uint32
   unit: s
-  description: The time displayed to the customer which is needed for the current charging process to reach complete charge.
+  description: The time displayed to the customer which is needed for the current charging process to reach complete NetCapacity of the battery.
+  comment: Calculated time for reaching 100% of NetCapacity.
 
 Charging.Timer:
   type: branch
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -104,6 +104,12 @@ StateOfCharge.Displayed:
   max: 100.00
   description: State of charge displayed to the customer.
 
+StateOfCharge.CurrentNetCharge;
+  type: sensor
+  datatype: float
+  unit: kWh
+  description: Current net state of charge of the high voltage battery expressed in kWh.
+
 NominalVoltage:
   datatype: uint16
   type: attribute
@@ -375,6 +381,12 @@ Charging.TimeToComplete:
   comment: Shall consider time set by Charging.Timer.Time.
            E.g. if charging shall start in 3 hours and 2 hours of charging is needed,
            then Charging.TimeToComplete shall report 5 hours.
+
+Charging.Displayed:
+  type: uint32
+  datatype: float
+  unit: s
+  description: The time displayed to the customer which is needed for the current charging process to reach complete charge.
 
 Charging.Timer:
   type: branch
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -104,7 +104,7 @@ StateOfCharge.Displayed:
   max: 100.00
   description: State of charge displayed to the customer.
 
-StateOfCharge.CurrentNetCharge;
+StateOfCharge.CurrentNetCharge:
   type: sensor
   datatype: float
   unit: kWh
@@ -383,8 +383,8 @@ Charging.TimeToComplete:
            then Charging.TimeToComplete shall report 5 hours.
 
 Charging.Displayed:
-  type: uint32
-  datatype: float
+  type: sensor
+  datatype: uint32
   unit: s
   description: The time displayed to the customer which is needed for the current charging process to reach complete charge.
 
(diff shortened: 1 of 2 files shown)


spec/Powertrain/Battery.vspec
```diff
@@ -104,6 +104,12 @@ StateOfCharge.Displayed:
   max: 100.00
   description: State of charge displayed to the customer.
 
+StateOfCharge.CurrentNetCharge;
```
erikbosch, line 107: Ends with semicolon

spec/Powertrain/Battery.vspec
```diff
@@ -376,6 +382,12 @@ Charging.TimeToComplete:
            E.g. if charging shall start in 3 hours and 2 hours of charging is needed,
            then Charging.TimeToComplete shall report 5 hours.
 
+Charging.Displayed:
+  type: uint32
+  datatype: float
+  unit: s
+  description: The time displayed to the customer which is needed for the current charging process to reach complete charge.
```
erikbosch, line 389: How will this one differ from Charging.TimeToComplete?

Is it remaining time or time stamp it is supposed to continue?

spec/Powertrain/Battery.vspec
```diff
@@ -376,6 +382,12 @@ Charging.TimeToComplete:
            E.g. if charging shall start in 3 hours and 2 hours of charging is needed,
            then Charging.TimeToComplete shall report 5 hours.
 
+Charging.Displayed:
+  type: uint32
```
erikbosch, line 386: Should be sensor

spec/Powertrain/FuelSystem.vspec
```diff
@@ -40,6 +40,12 @@ TankCapacity:
   unit: l
   description: Capacity of the fuel tank in liters.
 
+TankNetCapacity:
+  datatype: float
+  type: sensor
+  unit: l
+  description: Net capacity of fuel tank in liters.
```
erikbosch, line 47: What is the difference between "capacity" and "net capacity"? How come that it varies? I think a comment explaining difference would be good.

spec/Powertrain/Battery.vspec
erikbosch, line 107: ends with semicolon

spec/Powertrain/Battery.vspec
erikbosch, line 386: type should be sensor

erikbosch (changes requested):

spec/Powertrain/Battery.vspec
adobekan: TimeToComplete is related to ChargeLimit. And the one I have proposed is always related to full charge of current net capacity. I will add comment here.

spec/Powertrain/FuelSystem.vspec
adobekan: I think i did here mistake with naming.

spec/Powertrain/Battery.vspec
```diff
@@ -376,6 +383,13 @@ Charging.TimeToComplete:
            E.g. if charging shall start in 3 hours and 2 hours of charging is needed,
            then Charging.TimeToComplete shall report 5 hours.
 
+Charging.Displayed:
```
SebastianSchildt, line 386: This seesm something between a "technical" value (i.e. when wen the concept is, that the raw data is in VSS and an UI may want to modify how to represent it to the user), and a pure HMI info (e.g. when we modfelled somehwere else whether speed is displayed in km/h or mph)

For example, I guess charging to 100% is sort of a corner case, where most users probably charge to 90% or whatever default a vehicle has. And as a user (if that is what "displayed" means, I am interested when my charging process finishes (i.e 50%; 19%, 100%, "until I am ready to go again", whatever I have chosen in UI).

Maybe I misunderstood what is meant with this datapoint, but then I would still argue either the name or the description is wrong

spec/Powertrain/FuelSystem.vspec
```diff
@@ -40,6 +40,12 @@ TankCapacity:
   unit: l
   description: Capacity of the fuel tank in liters.
 
+FuelStatus:
```
SebastianSchildt, line 43: What is the difference to level? I would suggest when we introduce this, we remove level. Otherwise we store the same thing twice

adobekan: In the description you can notice that main difference is in unit. Question is more how and where these values would be mapped in the actual system, you could have % on much higher level like HMI and liters could be used elsewhere. @SebastianSchildt did you use level before?

spec/Powertrain/Battery.vspec
erikbosch: Would not `Charging.TimetoFull` or something similar fit better `Charging.Displayed` does not say much what the signal contain

spec/Powertrain/FuelSystem.vspec
erikbosch: Alternatively we could consider calling it `AbsoluteLevel` (and possibly rename the other to `RelativeLevel`)

adobekan: @erikbosch that could work. FuelSystem.AbsoluteLevel + Relative

spec/Powertrain/Battery.vspec
adobekan: I have used same logic as we have for SoC. Current vs Displayed

erikbosch:
Meeting notes: Please review/comment. Possible merge decision Feb 21st.

spec/Powertrain/Battery.vspec
erikbosch: One can argue that the term `StateOfCharge` is a well known characteristic of battery, i.e. if you see a signal name like `Vehicle.Powertrain.TractionBattery.StateOfCharge.Displayed` then quite many people would be able to guess what unit/range the signal has. I am not as sure that `Vehicle.Powertrain.TractionBattery.Charging.Displayed` is as easy to understand.

But we maybe should take a step back and look at the scenarios. We have `Charging.ChargeLimit` which indicates max-limit, like if we reach it charging shall stop automatically, and  `Charging.TimeToComplete` is the time to that limit. In the meeting yesterday we rather talked about charging that shall continue after reaching the target, and two times/durations are needed, one for remaining time until target and one for remaining time until full. Now I assume that it is the "real" value we want to show - the reason for having two signals for StateOfCharge (`StateOfCharge.Displayed` and `StateOfCharge.Current`)  is that by some reasons OEMs does not want to show the real value. 

If we assume that we need to support two scenarios:

- The user/driver can specify a minimum limit. When charging has reached that limit the user will be informed in his app, but charging will continue
- The user/driver can specify a maximum limit (or it is decided by vehicle, possibly fixed at 100%).  When charging has reached that limit the user will be informed in his app, and charging will stop.

Then we maybe need four signals (names can be discussed):

- `Charging.ChargeLimit` to be replaced with  `Charging.MinimumChargeLimit` and  `Charging.MaximumChargeLimit`
- `Charging.TimeToComplete` to be replaced by `Charging.TimeToMinimumChargeLimit` and `Charging.TimeToMaximumChargeLimit`

But which time that will actually will be shown in the vehicle or in the app may vary, I assume. If you just want enough charge to be able to drive home then I assume `Charging.TimeToMinimumChargeLimit` is what you are most interested in.

adobekan: I think i will change setup and use existing time information with modification on ChargeLimit with default value to 100. Then it should cover use-cases. 
@erikbosch but also your proposal is good with adding more signals. Maybe I can add those as well, I would not add TimeToMinimum as an actuator. It is hard to say what is the minimum for driver. Or you have some scenario in mind?

ChargeLimit is what customer defines as an input, actuator. e.g. 70%
TimeToMinimumChargeLimit I see as a value that could be proposed to customer, sensor.

erikbosch:
Meeting notes: Adnan to update PR today, all review, merge decision next meeting

erikbosch (approved):

spec/Powertrain/Battery.vspec
ppb2020: Regarding:
> The user/driver can specify a minimum limit. When charging has reached that limit the user will be informed in his app, but charging will continue

Would it make sense for the app to inform the user based on its monitoring of the charge level, as opposed to pushing the decision and monitoring into the vehicle? What if there are multiple applications each with their own minimum and maximum charge limits for the same vehicle (two partners sharing a vehicle).

If a limit is set using `Charging.MinimumChargeLimit`, what is the signal that is triggered by reaching this limit? As well, from what direction is this limit monitored? Charging (as implied by the name) or discharging (where the user may want to be informed when the charge level reaches this limit while discharging)?

erikbosch:
Meeting notes: merge

PR #538, opened 2023-02-13
Update instance description of Seat, Door, HVAC, Light and Mirror

adobekan:
Change instance description of Left and Right into Driver and Passenger.

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -242,7 +242,7 @@ Lights.DirectionIndicator:
 ##
 Mirrors:
   type: branch
-  instances: ["Driver", "Passenger"]
+  instances: ["DriverSide", "PassengerSide"]
   description: All mirrors.
 # Include mirrors specification and attach it to the Mirrors branch
 #include ExteriorMirrors.vspec Mirrors
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -242,7 +242,7 @@ Lights.DirectionIndicator:
 ##
 Mirrors:
   type: branch
-  instances: ["Left", "Right"]
+  instances: ["Driver", "Passenger"]
   description: All mirrors.
 # Include mirrors specification and attach it to the Mirrors branch
 #include ExteriorMirrors.vspec Mirrors
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -110,7 +110,7 @@ Door:
   type: branch
   instances:
     - Row[1,2]
-    - ["Driver","Passenger"]
+    - ["DriverSide","PassengerSide"]
   description: All doors, including windows and switches.
 #include SingleDoor.vspec Door
 
@@ -138,7 +138,7 @@ Seat:
   type: branch
   instances:
     - Row[1,2]
-    - ["Driver","Middle","Passenger"]
+    - ["DriverSide","Middle","PassengerSide"]
   description: All seats.
 #include SingleSeat.vspec Seat
 
@@ -148,7 +148,7 @@ Seat:
 DriverPosition:
   datatype: string
   type: attribute
-  allowed: ['LEFT', 'RIGHT']
+  allowed: ['LEFT', 'MIDDLE', 'RIGHT']
   default: 'LEFT'
   description: The position of the driver seat in row 1.
   comment: Default value is position 'LEFT', i.e. a typical LHD vehicle.
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -151,7 +151,7 @@ DriverPosition:
   allowed: ['LEFT', 'MIDDLE', 'RIGHT']
   default: 'LEFT'
   description: The position of the driver seat in row 1.
-  comment: Default value is position 'LEFT', i.e. a typical LHD vehicle.
+  comment: Default value is position 'LEFT', i.e. a typical LHD vehicle. Some signals use DriverSide and PassengerSide as instances. If this signal specifies that DriverPosition is LEFT or MIDDLE, then DriverSide refers to left side and PassengerSide to right side. If this signal specifies that DriverPosition is RIGHT, then DriverSide refers to right side and PassengerSide to left side.
 
 SeatRowCount:
   datatype: uint8
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -1,4 +1,5 @@
 #
+# (C) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
 #
@@ -109,7 +110,7 @@ Door:
   type: branch
   instances:
     - Row[1,2]
-    - ["Left","Right"]
+    - ["Driver","Passenger"]
   description: All doors, including windows and switches.
 #include SingleDoor.vspec Door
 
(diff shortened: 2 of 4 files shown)


spec/Cabin/Cabin.vspec
```diff
@@ -109,7 +110,7 @@ Door:
   type: branch
   instances:
     - Row[1,2]
-    - ["Left","Right"]
+    - ["Driver","Passenger"]
```
SebastianSchildt, line 113: I am not convinced it makes sense to call a backdoor "Passenger" door?

spec/Cabin/Cabin.vspec
```diff
@@ -130,33 +131,27 @@ DoorCount:
...
-    - Pos[1,3]
+    - ["Driver","Middle","Passenger"]
   description: All seats.
 #include SingleSeat.vspec Seat
 
 #
 # Seat attributes.
 #
 DriverPosition:
-  datatype: uint8
```
SebastianSchildt, line 155: This makes a lot of sense, but wouldn't it then be better not touching the seats/doors because while it is useful for the front row, it seems a little weird for everything else.

Also, may we want to add "center" to allowed (basically all Vehicles having only one seat, Tesla Semi, Forklifts, racecars).

spec/Cabin/HVAC.vspec
```diff
@@ -14,7 +15,7 @@ Station:
   type: branch
   instances:
     - Row[1,4]
-    - ["Left","Right"]
+    - ["Driver","Passenger"]
```
SebastianSchildt, line 18: Same argument as above

spec/Cabin/InteriorLights.vspec
```diff
@@ -46,7 +47,9 @@ LightIntensity:
 #
 Spotlight:
   type: branch
-  instances: Row[1,4]
+  instances: 
+    - Row[1,4]
```
SebastianSchildt, line 51: Same argument

SebastianSchildt (commented):
Basically I agree information about where the driver sits is missing. My feeling is, it is enough to add this information and not change all left/right instances, as that does not feel very logical to me. But always ready to be convinced, overvoted by the majority :)

Maybe call it larboard/starboard :D

spec/Cabin/Cabin.vspec
adobekan: Think about someone who is delivering use-cases which are driver focused, and now add complexity of LHD and RHD vehicles. "Left" and "Right" is challenging proposal to understand and implement, then you have seat concepts Pos1,2,3.
I understand potential of describing large vehicles, but still our starting point is passenger vehicle.

And I prefer Latin Auriga.

erikbosch:
Meeting notes: Please review/comments. To be further discussed Feb 21st. Possibly wait with merge until 4.0

spec/Cabin/Cabin.vspec
erikbosch: Terms like `VDAdata_doorPassengerRear` can be found sometimes

spec/Body/Body.vspec
```diff
@@ -242,7 +242,7 @@ Lights.DirectionIndicator:
 ##
 Mirrors:
   type: branch
-  instances: ["Left", "Right"]
+  instances: ["Driver", "Passenger"]
```
erikbosch, line 245: If we go for this we could possibly consider redefining it for motorbikes as left/right in [the motorbike profile](https://github.com/COVESA/vehicle_signal_specification/blob/master/overlays/profiles/motorbike.vspec)

spec/Cabin/Cabin.vspec
erikbosch: Or "middle" as used for Seat above, as long as we use same term (middle or center) I am happy.

But if we support middle/center for driver position, then we need to discuss how to handle other signals that possibly use Driver/Passenger as instances, like doors, mirrors, (rear) seats and so on. I can see a few options:

- Consider those vehicles to not have "driver-side", and require them to use an overlay to redefine instances just as we might do for motorbikes.
- Add another signal/attribute `Vehicle.LogicalDriverPosition` that just contains Left and Right which can be used to identify for example which mirror which is considered to be the "driver mirror" and which is the "passenger mirror". (I was thinking first about a signal indicating if a vehicle is intended for left side traffic or right side traffic, and deduce logical driver side from that, but then I realized that there sometimes are corner cases like postal-vehicles where the driver sits on the "wrong side" for convenience reasons)

SebastianSchildt:
So, thinking about this, it is not super easy to find (publicly) accessible examples, what is "better"/more common


In [this DBC](https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/Model3CAN.dbc) there is quite a bit of left/right, e.g. https://github.com/eclipse/kuksa.val.feeders/blob/f6b981f08bb2cc1247e944358a2d65eaca4a0803/dbc2val/Model3CAN.dbc#L135 but driver passenger can also be found e.g. https://github.com/eclipse/kuksa.val.feeders/blob/f6b981f08bb2cc1247e944358a2d65eaca4a0803/dbc2val/Model3CAN.dbc#L789 and https://github.com/eclipse/kuksa.val.feeders/blob/f6b981f08bb2cc1247e944358a2d65eaca4a0803/dbc2val/Model3CAN.dbc#L1945 

Generally, I think `["passenger", "driver"]`  is not a good wording when using this to refer to "positions", so while I like "left" and "right", for going the "passenger"-route I would suggest using `passengerside` and `driverside`, so in case somebody encounters this in some API somehwhere in the stack, he will at least think twice (and depending on his usecase, i.e. when visualising something, search for the datapoint that tells him which side that is)

erikbosch:
Meeting discussion: Left/Right may be preferably on hw-side, on sw-side Driver/Passenger is better.
Adnan to push changes this week. Alignment on PR next week.

SebastianSchildt:
Just cam across another datapoint sw-side: Andorid VHAL  is also using LEFT/RIGTH, for Seats and HVAC
```
ROW_1_LEFT = 0x0001
ROW_1_CENTER = 0x0002
ROW_1_RIGHT = 0x0004
ROW_2_LEFT = 0x0010
ROW_2_CENTER = 0x0020
ROW_2_RIGHT = 0x0040
ROW_3_LEFT = 0x0100
```
https://source.android.com/docs/devices/automotive/vhal/properties

spec/Cabin/Cabin.vspec
ppb2020: Perhaps consider instead referring to the Seat position, being Row<n> and Pos<m>?

erikbosch:
Meeting notes:

- Adnan: Think it is better than before
- Sebastian: Better to keep.

Please vote yes/no until next week, Merge decision next week.

SebastianSchildt:
I am still not convinced the new approach is "better", but I'd request my "vote" to be counted as "abstain". I can live with the change

spec/Body/Body.vspec
```diff
@@ -242,7 +242,7 @@ Lights.DirectionIndicator:
 ##
 Mirrors:
   type: branch
-  instances: ["Left", "Right"]
+  instances: ["DriverSide", "PassengerSide"]
   description: All mirrors.
```
erikbosch, line 246: To cover odd vehicles like McLaren F1 and motorbikes should we possibly add a comment:
`For vehicles where driver position is Middle the left mirror is considered to be the DriverSide the mirror`

erikbosch:
Meeting notes:

- Adnan to add comment on how to handle DriverSide/PassengerSide when driver is in middle (e.g. motorbike mirrors), then ready to merge

erikbosch (approved):

Issue #531, opened 2023-02-07
Improve specification of InteriorLights


jdacoello:
### LightIntensity as actuator
In [vehicle_signal_specification/spec/Cabin/InteriorLights.vspec](https://github.com/COVESA/vehicle_signal_specification/blob/d5b0c2173e383e3003b54f1a4bc4075988ee7d53/spec/Cabin/InteriorLights.vspec), `LightIntensity` is defined as `sensor`. Should it not be an `actuator` instead?

```
LightIntensity:
  datatype: uint8
  type: sensor
  unit: percent
  min: 0
  max: 100
  description: Intensity of the interior lights. 0 = Off. 100 = Full brightness.
```

### Lights individually controlable
Right now the intensity is referring to all the interior lights. 

### Adding other aspects
For example:
- Color for each light
- Flex (decorative) lights

erikbosch:
Fixed by #587

PR #540, opened 2023-02-14
Interior lights

jdacoello:
feat: Improve the specification of interior lights to address issue #531 . Key aspects are listed below:

* `SingleConfigurableLight.vspec` defined for generic lights whose color, intensity, and state can be acted on.
* Changed previous name of `AmbientLight` to `PerceivedAmbientLight`. Previous definition was referring to the act of sensing the existing light in the ambient. 
* `AmbientLight` redefined as `actuator`.
* Examples of configurable lights included: `AmbientLight` and `Spotlight`
* Add new concept for interactive lights, where an effect can be selected.

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -51,6 +51,12 @@ Trunk.IsLocked:
   type: actuator
   description: Is trunk locked or unlocked. True = Locked. False = Unlocked.
 
+Trunk.IsLightOn:
+  datatype: boolean
+  type: actuator
+  description: Is trunk light on
+  comment: V3.0 Moved from Vehicle.Cabin.Lights.IsTrunkOn
+
 
 #
 # Horn description
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -1,5 +1,5 @@
 #
-# (C) 2023 BMW Group
+# (C) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
 #
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -1,4 +1,5 @@
 #
+# (C) 2023 BMW Group
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
 #
--- a/spec/Cabin/InteriorLights.vspec
+++ b/spec/Cabin/InteriorLights.vspec
@@ -1,4 +1,5 @@
 #
+# (C) 
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
 #
@@ -10,6 +11,9 @@
 # All interior lights and sensors
 #
 
+#
+# Non-configurable lights
+#
 IsGloveBoxOn:
   datatype: boolean
   type: actuator
@@ -18,20 +22,22 @@ IsGloveBoxOn:
 IsTrunkOn:
   datatype: boolean
   type: actuator
-  description: Is trunk light light on
+  description: Is trunk light on
+  deprecation: V3.0 moved to Vehicle.Body.Trunk.isLightOn as the Trunk is not defined as part of the Cabin.
 
 IsDomeOn:
   datatype: boolean
   type: actuator
   description: Is central dome light light on
 
-AmbientLight:
+PerceivedAmbientLight:
   datatype: uint8
   type: sensor
   unit: percent
   min: 0
   max: 100
-  description: How much ambient light is detected in cabin. 0 = No ambient light. 100 = Full brightness
+  description: The percentage of ambient light that is measured (e.g., by a sensor) inside the cabin. 0 = No ambient light. 100 = Full brightness.
+  comment: See also Vehicle.Cabin.Lights.AmbientLight, which is used for acting on the decorative coloured lights available in the cabin.
 
 LightIntensity:
   datatype: uint8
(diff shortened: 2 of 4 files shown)


erikbosch:
Meeting notes: Please review/comment. To be discussed at meeting Feb 21st.

erikbosch:
PR based on old VSS version. I suggest rebasing on latest master to make sure that shown diff is up to date.

spec/Body/Body.vspec
```diff
@@ -51,6 +52,12 @@ Trunk.IsLocked:
   type: actuator
   description: Is trunk locked or unlocked. True = Locked. False = Unlocked.
 
+Trunk.IsLightOn:
+  datatype: boolean
+  type: actuator
+  description: Is trunk light on
+  comment: V3.0 Moved from Vehicle.Cabin.Lights.IsTrunkOn
```
erikbosch, line 59: Here I see two options. Either we plan this PR for the minor release planned for end of February (v3.1), or for v4.0 planned for end of April. If we go for v4.0 we do not need to care about deprecation-comments, we just need to make sure it is mentioned in the CHANGELOG file. I think v4.0 is the best approach, as we otherwise for v3.1 would need to have both `Light` and `Lights` branch present.

(If we go for the minor release the comment shall say v3.1, as that is the first release where it will be deprecated.  This might be a side effect of using an old base branch, compare e.g. your [VERSION file](https://github.com/jdacoello/vehicle_signal_specification/blob/interior-lights/VERSION) with the latest [VERSION file](https://github.com/COVESA/vehicle_signal_specification/blob/master/VERSION)

spec/Cabin/SingleConfigurableLight.vspec
```diff
@@ -0,0 +1,29 @@
...
+IsLightOn:
+  type: actuator
+  datatype: boolean
+  description: Indicates whether the light is turned on. True = On, False = Off.
+
+Intensity:
+  type: actuator
+  datatype: uint8
+  unit: percent
+  min: 1
```
erikbosch, line 21: How come we have 1 as minimum? Can they not be completely turned off?

spec/Cabin/SingleConfigurableLight.vspec
```diff
@@ -0,0 +1,29 @@
...
+  max: 100
+  description: How much of the maximum possible brightness of the light. is used. 1 = Maximum attenuation, 100 = no attenuation (i.e., full brightness).
+  comment: Attenuation = 100 - Intensity.
+
+Color:
+  type: actuator
+  datatype: string
+  description: Hexadecimal color code.
```
erikbosch, line 29: Here I would prefer if we can refer to standard/spec/wiki on how the color code shall be formatted, or just describe it. Like `Color shall be given as a 3 byte RGB value preceded by a hash, allowed range #000000 to #FFFFFF` and potentially include an example like `Examples: #C0C0C0 = Silver, #FFD700 = Gold`

One could of course argue that a uint32 (with range) would work as well.

spec/Cabin/SingleConfigurableLight.vspec
jdacoello: I had initially the same question. However, after some discussions, I understood the reason behind that decision.
The idea is that a light `Intensity` controller can be used to attenuate (i.e., dimmerize) the light. Therefore, its main purpose is not to serve as a switch, but rather to set the intensity. 

In this sense, for example, you might have a light configured with an intensity of 50%. This configuration can be stored for a particular user profile. Apart from that, you can have the switch `IsLightOn` to just turn it on/off without the need to remember the previous state of the intensity. 

That is why the `Intensity` has a minimum value of 1%, so that it is not used as a switch. Which means that you can set the intensity to 100%, but the light is off unless you turn on the switch `IsLightOn`.

spec/Body/Body.vspec
jdacoello: I am Ok with any of them. 

* If it is the minor release, I can also name it `Lights` instead, to be compatible with the existing branches. And then I can create a new issue to address the Singular names in branches.
* If it is the major release, I think there are some other aspects about the branches that can be considered. So far, I see that most of the effort is centred on the leaves of the tree. However, less attention is paid to the branches and the criteria by which the tree is categorising the information.

spec/Cabin/SingleConfigurableLight.vspec
erikbosch: I understand. We could possibly add something in a comment, possibly something like "Minimum value not zero as on/off is controlled by the signal IsLightOn", there will likely be more that will have the same question.

ppb2020: Assume there are a separate switch and a separate dial for intensity. The switch can be on, with the intensity zero. The switch can be off, with the intensity 50%.
 
No point in trying to represent or avoid representing the switch state in the intensity: they are independent. That is, don’t limit the intensity to anything else than 0-100%...

spec/Body/Body.vspec
jdacoello: @erikbosch FYI, I will adapt the PR to target the major release v.4.0. If the proposal is accepted, it should be merged in March.

erikbosch: I think that makes sense, then we do not need the "deprecated" stuff. Also good from a timing perspective, as we have planned VSS 3.1 for end of February I would more or less like to have code freeze this week, and I believe this one needs to saty open until at least next week.

spec/Cabin/SingleConfigurableLight.vspec
jdacoello: Hexadecimal value is not part of the [supported datatypes in VSS](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types/). I think simply defining it as an integer with base 10 can cause confusion as the user would be responsible for making the conversion. 

Another option is to have `Color.Red`, `Color.Green`, `Color.Blue` with min and max values (i.e., from 0 to 255). However, I think that approach would be too verbose. 

As for now, we can keep the hexadecimal representation with string datatype. I will add the recommended comments in the definition.

jdacoello:
All comments were addressed, and the proposed specification was squashed into one commit.

spec/Cabin/InteriorLights.vspec
```diff
@@ -10,57 +11,56 @@
...
+  type: branch
+  description: Decorative coloured light bar that supports effects. It is usually mounted on the dashboard (e.g., BMW i7 Interactive bar).
+#include SingleConfigurableLight.vspec InteractiveLightBar
 
-Spotlight.IsRightOn:
-  datatype: boolean
+InteractiveLightBar.Effect:
   type: actuator
-  description: Is light on the right side switched on
+  allowed: ['NO_EFFECT', 'GRADIENT_DRIVER', 'GRADIENT_PASSENGER']
```
erikbosch, line 63: I would like to have a short comment on what is meant by gradient.

Does "GRADIENT_DRIVER" means that the effect is mostly (brightest) on driver side?
Does "NO_EFFECT" means "turned off", or rather "symmetrical effect"?

jdacoello: Those effects' names are just given as examples for possible default values. The actual values in practice will be implementation- or OEM-specific.
- `GRADIENT_DRIVER` means that the light bars has more intensity on the driver's side, and the intensity fades down towards the passenger.
- `GRADIENT_PASSENGER` means that the light bars has more intensity on the passenger's side, and the intensity fades down towards the driver.
- `NO_EFFECT` means that the whole light bar is illuminated with the same intensity.

erikbosch: If they are just examples and will vary then it maybe fits better as a comment.

erikbosch: @jdacoello - if it is just examples I would like it to instead be a comment, similar to what was added in https://github.com/COVESA/vehicle_signal_specification/pull/535/files

jdacoello: Example allowed values were removed. Information was added to the comment field.

erikbosch:
Meeting notes: Awaiting left/right vs driver/passenger decision

erikbosch:
Meeting notes:

- When #538 is merged, update this PR to use the same terms for DriverSide/PassengerSide

spec/Body/Body.vspec
claireqiu, line 1: Thanks for the PR. 
Small comment, In line 232, it is missing the "descritpion" of Lights.Brake.  The was the error occurred while running the vespc2ttl tool.

erikbosch: That should have been fixed by https://github.com/COVESA/vehicle_signal_specification/pull/559. So if this PR is rebased that should hopefully be included :-)

jdacoello: I added the missing description.

jdacoello:
Updated to use `DriverSide`and `PassengerSide`

erikbosch:
There is a conflict that needs to be fixed

erikbosch:
@jdacoello - would be great if you could update this PR to solve the conflicts

erikbosch:
Hi @jdacoello - For convenience I created a clean PR at #587 for both of your PRs, starting from a clean master. There the build succeeds. I think we can either take that one and use it for merge. What I did was taking the latest master and then replacing file content with those from your PRs, to get two clean and signed-off commits. The only additional change was to fix to linter errors depending on old code in the files.

erikbosch:
Merged by #587

Issue #544, opened 2023-02-17
Wrong types defined in OBD.vspec


jdacoello:
Some corrections are needed for the types defined in [OBD.vspec](https://github.com/COVESA/vehicle_signal_specification/blob/2b24d0d9ac6b2d357e1d97ca3064cc2ca5bfa089/spec/OBD/OBD.vspec).

For example:

```
FuelType:
  datatype: string
  type: sensor (it should be attribute)
  description: PID 51 - Fuel type
```

```
Status.IgnitionType:
  datatype: string
  type: sensor (it should be attribute)
  allowed: ['SPARK', 'COMPRESSION']
  description: Type of the ignition for ICE - spark = spark plug ignition, compression = self-igniting (Diesel engines)
```

BTW, what is the take on OBD data described in VSS? Is this actually used?

erikbosch:
Have not heard about anyone actually using them. I have done some corrections in the past to solve errors. Feel free to create a PR for the errors above.

erikbosch:
Merged by #587

PR #546, opened 2023-02-21
fix: Correct types in OBD specification

jdacoello:
It addresses the issue #544. Minor changes in the types described for OBD data.

### DIFF
--- a/spec/OBD/OBD.vspec
+++ b/spec/OBD/OBD.vspec
@@ -1,4 +1,5 @@
 #
+# (C) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
 # (C) 2020 Robert Bosch GmbH
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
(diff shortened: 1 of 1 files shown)


spec/OBD/OBD.vspec
```diff
@@ -33,9 +34,10 @@
 #
 
 PidsA:
-  datatype: uint32
-  type: sensor
-  description: PID 00 - Bit array of the supported pids 01 to 20
+  datatype: string[]
+  type: attribute
+  allowed: ["01",	"02", "03",	"04",	"05",	"06",	"07",	"08",	"09",	"0A",	"0B",	"0C",	"0D",	"0E",	"0F",	"10",	"11",	"12",	"13",	"14",	"15",	"16",	"17",	"18",	"19",	"1A",	"1B",	"1C",	"1D",	"1E",	"1F",	"20"]
```
erikbosch, line 39: Seems to be some tabs on this line which our tooling does not like, that is the reason the build faols

spec/OBD/OBD.vspec
```diff
@@ -33,9 +34,10 @@
...
+  description: PID 00 - Bit array of the supported pids 01 to 20 in Hexadecimal.
```
erikbosch, line 40: Previously it was a bit array, i.e. a 32 bit uint where each bit represented a pid. Now it is a real array instead so shall better be "PID 00 - Array of ..."

By the way - if we do this change, should we then not change PidsB and PidsC as well?

erikbosch:
@jdacoello - please review comments

erikbosch:
@jdacoello - would be great if you could update this PR according to comments

spec/OBD/OBD.vspec
jdacoello: Done

spec/OBD/OBD.vspec
jdacoello: Done. Pids B and C were updated to match the definition of A.

erikbosch:
Merged by #587

PR #549, opened 2023-02-22
Making Window Positon and ChildLock an Actuator

SebastianSchildt:
This makes `Window.Position` and `Window. IsChildLockEngaged` actuator. I think it being a sensor is a bug, as a VSS App might want to set Window Position (I think I have  actually even seen this in examples on ppt level from time time)

I kept `IsOpen`  as pure sensor, as I interpret that to be true for any position != 0, so setting it has no real meaning.

There _are_ Window controls modelled in form of a switch in Window.Switch

```
Switch:
  datatype: string
  type: actuator
  allowed: ['INACTIVE', 'CLOSE', 'OPEN', 'ONE_SHOT_CLOSE', 'ONE_SHOT_OPEN']
  description: Switch controlling sliding action such as window, sunroof, or blind.
```

But I think that is not the same/complimentary. 

Example: If you set a certain Window position in for example in an app, or some hypothetical function in the car that prevents a dog from being grilled in a parked car, likely you would like to set the window to certain position directly and not "simulating" button presses

### DIFF
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -121,6 +121,12 @@ DoorCount:
   description: Number of doors in vehicle.
 
 
+IsWindowChildLockEngaged:
+  datatype: boolean
+  type: actuator
+  description: Is window child lock engaged. True = Engaged. False = Disengaged.
+  comment: Window child lock refers to the functionality to disable the move window button next to most windows,
+           so that they only can be operated by the driver.
 
 ##
 # Seat signals and attributes
--- a/spec/Cabin/SingleWindow.vspec
+++ b/spec/Cabin/SingleWindow.vspec
@@ -25,6 +25,7 @@ Position:
 IsChildLockEngaged:
   datatype: boolean
   type: actuator
+  instantiate: False
   description: Is window child lock engaged. True = Engaged. False = Disengaged.
 
 # Include the window controlling switch and attach it to the
--- a/spec/Cabin/SingleWindow.vspec
+++ b/spec/Cabin/SingleWindow.vspec
@@ -22,12 +22,6 @@ Position:
   unit: percent
   description: Window position. 0 = Fully closed 100 = Fully opened.
 
-IsChildLockEngaged:
-  datatype: boolean
-  type: actuator
-  instantiate: False
-  description: Is window child lock engaged. True = Engaged. False = Disengaged.
-
 # Include the window controlling switch and attach it to the
 # window branch
 #include SingleSliderSwitch.vspec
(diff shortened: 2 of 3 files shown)


erikbosch:
I think the current setup reflects some old thoughts were the actuators reflected what you typically can do from within a vehicle cabin. With that mindset it is "correct" to not have window position as actuator. But that mindset we have abandoned quite long time ago and for seats we have changed so that you now have both "buttons" and "position" as mechanisms to control them, so it makes sense here as well.

spec/Cabin/SingleWindow.vspec
```diff
@@ -16,15 +16,15 @@ IsOpen:
...
 Position:
   datatype: uint8
-  type: sensor
+  type: actuator
   min: 0
   max: 100
   unit: percent
   description: Window position. 0 = Fully closed 100 = Fully opened.
 
 IsChildLockEngaged:
```
erikbosch, line 25: This could possibly be a good opportunity to discuss if this signal needs more improvements in the long run (not necessarily part of this PR). Is this actually a signal that is needed for every window or would a joint "cabin" signal be sufficient. In most cars it is a single button, and it disables all buttons except the buttons controlled by the driver. Are there vehicles were you can engage child lock individually for windows? If not, would it possibly suffice with a single signal.

One can also argue that our model does not consider that there typically are two switches for most windows (except for the driver window). If `IsChildLockEngaged==True` then the switch next to the window shall be disabled, but normally not the switch next to the driver. But with the current setup with a single switch below it is not possible by the VSS signals to identify which switch that is used.

erikbosch:
Meeting notes: Please review, merge decision next week.

spec/Cabin/SingleWindow.vspec
SebastianSchildt: I am open to moving it out, either here or in a separate PR. I think `instantiate=false` would do the trick?

adobekan: @erikbosch 
Usually it is always one switch for Row2 windows. Might be logical to have this on higher level.

erikbosch:
Meeting notes: Sebastian to update to move childlock up "one level". Merge decision next week

erikbosch:
@SebastianSchildt - will you update to move childlock up "one level"?

erikbosch:
Meeting notes: Erik to check if last change is as accepted, if so merge

spec/Cabin/Cabin.vspec
```diff
@@ -121,6 +121,12 @@ DoorCount:
   description: Number of doors in vehicle.
 
 
+IsWindowChildLockEngaged:
```
erikbosch, line 124: @SebastianSchildt - it was not sufficient to add `instantiate: False` as the signal was in the Window branch which is not instantiated as each door has only one instance. Now moving it up to Cabin level. Please check

erikbosch (approved):

PR #555, opened 2023-02-27
Add battery cell voltage information

erikbosch:
We already have temperature-related information to cells, this one adds voltage.

### DIFF
--- a/spec/Powertrain/Battery.vspec
+++ b/spec/Powertrain/Battery.vspec
@@ -60,6 +60,22 @@ Temperature.Max:
   unit: celsius
   description: Current maximum temperature of the battery cells, i.e. temperature of the hottest cell.
 
+CellVoltage:
+  type: branch
+  description: Voltage information for cells in the battery pack.
+
+CellVoltage.Min:
+  datatype: float
+  type: sensor
+  unit: V
+  description: Current voltage of the battery cell with lowest voltage.
+
+CellVoltage.Max:
+  datatype: float
+  type: sensor
+  unit: V
+  description: Current voltage of the battery cell with highest voltage.
+
 GrossCapacity:
   datatype: uint16
   type: attribute


erikbosch:
Meeting notes: Please review. Merge decision possibly next week
Sebastian: How can cell voltage be used - how useful it
Raikar: Is cell if needed.

erikbosch:
@vraikar-covesa - the background reason for adding these two is to have something that corresponds to the PIDS 0xB9/0xF4B9 in the standard SAE J1979DA_202203.xlsx

That is however as I see it not in conflict to that we could add individual cell information as well. But there it could be good to discuss how we want to represent it. I am no cell expert. I can see that we either use the branch-instance-concept so you address them like you propose, or as one or more arrays (of either primitive type or even struct type) where we include both voltage and identifier. I think it partially depends on how OEMs address individual cells. Do they typically have consecutive identifiers from for example 1 to 100, or do they have more cryptic identifiers like "cell_3_driver_seat_lower". If naming is not standardized I assume the standard catalog then would contain only "example names" in the instance-description.

neil-rti:
Not a battery expert, but when I hear 'cell' I think "small cylinder, like an 18650" - There can be thousands of these cells in an EV.  Is this to measure every single cell, or larger groupings of cells within the entire battery?  How does the reader of this data know the battery configuration and expected range of these voltage measurements?

erikbosch:
Meeting notes:

- Check with charging expert group (AP: Erik)

erikbosch:
I got positive feedback from Felix Loesch, suggest merging.

erikbosch:
Reference to document requiring presence of this data: https://ww2.arb.ca.gov/sites/default/files/barcu/regact/2022/accii/appa6.pdf

Data Stream: The following signals shall be made available on
demand through the standardized data link connector in
accordance with SAE J1979-3 and SAE J1979-DA specifications.
The actual signal value shall always be used instead of a default or
Appendix A-6
limp home value that may be substituted in place of the actual
signal by the receiving onboard electronic control for use by it or
other onboard electronic control units.

1. For all vehicles:

a. vehicle speed, absolute accelerator pedal position,
time elapsed since start of trip, odometer reading,
distance traveled since fault memory last cleared,
and number of propulsion system active trips since
fault memory last cleared

b. high voltage battery pack: state of charge, maximum
cell voltage, minimum cell voltage, battery system
voltage, cumulative battery system current for the
last 1 second, cumulative battery system energy (i.e.,
power via integrated voltage times current)
consumption for the last 1 second

erikbosch:
Meeting notes:

- Sebastian: As requested by regulation lets merge
- Decision: merge

SebastianSchildt (approved):
Needed by regulation -> Should add it.

Generally our HEV battery model still needs some love, but maybe not in this PR

Issue #548, opened 2023-02-21
Multiple and/or unclear criteria used for classifying the information in the tree


jdacoello:
# The issue
Considering the following:
- The VSS tree uses a tree-like hierarchy to define the name paths of the vehicle data points.
- A tree-like hierarchy can only cover one dimension effectively. This means that there is an implicit relationship between the branches.
- _**VSS is not a taxonomy**_ - When the implicit relationship is `ChildBranch --subClassOf--> ParentBranch`, we are talking about a taxonomy. Therefore, it is misleading to say that VSS is a taxonomy, because it does not comply with this criteria. For example: we cannot say that a `Cabin` is a sub class of a `Vehicle`, or that the `Hood` is a sub class of a `Body`. In contrast, saying that a `Car` is a sub class of a `Vehicle` is compliant with what a taxonomy does because the `Car` is a more specific type of a `Vehicle`.
- _**VSS seems to be a meronymy**_ - When the implicit relationship is `ChildBranch --partOf--> ParentBranch`, we are talking about a meronymy . Here is valid to say that `Cabin` is a part of the `Vehicle`, or that the `Hood` is part of the `Body`. 
 
However, there is no clear statement in the documentation indicating the way the information is (and must be) classified. Sometimes you find paths that represent the physical distribution of the vehicle components, such as `Vehicle.Cabin.Infotainment.HMI`. However, there are other uses of branches as a way to group (for convenience) some data points that should appear together, such as `Vehicle.VehicleIdentification`.

# Possible treatment
Some ideas to cope with this issue could be:
* Define ONLY ONE criteria for organising the information in the tree. For example: state that the implicit relationship between branches means `ChildBranch --partOf--> ParentBranch`.
* Enforce the use of this criteria in the definition of any branch.
* Use a different separator to indicate the instances. 
  * For example, consider the path `Vehicle.Cabin.Door.Row2.Left.IsLocked`. Here `Row2` is not part of the `Door` but it is actually one of the available doors. The same holds for `Left` not being part of `Row2`
  * To me, it makes more sense to have something like `Vehicle.Cabin.Door(Row2.Left).IsLocked`. Here, it is clearer that it is an instance of a `Door` that is part of the `Cabin`, which is part of the `Vehicle`

erikbosch:
[...]

PR #556, opened 2023-02-27
Move track to axle

erikbosch:
Background is that vehicles may have different track on front and rear axle.
If we are planning for VSS 4.0 no need to deprecate olf signal, we can just remove it.


adobekan:
Copy-paste
"Track Width Definition Wheel track or track width is the distance between the centerline of two wheels on the same axle. Some might confuse it with axle track, which is the distance between the hub flanges on an axle. In most scenarios, wheel track and axle track are the same length."

W101-TREAD-FRONT—The dimension measured between the tire centerlines at the ground.
W102-TREAD-REAR—The dimension measured between the tire centerlines at the ground. In case of dual wheels, the dimension will be measured to the centerline of tire and wheel assemblies..

Not sure if we consider SAE that axle is right position for track width. It seems that they are always considering middle of the wheel.

adobekan:
Meeting note 07.03.2023: Check the details of w101 and w102. Also, possible instances per axle.

erikbosch:
> 
> Not sure if we consider SAE that axle is right position for track width. It seems that they are always considering middle of the wheel.

It is a bit related to the ontology discussion in #548, as of today "Axle" contain both data directly "owned"/"controlled" by the axle, but also data that is related to the axle but not really "owned"/"controlled" by the axle. An example is `Axle.TireAspectRatio` where we currently expects that a vehicle has (or better should have) the same profile on left and right tire.

For track, we could possibly consider putting it in wheel but with "instantiate: false", i.e. so that it shows up as `Vehicle.Chassis.Axle.Row1.Wheel.Track` and `Vehicle.Chassis.Axle.Row2.Wheel.Track` We could off course also use a different location.

spec/Chassis/Chassis.vspec
```diff
@@ -43,6 +37,15 @@ AxleCount:
 #
 # Axle attributes
 #
+
+
+Axle.Track:
```
erikbosch, line 42: Based on the discussion and definition in J1100 Revised NOV2009 I see that we could have at 3 signals

- Axle.TreadWidth (W101, The lateral distance between the centerlines of the base tires at ground, including camber angle. If there are dual wheels, measure from the midway points between the inner and outer tires)
- Axle.TrackWidth (W102, Track Width at Wheel Center. If there are dual wheels, measure from the midway points between the inner and outer tires)
- Axle.Width (W103, The lateral distance between the wheel mounting faces, measured along the spindle axis. Measurement is independent of both wheel offset and camber angle.)

As of now the PR only include one of them (W102). 

(We could also discuss if we would like an Axle.WheelOffset as well, i.e. ET = einpresstiefe)


I think having them in Axle branch is OK, if we consider the Axle branch to contain both information on the Axle itself, but also information concerning Axle including all wheels mounted to it.

erikbosch:
Meeting notes:

- Adnan to check, if approved ready to merge

adobekan (approved):
Looks good.

PR #566, opened 2023-03-17
proposal for stringlength keyword

neil-rti:
This proposes adding a `stringlength` keyword to the VSS vspec files that use string data types -- intended as a starting point for discussion.

### DIFF
--- a/stringlength.md
+++ b/stringlength.md
@@ -1,0 +1,112 @@
+# Proposal to add Suggested Lengths for String Signals in the VSS
+
+This document covers a concept proposal to add to the VSS .vspec files an additional keyword 
+(`stringsize:` or similar) to assist the code generators (in vss-tools) in setting an appropriate
+default size for strings.
+
+This new keyword would be added to the .vspec files for all `datatype: string` elements (approx. 85 places),
+and would also affect the arrays-of-strings (`datatype: string[]`, approx 6 places) to set a
+correctly-sized default boundary length on these string members.
+
+A few additional questions have been added below as well, for discussion in VSS meetings.
+
+
+## Rationale
+
+When creating software for a system with safety, security, or resource constraints, the use of 
+unbounded variables is strongly discouraged.   Unless extra steps are taken to limit their size,
+they can act as a vector for instability and exploitation.
+
+In practice, the signals in the VSS will never require unbounded lengths.  Some string signals
+in the VSS include the length in its `description:` field, but this information is not available for 
+every `string` signal.  Specifying the `stringlength`, and the `arraysize` of VSS elements can help
+create more concise implementations of the VSS, and reduce the burden on users.
+
+
+## Proposed Implementation
+
+The proposed implementation would add in the .vspec files a new key:value pair (herein called `stringlength:`) 
+to set the string length of any signal with a `string` or `string[]` datatype.  
+When used with arrays-of-strings (`string[]`), this keyword sets the max length of the strings in the array, while
+the existing `arraysize` keyword sets the number of strings in the array.
+
+
+## Examples from VSS
+
+An example implementation might look like:
+```
+VehicleIdentification.VIN:
+  datatype: string
+  stringlength: 17
+  type: attribute
+  description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
+```
+
+An array-of-strings example might be:
+```
+DTCList:
+  datatype: string[]
+  stringlength: 6
+  arraysize: 20
+  type: sensor
+  description: List of currently active DTCs formatted according OBD II (SAE-J2012DA_201812) standard ([P|C|B|U]XXXXX )
+```
+In this example, `stringlength` was estimated from the text in `description`, and `arraysize` was set to 
+a value (20) based on an unqualified guess of what a reasonable number of DTCs should be.
+
+In some instances, the string length and array dimensions are implied by the use of `allowed` elements, 
+and do not need to be explicity stated:
+```
+SmartphoneProjection.SupportedMode:
+  datatype: string[]
+  type: attribute
+  allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
+  description: Supportable list for projection.
+```
+In this case, if this were to be implemented as an array-of-strings the `arraysize` can be obtained from the 
+length of the `allowed` list, and the `stringlength` can be found from the length of the longest string in
+the `allowed` list.   The `stringlength` concern does not apply if this is implemented as an array of `enum`
+(integer) values created from the `allowed` list.  (see question below about 'Allowed Values')
+
+
+## Questions Raised in Writing this Proposal
+
+### Allowed Values: Is there a recommendation or enforcement on how this should be implemented in code?
+
+In the VSPEC files, most `string` members are used to hold `allowed:` values, contained in a list of strings.
+This list could be implemented in code as an `enum`, or as a verbatim array of strings, or using an order-preserving
+container such as: `sequence` / `vector` / `list` of strings.  In each case, the index/enumeration would be 
+the same (enum 0 == index 0, etc.).  Is there any guidance on this?   
+(this is a potential place for improving the Wiki documentation.)
+
+### Should this be used to specify fixed sizes, or an allowed range (min:max) of sizes for the element?
+
+Open for discussion.
+
+### Should this accommodate NULL-termination on strings, or be sized exact?
+
+Open for discussion.
+
+### Can `arraysize` be implemented on the arrays in the VSPEC files?
+
+For example:
+```
+SeatPosCount:
+  datatype: uint8[]
+  type: attribute
+  default: [2, 3]
+  description: Number of seats across each row from the front to the rear.
+  comment: Default value corresponds to two seats in front row and 3 seats in second row.
+```
+
+A default value is used here and it *implies* the dimension of the array, but it also could be
+misread as "this array has a default size of 2x3 elements".  This seems like a good place to be explicit in the
+array dimensions, by including the `arraysize` keyword and value.
+Apologies if this has been implemented in a *development* branch of the VSS.
+
+
+## Impact to Existing Uses
+
+With limited testing, this new keyword appears to be ignored by the parser in VSS-TOOLS, so it should
+be safe to add without breaking compatibility.   Should this be implemented in VSS, then it should
+also be supported in VSS-TOOLS, for use in the generators that are capable of setting variable sizes.


stringlength.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Questions Raised in Writing this Proposal
+
+### Allowed Values: Is there a recommendation or enforcement on how this should be implemented in code?
+
+In the VSPEC files, most `string` members are used to hold `allowed:` values, contained in a list of strings.
+This list could be implemented in code as an `enum`, or as a verbatim array of strings, or using an order-preserving
+container such as: `sequence` / `vector` / `list` of strings.  In each case, the index/enumeration would be 
+the same (enum 0 == index 0, etc.).  Is there any guidance on this?   
+(this is a potential place for improving the Wiki documentation.)
```
erikbosch, line 80: It is a recurring discussion if VSS needs enums, or if `allowed` can be considered/used as enums. From a strict VSS perspective the data is still of the underlying type (i.e. most often string) and if you have an "untyped" API like VISS we expect that you get the string value if using a "get"/"read", rather than an index or enum-value. This off course does not prevent "typed" APIs to represent the value in some other form.

There have been discussions now and then to "simplify" mapping to enums in various languages, for instance by stating that the first value always shall be the "default" value, that it shall be possible to specify numerical values for the literals (e.g. to be able to say that one of them has value 0 == default). But as of now that has been considered to be "out of scope" of VSS. So the only thing we do is to give some naming guidelines in https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/ 

As I see it this might change as we in COVESA starts discussing/defining/recommending how a Vehicle API shall look like. Then we might end up in a recommendation on how signals with `allowed` shall be managed in an API.

SebastianSchildt:
I think the argument, why as an implementor , you  would need/want a bounded length in certain conditions is obvious.

Let's assume we would agree on some way to describe this (i.e. adding a stringlength attribute to the description of signals in the spec), I feel the larger discussion is, how to use it:

 - would we "use" it in default service catalogue? (I guess not, as well, mostly there are no good defaults. Well _Maybe_ in DTCList because maybe those are always Pxxx or Uxxx, but that is maybe the only corner case
 - would we define it as "official" keyword (like arraylength, like aggregate), i.e. supported/generated by tooling as a default, or would we define it as an attribute for an overlay (i.e. how dbc mpppings are epxressed in KUKSA)

I think this is a relevant discussion (although maybe not exactly the same as this one), because due to historic reasons we are not very consistent. If we look at the current state of "deployment" information defined (e.g. information that might change with each implmentation of VSS in a specific vehicle, we have

| Deployment Information | Used in std catalogue | Supported in std tools | Comment/Opinion                                                                                                                                                                                                                                                                                                                                                                                    |
|------------------------|-----------------------|------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `instance`             | yes                   | yes                    | Should maybe not be defined in std catalgue (no good default),  on the other hand, might still want to define some rules (i.e. consider left/right passenger/driver discussion). (Side note:  Android seems to have generic - albeit not very amazing - concept  of zones of "Positions)                                                                                                           |
| `aggregate`            | no (I think)          | yeah -sort of?         | I think tools largely ignore it? Here it is a bit unclear whether it  "should" be used in std catalgue (i.e. wouldn't it be super helpful for GPS) or whether this is more a marker you can use in your impl, to define what  belongs together. Also a question, with struct support, whether this has relevance, and what it would be                                                             |
| `arraysize`            | no                    | yes                    | I think this a sane choice right here: Might be needed for use cases, implementations, but no good default choice available in std catalogue                                                                                                                                                                                                                                                       |
| `default`              | yes                   | yes                    | I think this is still used too much in std catalogue. For VSS version it is fine, for `Wheelbase` nad `Door.Count` not so much                                                                                                                                                                                                                                                                     |
| example: `dbc`         | no                    | no                     | One example for some thirdparty overly-attributes: KUKSA defines some keys to allow mapping of DBC defined CAN signals to VSS signals. Obviously not used in std catalogue and not specifically supported in std tools (except via generic overlay  format). And this seems the correct choice (Probably even if in COVESA at one  point we would agreee to define a "recommended" way to map CAN) |

Where does this leave something like `stringlength`. In my view 

| Deployment Information | Used in std catalogue | Supported in std tools | Comment/Opinion                                                                                                                                                                                                                                                                                                                                                                                    |
|------------------------|-----------------------|------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `instance`             | yes                   | yes                    | Should maybe not be defined in std catalgue (no good default),  on the other hand, might still want to define some rules (i.e. consider left/right passenger/driver discussion). (Side note:  Android seems to have generic - albeit not very amazing - concept  of zones of "Positions)                                                                                                           |
| `aggregate`            | no (I think)          | yeah -sort of?         | I think tools largely ignore it? Here it is a bit unclear whether it  "should" be used in std catalgue (i.e. wouldn't it be super helpful for GPS) or whether this is more a marker you can use in your impl, to define what  belongs together. Also a question, with struct support, whether this has relevance, and what it would be                                                             |
| `arraysize`            | no                    | yes                    | I think this a sane choice right here: Might be needed for use cases, implementations, but no good default choice available in std catalogue                                                                                                                                                                                                                                                       |
| `default`              | yes                   | yes                    | I think this is still used too much in std catalogue. For VSS version it is fine, for `Wheelbase` nad `Door.Count` not so much                                                                                                                                                                                                                                                                     |
| example: `dbc`         | no                    | no                     | One example for some thirdparty overly-attributes: KUKSA defines some keys to allow mapping of DBC defined CAN signals to VSS signals. Obviously not used in std catalogue and not specifically supported in std tools (except via generic overlay  format). And this seems the correct choice (Probably even if in COVESA at one  point we would agreee to define a "recommended" way to map CAN) |



Coming back to the orginal discussion, it may be worth thinking about having a "generic" format string for strings, i.e. restrict things by a RegExp? That gives you a more powerful way to describe string limitations, that _includes_ setting the length if you wish - but would of course be a PITA to parse if your only use case actually is limiting the size of some C array

erikbosch:
Some comments on the last comment from Sebastian:

- I agree on that in general instance-definition in standard catalog does not make sense as they often need to be changed to fit number of doors/seats/axles, at least if we intend to have them optimized for individual vehicles. On the other hand, we need some instance-information to be able to generate yaml/json/csv in the same way as we do today, and as they are part of release we need a "default instantiation"
- Even if we in general not will need to specify `stringlength` or `arraysize` in standard catalog, I would rather like them to be described as "rarely" rather than "no", as there might be special cases like `VIN` and possibly timestamps in standardized format where a special number of characters are expected

But I like the idea of having a table of keywords and state how they are used/checked in different contexts. For the proposed `stringlength` keyword I can think of the following behavior in VSS-tools:

- VSS-tools shall give an error if `stringlength` is used on something not a string
- VSS-tools shall for Yaml/JSOn exporter just print the keyword as is, for CSV add in new column
- Behavior for other exporters needs to be defined/investigated - can it be supported? Silently ignored or warning, or even error?

erikbosch:
Meeting: Please review/discuss until next meeting

stringlength.md
```diff
@@ -0,0 +1,112 @@
...
+the existing `arraysize` keyword sets the number of strings in the array.
+
+
+## Examples from VSS
+
+An example implementation might look like:
+```
+VehicleIdentification.VIN:
+  datatype: string
+  stringlength: 17
```
kkoppolu1, line 40: Strings (or collections) have two properties:
1. capacity - Maximum allowed size
2. length - Occupied size

Taking the example of C++ basic string, these are modeled as 
max_size - https://en.cppreference.com/w/cpp/string/basic_string/max_size
size/length - https://en.cppreference.com/w/cpp/string/basic_string/size

Size/length seem to be implementation detail as they may vary at the data sample level.

I assume what we are trying to model here is the capacity/max size. Is that correct?

neil-rti:
That is the intent, and note that I'm looking at this through the lens of
working with DDS, and with constrained or critical systems where byte
allocations need to be accounted for.   There may be other use cases that
need both numbers.
This is a good clarifying question.  Thanks for bringing it up.

erikbosch:
Meeting notes:

- Important to clarify if it concerns max capacity, implementation may use flexible string
- Some exporters may not support it.
- Neil to start work on updating documentation and updating tools.
- Keep this PR open as reference and for discussion while developing

stringlength.md
```diff
@@ -0,0 +1,112 @@
+# Proposal to add Suggested Lengths for String Signals in the VSS
+
+This document covers a concept proposal to add to the VSS .vspec files an additional keyword 
+(`stringsize:` or similar) to assist the code generators (in vss-tools) in setting an appropriate
+default size for strings.
+
+This new keyword would be added to the .vspec files for all `datatype: string` elements (approx. 85 places),
```
SebastianSchildt, line 7: I think this is a useful keyword, but only if it is optional: I.e. I would set a stringsize in standard catalogue only, when it makes semantic sense (e.g. DTCs, where we know they look like X0000).
In case of allowed values the stringlenth is implicitly clear

There are however places where you want to model an unbounded string (Of course your implementation/generator might need to enforce bounds but that is either deployment specific configuration, or implicit in the  tooling). One example: If you had a model defining a Drivers's name. Obviously you implementation might not accept a 1MiB string, but from a data modelling perpsective it does not make sense to limit to something on the base VSS level.

tldr.: Like the keyword, if it is optional (Tooling/exporters that requires it might just substitute a default value for unbounded strings, and recommend you adding a  specific overlay)

erikbosch:
I think we should make a decision on how to progress with this one, so that development can start.

- Decide if we shall introduce a keyword `stringlength` that specifies **max** size of strings for this signal. (My understanding is that it is max size, not fixed size)
- Agree that if not given we anyway have a limit, implicitly or explicitly set by implementation/deployment.
- Agree if length shall exclude trailing null (if any)
- Verify that we have common understanding of `arraysize` - current documentation is a bit ambiguous - does it refer to fix size or max size?

If we agree on introducing it, we need to align on "minimal handling" in vss-tools. Points to agree on:

- Shall it be included as new column in csv.

erikbosch:
Meeting notes:

- Erik: propose to make decision next week
- Sebastian: Where to use? DBC error codes, any more?. 
- Erik: Some examples- VIN, ACRISS, Time in UTC format (as string)
- Sebastian: Good if we can mention examples.

adobekan:
6.6.23
@adobekan  to check details.

stringlength.md
adobekan: I would see it as well like that. But  maybe @neil-rti  can comment?

erikbosch:
Discussion:

- Way forward : @neil-rti (or someone else) to create "real" PR. One for VSS - update doc and add stringlength where reasonable, one for vss-tools so it accepts keyword
- Only include it for those exporters where it makes sense (i.e. optional for exporters to take it into account)

erikbosch:
Converted PR to draft as refactoring is required

erikbosch:
Anyone interested in driving the topic of creating PRs (as described above) @neil-rti ?

erikbosch:

Issue #562, opened 2023-03-03
Case-sensitivity of VSS (type)


erikbosch:
I noticed that there is a slight contradiction between this repository and vss-tools.

- In VSS documentation, it is implicitly stated that type is mandatory as it is not marked as optional. We do not explicitly state anything on case sensitivity, but the list of values use lower case.
- In VSS-tools we actually use "branch" if type is not specified, and we accept any case and convert it to wanted case

I do not really see any point in the vss-tools handling, so I would like to know if your share my view that type shall be mandatory and must be lower case. If so, we could do 4 things:

- Improve documentation to make it more explicit
- Change some existing signals that use other case, like the one below with `Actuator`
- Then refactor vss-tools so that it does not add type by default. This is actually already covered by https://github.com/COVESA/vss-tools/pull/252
- Refactor vss-tools so that it handle types as case-sensitive.

```
Charging.IsChargingCableLocked:
  datatype: boolean
  type: Actuator
  description: Is charging cable locked to prevent removal.
  comment: Locking of charging cable can be used to prevent unintentional removing during charging.
```

erikbosch:
Done

PR #569, opened 2023-03-28
Clarify that VSS is case sensitive.

erikbosch:
This is part of #562.
The idea is that when this has been accepted we can refactor vss-tools and remove functionality for case-insensitive handling. (Which by the way is not handled consistently)

For that to work we must make sure we specify type consistently for existing signals

### DIFF
--- a/docs-gen/content/rule_set/basics.md
+++ b/docs-gen/content/rule_set/basics.md
@@ -17,6 +17,11 @@ The raw specification files can, with help of tools in the [vss-tools repository
 be converted to other formats that are more user friendly to read.
 Converted representations are also included as release artifacts for each [VSS release](https://github.com/COVESA/vehicle_signal_specification/releases).
 
+VSS is in itself case sensitive.
+This means that keywords, signal names, types and values normally shall be given with the case specified.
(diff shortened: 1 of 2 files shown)


erikbosch:
Meeting notes:

- Please review
- Possibly also discuss if additional naming rules are need, like names start with capital letter, camel case?
- Shall "DriveR" and "Driver" be considered as different identifiers?

erikbosch:
Following up on the discussion yesterday. We have some guidelines here:

- https://covesa.github.io/vehicle_signal_specification/rule_set/basics/#style-guide
- https://github.com/COVESA/vehicle_signal_specification/blob/master/CONTRIBUTING.md

They recommend camel case starting with capital letter.

We are currently in a process of separating VSS-syntax (and tools) from the VSS standard catalog itself, as the syntax and tooling might be useful also for other trees like Charging infrastructure. As part of that, if we want to introduce stricture rules we need to decide if the just shall be informal rules for the standard VSS catalog, or if they shall be rules always enforced by tooling, or if we even want the rules to be configurable in vss-tools in some way.

erikbosch:
No comments received. We can always improve later. Merging

PR #571, opened 2023-03-29
Add pre-commit hook

erikbosch:
And make it mandatory to pass and fix present errors


.pre-commit-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+        exclude_types: ["csv"]
+    -   id: end-of-file-fixer
+        exclude_types: ["json"]
+    -   id: check-yaml
+    -   id: check-added-large-files
+
+-   repo: https://github.com/pycqa/flake8
```
erikbosch, line 14: This is currently the same config as vss-tools. One could argue that we possibly could clean up as some parts like flake8 is not relevant as we do not have code here. Opinions are welcome

erikbosch:
No comments received, to avoid conflicts merge this one now. I will keep an eye on any regressions in PR builds, if any

PR #573, opened 2023-04-10
Add DMS support

pmundt:
This adds basic support for Driver Monitoring Systems (DMS) to the ADAS spec. These have been derived from the other ADAS features and presently include:
* ``DMS.IsEnabled`` to determine if the system has been enabled or not
* ``DMS.IsWarning`` to determine if the system is in an alert state
* ``DMS.IsError`` to determine if the system is in an error state

``DMS.IsEnabled`` can be used to determine whether driver monitoring signals under Driver are expected to be valid, while ``DMS.IsWarning`` can be used as a basis for interrogating the cause of a specific alert condition from the Driver state.

An additional ``IsHandsOnWheel`` signal, which is optionally fed by the DMS, is also added to the Driver model in order to prepare for UN R79 and LKAS.

Signed-off-by: Paul Mundt <paul.mundt@harman.com>

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -285,5 +285,27 @@ EBA.IsEngaged:
   type: sensor
   description: Indicates if EBA is currently regulating brake pressure. True = Engaged. False = Not Engaged.
 
+#
+# Driver Monitoring System (DMS)
+#
+DMS:
+  type: branch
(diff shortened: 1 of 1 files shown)


ppb2020:
Regarding `DMS.IsError`, presumably there would be an error code reported in this conditions (and possibly for the warning). Should this signal instead provide an error code, where one value is no error?

PR #575, opened 2023-04-11
Update makefile target for protobuf exporter

kkoppolu1:
## Summary
Update makefile target for protobuf exporter

### Testing
```
make protobuf
```

ran successfully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ binary:
 	${TOOLSDIR}/vspec2binary.py --uuid -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).binary
 
 protobuf:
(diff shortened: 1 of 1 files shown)


erikbosch:
Protobuf is currently part of optional builds, so CI passes but build fails.

```
./vss-tools/vspec2protobuf.py  -I ./spec -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).proto
/bin/sh: 1: ./vss-tools/vspec2protobuf.py: not found
```

Most likely reason is that the submodule reference has not been updated. @kkoppolu1 - feel free to add an update of the vss-tools submodule reference in this PR. If you need help I can look at it after the weekend

erikbosch:
This commit was merged as part of #578, so closing this one

PR #576, opened 2023-04-12
Proposal for improved vss-tools export format control

neil-rti:
Proposal to add new keywords to VSS, and format specification documents to vss-tools.
From these, the exporter modules in vss-tools can be modified to take advantage of the new keywords to produce output that conforms to the format specification documents.   First use case is for IDL, but the intent is to benefit other export formats as well.
The specifics of the grouping of every VSS signal is 'to be determined', enabled by the new keywords.
This effort has been separate and in parallel with https://github.com/COVESA/vss-tools/pull/262 -- but working toward a common goal.
(note that some changes in this PR are trivial doc changes; ignore)
Signed-off-by: Neil Puthuff <neil@rti.com>

### DIFF
--- a/EXPORT_HELPERS.md
+++ b/EXPORT_HELPERS.md
@@ -1,9 +1,10 @@
 ## Proposal / Discussion Draft: Consistent & Documented Export Formats
 
-Goal of this proposal is to get consistent and conformant formatted exports of the VSS,
-from VSS-TOOLS or other potential 3rd-party implementations, resulting in a high degree of
-interoperability between implementations using a particular format (such as IDL), and to
-ease the creation of interoperable examples and reference implementations for a given format.
+Goal of this proposal is to produce consistent and conformant formatted exports of the VSS via
+the VSS-TOOLS or other potential 3rd-party implementations, resulting in a high degree of stability and
+interoperability between implementations using a particular format (such as IDL).
+
+This consistency and stability can ease the creation of interoperable examples and reference implementations for a given format.  
 This draft focuses on IDL, but is intended to also maximize interoperability within other export formats.
 
 To enable this capability, the following changes are proposed:
@@ -11,12 +12,14 @@ To enable this capability, the following changes are proposed:
 - Create a document to describe the recommended formatting and structure of the resulting exported (IDL or other) files.
 - Modify the exporters in VSS-TOOLS to conform to the above document.
 - Utilize the `arraysize` and `stringsizemax` to set the dimensions of the exported array/sequence/string types.
-- Introduce 3 new keywords to the VSS: `memberof`, `membergroup`, `isogroup` to be used as:
+- Introduce 3 new keywords to the VSS: `memberof`, `membergroup`, `isogroup`, used as:
     - To have a `branch` entity create a namespace in the export (named for itself), add:
         - `memberof: namespace`
     - To create named group containers in the export for elements, add to the branch entry:
-        - `membergroup <list of names>` to create containers that may be referenced in the parent of this branch.
+        - `membergroup <list of names>` to create containers that may also be referenced in the parent of this branch.
         - `isogroup <list of names>` to create containers that will not be referenced in the parent of this branch.
+    - To assign `membergroup` struct names to be referenced in a parent struct:
+        - `memberof: <list of parent struct names to insert reference into>`
     - To assign VSS elements to specific named group containers in the export, add to element entry:
         - `memberof: <list of group containers>`
 
@@ -27,7 +30,7 @@ information to consistently produce IDL that conforms to the above document.
 
 Examples of this export guidance on IDL includes:  
 
-**Namespaces**  
+### Namespaces
 In IDL, namespaces are expressed with a `module { ... }` container around the affected definitions.
 A sensible approach to namespacing in VSS would be to use containers for the major sections of
 the specification, such as: `ADAS`, `Body`, `Cabin`, `Chassis`, `OBD`, and `Powertrain`, all residing
@@ -45,7 +48,7 @@ within a top-level container for `Vehicle`.   The resulting IDL would look like:
 Note that only the `branch` entries that have this keyword will receive a namespace for their descendent types.
 All `branch` entries beneath them will be implemented as `struct`.
 
-**Creating structs from VSS elements**  
+### Creating structs from VSS elements
 Using the new keywords to define and assign elements to structs, the VSS can be tailored to match the intended usage
 patterns of the underlying signals.  For example:
 
@@ -119,10 +122,10 @@ such as:
   };
 ```
 
-**Including Descendent Structs in Parent Structs**
+### Including Descendent Structs in Parent Structs
 The VSS structure in some places implies that a struct might be included in its parent struct, 
 such as how `Vehicle.Powertrain.TractionBattery` might want to include its `Temperature` and `StateOfCharge` 
-descendant structs -- but there is no guidance in the VSS to support this grouping.
+descendant structs -- but there is no directive in the VSS to enforce this grouping.
 
 The added keywords could enforce the inclusion of references to descendant types within a parent,
 enabling data structures such as:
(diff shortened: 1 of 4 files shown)


neil-rti:
Also note that the 'vss-dedupe.txt' file is a de-duplicated version of the VSS, wherein any duplicate signal groupings are reduced to just one, with the VSS nodes that use them contained [in a bracketed list].
This reduces the groups of signals from the original IDL export of 850+ struct definitions to just 96 groups -- and it's a convenient way to look at the VSS in a single file.
Included here for reference.

vss_dedupe.txt
```diff
@@ -0,0 +1,572 @@
...
+['Vehicle:Powertrain:Transmission']
+  attribute string Type
+  attribute int8 GearCount
+  attribute string DriveType
+  sensor    float TravelledDistance
+  sensor    int8 CurrentGear
+  sensor    int16 Temperature
+  sensor    uint8 ClutchWear
+  actuator  int8 SelectedGear
+  actuator  boolean IsParkLockEngaged
```
pmundt, line 130: One issue that I see is that the VSS entries are in arbitrary/logical order in order to flow more logically and be more human readable, but given the variability in data types, this is not conducive to optimal struct layout. In this case, there are mixed 8/16/32bit entries, where if we were to take this with a 1:1 mapping, any dereference of DriveType on a non-byte addressable system will generate an alignment trap. The in-memory size of the struct could also change depending on whether the compiler automatically inserts padding bytes or not, which could be an issue if there is an intention to do binary serialization/deserialization of the structure via things like gRPC across systems with different memory models. While this could be mitigated by forcing the compiler to pack the struct and leave it as-is, this would again go back to triggering alignment traps on non-aligned dereferencing.

One possible solution would be to have an intermediate step in which the generated items are passed through an analysis tool that reorders struct elements and adds implicit padding bytes where needed to match alignment needs.

neil-rti: For DDS (at least), the details of byte-addressability, endianness, word size, etc. are handled by the DDS implementation for that platform.  Systems with different endianness/CPU/OS/ProgLang will interoperate without knowing what's on the other end(s) of the DDS connection.
But this is a good point -- the order of appearance in a struct is implied by where the element is placed in the .vspec file, not by something explicit (such as a keyword that states the position of this element in a struct).
Is implicit order good enough?

docs-gen/content/introduction/overview.md
```diff
@@ -40,7 +40,7 @@ The figure below shows an example snapshot of a generated tree of the
 specification. The leafs contain the actual information as shown in the figure.
 Before going into detail of the specification, let's dig deeper into taxonomies.
 
-![Example tree](/vehicle_signal_specification/images/tree.png?classes=shadow&width=60pc)
+![Example tree](https://covesa.github.io/vehicle_signal_specification/images/tree.png?classes=shadow&width=60pc)
```
erikbosch, line 43: I think adding the "full path" shall not be necessary, i.e. this tree is visible today if you look at https://covesa.github.io/vehicle_signal_specification/introduction/overview/ (but there shall be no md as used in the changed like above)

EXPORT_HELPERS.md
```diff
@@ -0,0 +1,215 @@
...
+
+This consistency and stability can ease the creation of interoperable examples and reference implementations for a given format.  
+This draft focuses on IDL, but is intended to also maximize interoperability within other export formats.
+
+To enable this capability, the following changes are proposed:
+
+- Create a document to describe the recommended formatting and structure of the resulting exported (IDL or other) files.
+- Modify the exporters in VSS-TOOLS to conform to the above document.
+- Utilize the `arraysize` and `stringsizemax` to set the dimensions of the exported array/sequence/string types.
+- Introduce 3 new keywords to the VSS: `memberof`, `membergroup`, `isogroup`, used as:
```
erikbosch, line 15: Does this possibly need to be aligned with the VSSo work (FYI: @jdacoello )

SebastianSchildt, line 15: I can see the value of this, for some exporters, but for others it woudl be of no relevance (CSV, JSON), the question is, should this be defined as "Core VSS", or maybe as reserved atrributes in an overlay (and we could still think about shipping/providing one for standard catalogue).

The challenge I see is, if we make the "core" VSS to powerful, basically every exporter/techstack, would feel like "I am VSS compliant except x,y,z is not supported"

EXPORT_HELPERS.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Effect on IDL Export
+Implementing the above keywords into the VSS -- to enforce the guidance of the above reference document -- 
+should have no effect on existing exporters in VSS-TOOLS, however a modified IDL exporter could use this new
+information to consistently produce IDL that conforms to the above document.
+
+Examples of this export guidance on IDL includes:  
+
+### Namespaces
+In IDL, namespaces are expressed with a `module { ... }` container around the affected definitions.
```
SebastianSchildt, line 34: Just a nitpick, but as I assume this refers to DDS/OMG IDL, maybe when referring to it, should be spelt out, because in my view "IDL" is a pretty generic term, that is applicable to many different IDLs?

EXPORT_HELPERS.md
```diff
@@ -0,0 +1,215 @@
...
+    };
+    module Chassis {
+      (struct and enum definitions for Chassis descendants)
+    };
+  };
+```
+Note that only the `branch` entries that have this keyword will receive a namespace for their descendent types.
+All `branch` entries beneath them will be implemented as `struct`.
+
+### Creating structs from VSS elements
```
SebastianSchildt, line 51: Generally, it makes sense to group signals into structs, but how does this "tagging" approach later merges/matches with "real structs" as we recently defined

neil-rti: However, I agree that if both PR's are implemented, it could represent 2 different ways of specifying a struct:
269 enforces using a struct (new signals only)
576 suggests a struct and is optional (all signals)

petervolvowinz:
So, we believe that adding this to VSS is not the correct way. Adding support for IDL should perhaps better fit on top of VSS as @erikbosch suggested perhaps as part of VSSo. Keeping VSS simple is key to its continued success and adaption.

docs-gen/content/introduction/overview.md
erikbosch: A side note - after working a bit with the documentation for VSS 4.0. Paths like "/vehicle_signal_specification/images/tree.png" reflects location after Hugo-generation, when for instance looking at https://covesa.github.io/vehicle_signal_specification/introduction/overview/. They do not work out of the box if looking at github source at https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/introduction/overview.md.

Changing to full link like here makes it work for both raw markdown and generated Hugo, but would give problems if we later would like to publish multiple versions of documentation (like 4.0, 3.1 and so on) in parallel, as it then will point to the same location for all versions. There might be some alternative ways to solve this, but it does not seem to be that easy.

erikbosch:
Some ideas on how to go forward.

**First phase: Optional DDSIDL feature** 

Improve vspec2ddsidl so that it can read (and take advantage of) additional metadata. I.e. define and document a mapping format (similar to [KUKSA.val dbcfeeder](https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/mapping/mapping.md)) and implement support for it in vspec2ddsidl. Initially aim to handle the extra information by overlays, and make the extra information optional, so that there is a default conversion if no overlay with extra information is used.

An (example) overlay for a subset of VSS signals may be created as part of vspec2ddsidl.

As I see it this phase can be handled completely in VSS-tools.

**Second phase: Investigate if it can be reused for other tools and if mapping can be standardized**

When we have an initial solution for vspec2ddsidl we can investigate if:

- Check if the same syntax/metadata can be valuable for other tools. If so which ones? Possibly useful for VSS->VSSo conversion?
- Check if actual mapping/grouping could be standardized? If all or parts of the mapping would be the same regardless of use-case scenario then we could consider standardizing (parts) of the extra definitions. Either as an overlay or directly in source tree.

Assuming for instance that we want to use an extra keyword to specify grouping for `Mirrors.Pan` like below - if all agree that this is the "standardized COVESA grouping for DDSIDL" I see no real problem if it is included in the source *.vspec files. That could simplify maintenance and as long as it is optional should not prevent addition of other signals. But to be able to make a decision on this I think we need to have tool support and an example overlay covering a subset of VSS in place.

```
Mirrors.Pan:
  datatype: int8
  type: actuator
  ddsidl:
    memberof: MirrorPosition
```

erikbosch:
Meeting notes:

- Peter - think that it be difficult to agree on mapping, think it should be on top
- Meeting conclusion: Implementing it first for VSPEC2DDSIDL seems to be a reasonable approach.
-

neil-rti:
Meeting Notes:

- Detracts from the simplicity of VSS as a whole
- If this is to be implemented, should be done in an overlay (and be specific to those exporters that might benefit from the groupings).

I tend to agree; will close this PR and investigate an implementation using overlays.  Great discussion, many thanks.

PR #577, opened 2023-04-17
Added IsScreenAlwaysOn, LastActionTime and DisplayOffTime Signals

UmangSharmaMobis:
Added IsScreenAlwaysOn, LastActionTime and DisplayOffTime Signals based on the discussion

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -189,6 +189,21 @@ HMI.DayNightMode:
   allowed: ['DAY', 'NIGHT']
   description: Current display theme
 
+HMI.IsScreenAlwaysOn:
+  datatype: boolean
+  type: actuator
+  description: used to prevent the screen going black if no action placed
+
+HMI.LastActionTime:
+  datatype: string
+  type: sensor
+  description: Time for last hmi action, formatted according to ISO 8601 with UTC time zone.
+
+HMI.DisplayOffTime:
+  datatype: string
+  type: actuator
+  description: Time used to make display off
+
 SmartphoneProjection:
   type: branch
   description: All smartphone projection actions.
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -192,7 +192,7 @@ HMI.DayNightMode:
 HMI.IsScreenAlwaysOn:
   datatype: boolean
   type: actuator
-  description: used to prevent the screen going black if no action placed
+  description: Used to prevent the screen going black if no action placed.
 
 HMI.LastActionTime:
   datatype: string
@@ -202,7 +202,7 @@ HMI.LastActionTime:
 HMI.DisplayOffTime:
   datatype: string
   type: actuator
-  description: Time used to make display off
+  description: Time used to make display off.
 
 SmartphoneProjection:
   type: branch


erikbosch:
A signoff is needed in the commit to state that the contribution is according to [COVESA contribution guidelines](https://www.covesa.global/contribute), that is the reason why one of the checks fails. If using git command line something like below should make the trick to sign and force-push the existing commit.

```
git commit --amend -s
git push --force ...
```

spec/Cabin/Infotainment.vspec
```diff
@@ -189,6 +189,21 @@ HMI.DayNightMode:
   allowed: ['DAY', 'NIGHT']
   description: Current display theme
 
+HMI.IsScreenAlwaysOn:
+  datatype: boolean
+  type: actuator
+  description: used to prevent the screen going black if no action placed
```
erikbosch, line 195: I think description is ok, but please change case to `Used...` and add a dot/period at end.

spec/Cabin/Infotainment.vspec
```diff
@@ -189,6 +189,21 @@ HMI.DayNightMode:
...
+
+HMI.LastActionTime:
+  datatype: string
+  type: sensor
+  description: Time for last hmi action, formatted according to ISO 8601 with UTC time zone.
+
+HMI.DisplayOffTime:
+  datatype: string
+  type: actuator
+  description: Time used to make display off
```
erikbosch, line 205: Do we want this signal to be an absoluate time or rather as a duration (in seconds?) relative to `HMI.LastActionTime`?

spec/Cabin/Infotainment.vspec
UmangSharmaMobis: I have updated the code.

spec/Cabin/Infotainment.vspec
```diff
@@ -189,6 +189,21 @@ HMI.DayNightMode:
...
+
+HMI.LastActionTime:
+  datatype: string
+  type: sensor
+  description: Time for last hmi action, formatted according to ISO 8601 with UTC time zone.
+
+HMI.DisplayOffTime:
+  datatype: string
+  type: actuator
+  description: Time used to make display off.
```
erikbosch, line 205: In a previous comment (no longer visible becuase of later commits) Umang said:

_we can have time as absolute time, DisplayOffTime will have the time specification value. so HMI.LastActionTime+ Displayofftime = time (when display will go to off state)._ 

I have no strong opinion on what fits best for the use case, but I think the signal could benefit from some clarification. Either, if we want absolute time, like the display shall go off at 2023-04-18 14.00 UTC then we can possibly rephrase to something like:

```
HMI.DisplayOffTime:
  datatype: string
  type: actuator
  description: Time when display shall be turned off, formatted according to ISO 8601 with UTC time zone.
  comment: If DisplayOffTime is in the past and IsScreenAlwaysOn==False then display shall be off
```

But maybe it is more practical to have it as duration, possibly like:

```
HMI.DisplayOffDuration:
  datatype: uint16
  type: actuator
  unit: seconds
  description: Duration in seconds before the display is turned off. Value shall be 0 if screen never shall turn off.
  comment: Display shall be turned off at HMI.LastActionTime + HMI.DisplayOffDuration, unless HMI.IsScreenAlwaysOn==True
```

Let us bring it up to get feedback at the meeting today.

erikbosch:
The failed build is caused by the DCO check as new commits have been added, the old one without DCO still remains. It is not an urgent problem to fix, but before merging you should preferably squash all commits to a single one and make sure that it includes a sign-off

UmangSharmaMobis:
Creating New Pull request of same changes.

PR #583, opened 2023-04-20
Add missing units to various signals

nickrbb:
This PR addresses the following issues related to the Units of some signals:

1. Vehicle.Driver.HeartRate is missing a defined Unit.
 - Suggested fix: Set Unit to “bpm” (beats per minute) and add to Units config file.

2. Vehicle.Cabin.Sunroof.Position is missing a defined Unit.
 - Suggested fix: Set Unit to “percent”.

3. Vehicle.Body.Windshield.Front.Wiping.WiperWear and Vehicle.Body.Windshield.Rear.Wiping.WiperWear are missing a Unit.
 - Suggested fix: Set Unit to “percent”.

4. Vehicle.Body.Windshield.Front.Wiping.System.Frequency and Vehicle.Body.Windshield.Rear.Wiping.System.Frequency missing Unit. Description states the value is “cycles per minute”.
 - Suggested fix: Set Unit to “cpm” (cycles per minute) and add to Units config file.

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -130,6 +130,7 @@ Windshield.Wiping.System:
 Windshield.Wiping.WiperWear:
   datatype: uint8
   type: sensor
+  unit: percent
   max: 100
   description: Wiper wear as percent.
                0 = No Wear.
--- a/spec/Body/WiperSystem.vspec
+++ b/spec/Body/WiperSystem.vspec
@@ -56,6 +56,7 @@ Mode:
 
 Frequency:
   datatype: uint8
+  unit: cpm
   type: actuator
   description: Wiping frequency/speed, measured in cycles per minute.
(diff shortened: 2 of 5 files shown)


erikbosch:
Signoff needed here as well, apart from that I think it looks good

nickrbb:
OK, this commit message should now contain a sign-off too. Hopefully all good now?

erikbosch:
> OK, this commit message should now contain a sign-off too. Hopefully all good now?

Unfortunately the check still does not pass.

A theory is that is requires `<` and `>` around the e-mail address, like below:

`Signed-off-by: Erik Jaegervall <erik.jaegervall@se.bosch.com>`

erikbosch:
Meeting notes:

- OK to merge
- Sebastian: Possibly we should discuss to add a unit check, but then we need a way to specify that there is no need.

erikbosch (approved):

PR #589, opened 2023-05-03
Removal of deprecated signals from VSS v3.1

nickrbb:
Two signals were listed as deprecated in VSS v3.1: Vehicle.TravelledDistance and Vehicle.Powertrain.FuelSystem.TimeSinceStart. Therefore, since the current (in-development) version is a major release (v4.0), these two signals should not appear.


erikbosch:
Thanks for the PR. We have removal of deprecated signals as a step in the release preparations (see https://github.com/COVESA/vehicle_signal_specification/wiki/Release-Instructions-and-Checklist), but there is nothing that prevents the from being removed as soon as we have decided that next release on "master" will be a major release.

erikbosch (approved):

PR #590, opened 2023-05-03
Miscellaneous editorial only corrections

nickrbb:
Editorial-only corrections to text in Description and Comment fields of various signals and branches. Summary:
- Change of "rainsensor" into two words (no such word as "rainsensor")
- Change of "Rearview mirror" to "Rear-view mirror" (no such word as "rearview", needs hyphenating as it's a compound adjective)
- Change of "Wheel base" to "Wheelbase"
- Correction to spelling of "ceiling" (was "cealing" in some instances)
- Removal of some instances of duplicate words
- Overhaul and alignment of Description texts for DistractionLevel and FatigueLevel
- Numerous corrections to case of letters and fixes to instances of "e.g." and "i.e."

Nick Russell <79584907+nickrbb@users.noreply.github.com>

### DIFF
--- a/spec/Body/Body.vspec
+++ b/spec/Body/Body.vspec
@@ -84,7 +84,7 @@ Horn.IsActive:
 #
 Raindetection:
   type: branch
-  description: Rainsensor signals.
+  description: Rain sensor signals.
 
 Raindetection.Intensity:
   datatype: uint8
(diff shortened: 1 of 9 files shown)


erikbosch (approved):
Thanks for the PR. I think this is as you state it editorial changes that are worthwhile to include in the upcoming 4.0 and no need to announce in a meeting to ask for reviews

PR #601, opened 2023-05-08
Adding additional signals to TractionBattery branch

erikbosch:
This PR includes a few possible improvements to the TractionBattery branch.
I could give an introduction to the wanted changes in an upcoming VSS meeting

### DIFF
--- a/spec/Powertrain/TractionBattery.vspec
+++ b/spec/Powertrain/TractionBattery.vspec
@@ -24,6 +24,13 @@ ProductionDate:
   type: attribute
   description: Production date of battery in ISO8601 format, e.g. YYYY-MM-DD.
 
+ErrorCodes:
+  datatype: string[]
+  type: sensor
+  description: Current error codes related to the battery, if any.
+  comment:     Error code format is not defined, it may be DTCs according to OBD II (SAE-J2012DA_201812)
+               standard ([P|C|B|U]XXXXX ) or any other format.
+
 IsPowerConnected:
   datatype: boolean
   type: sensor
@@ -60,6 +67,12 @@ Temperature.Max:
   unit: celsius
   description: Current maximum temperature of the battery cells, i.e. temperature of the hottest cell.
 
+Temperature.CellTemperature:
+  datatype: float[]
+  type: sensor
+  description: Array of cell temperatures. Length or array shall correspond to number of cells in vehicle.
+  comment: Cells are identified by relative position in array.
+
 CellVoltage:
   type: branch
   description: Voltage information for cells in the battery pack.
@@ -70,11 +83,29 @@ CellVoltage.Min:
   unit: V
   description: Current voltage of the battery cell with lowest voltage.
 
+CellVoltage.IdMin:
+  datatype: uint16
+  type: sensor
+  description: Identifier of the battery cell with lowest voltage.
+  comment: Identifier is supposed to be relative index of the cell, starting with 0 the first cell.
+
 CellVoltage.Max:
   datatype: float
   type: sensor
   unit: V
   description: Current voltage of the battery cell with highest voltage.
+
+CellVoltage.IdMax:
+  datatype: uint16
+  type: sensor
+  description: Identifier of the battery cell with highest voltage.
+  comment: Identifier is supposed to be relative index of the cell, starting with 0 the first cell.
+
+CellVoltage.CellVoltages:
+  datatype: float[]
+  type: sensor
+  description: Array of cell voltages. Length or array shall correspond to number of cells in vehicle.
+  comment: Cells are identified by relative position in array.
 
 GrossCapacity:
   datatype: uint16
@@ -87,7 +118,6 @@ NetCapacity:
   type: sensor
   unit: kWh
   description: Total net capacity of the battery considering aging.
-
 
 StateOfHealth:
   datatype: float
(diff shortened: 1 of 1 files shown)


erikbosch:
Meeting notes: 

- Please review
- Nick: Is V2X is a good fit for Charging.Mode, is discharge, not charge. V2X may not be a good name. Possible GRID.Charge/GRID.Discharge. Should we add a value for "neither charge nor discharge"

spec/Powertrain/Battery.vspec
```diff
@@ -290,12 +320,13 @@ Charging.ChargePlugType:
 Charging.Mode:
   datatype: string
   type: actuator
-  allowed: ['MANUAL', 'TIMER', 'GRID', 'PROFILE']
+  allowed: ['MANUAL', 'TIMER', 'GRID', 'PROFILE', 'V2X']
```
erikbosch, line 323: There were a short discussion on this signal last week - if V2X as name is misleading, if we need an IDLE state, if we need to split it into charge/discharge mode signals... That made me think - how is this signal intended to be used and does the current state reflect that. I am no expert in this area so feedback is appreciated. If we are to capture "modes" from a vehicle perspective, does the current "state space" reflects that? Do we possible need other states and a table, possibly like below.

I.e. That you in the vehicle (by settings) can control what shall happen if a charging cable is connected, and if the vehicle only will accept charge or also discharge? But do we then need to separate "GRID" from other charge/discharge modes? I.e. will the vehicle behave differently if you connect a "less intelligent load" like your boat or off-grid cottage or if you connect to a "smart Grid"? Is the difference possibly that in "manual/triggered discharge" we accept any load, but in grid-mode the "load/charger" must identify itself towards the vehicle with some trusted credentials.

Mode | What happens when cable is connected? | When will charging start? | When will discharge start?
--- |   --- | --- | ---
IDLE | Nothing	 | Never	 | Never
MANUAL	 | Charging starts	 | Cable connected 	 | Never
TRIGGERED	 | User dialog appears	 | When acknowledged by user	 | Never
TIMER	 | Depends on timer	 | Depends on timer	 | Never
GRID	 | Controlled by grid if authorized	 | Controlled by grid if authorized	 | Controlled by grid if authorized	
PROFILE	 | Depends on profile	 | Depends on profile	 | Never
MANUAL_CHARGE_DISCHARGE	 | Charge or Discharge starts	 | Cable connected	 | Cable connected
TRIGGERED_CHARGE_DISCHARGE	 | User dialog appears	 | When acknowledged by user	 | When acknowledged by user

One could off course also think off other combinations, like charging may start as soon as a charge is connected, but for discharge the driver needs to acknowledge it. Then having separate signals would be a benefit, potentially with different states.

If this discussion take too long time we could possibly handle it separately from the other changes.

SebastianSchildt: nitpick: "MANUAL_CHARGE_DISCHARGE" is not clearly defined, so WHAT does happen when a cable is connected? Also generally not sure if "manual" is the correct term indicating a mode where connecting a cable AUTOMATICALLY does something

erikbosch: > nitpick: "MANUAL_CHARGE_DISCHARGE" is not clearly defined, so WHAT does happen when a cable is connected? Also generally not sure if "manual" is the correct term indicating a mode where connecting a cable AUTOMATICALLY does something

AUTOMATIC_CHARGE_DISCHARGE is maybe a better term. My assumption here is that the vehicle charge port here would act similar to a USB-port on a laptop. If you connect a charger the laptop will start charging. If you connect your phone (i.e. a load) the laptop will start charging your phone. I.e. the item you connect decide if it will charge or discharge the laptop. But I have no clue if that is how it will work also for a vehicle.

spec/Powertrain/Battery.vspec
```diff
@@ -290,13 +320,18 @@ Charging.ChargePlugType:
...
   datatype: string
   type: actuator
-  allowed: ['MANUAL', 'TIMER', 'GRID', 'PROFILE']
-  description: Control of the charge process.
-               MANUAL means manually initiated (plug-in event, companion app, etc).
-               TIMER means timer-based.
-               GRID means grid-controlled (e.g. ISO 15118).
-               PROFILE means controlled by profile download to vehicle.
-  comment:     The mechanism to provide a profile to the vehicle is currently not covered by VSS.
+  allowed: ['DEACTIVATED', 'AUTOMATIC', 'TRIGGERED', 'TIMER', 'PROFILE', 'V2X']
```
erikbosch, line 323: Proposal refactored after internal discussion. For now we assume GRID can be considered as a special case for V2X, where you also have cases where you charge "stupid loads" by the vehicle, like charging your electrical chain saw. In both cases the vehicle is willing to provide energy to the load. It is assumed that the vehicle will have some internal limitations but apart from that it will try to give as much power as requested by the connected entity.

spec/Powertrain/Battery.vspec
```diff
@@ -70,12 +83,30 @@ CellVoltage.Min:
...
   unit: V
   description: Current voltage of the battery cell with highest voltage.
 
+CellVoltage.IdMax:
+  datatype: uint16
+  type: sensor
+  description: Identifier of the battery cell with highest voltage.
+  comment: Identifier is supposed to be relative index of the cell, starting with 0 the first cell.
+
+CellVoltage.CellVoltage:
```
erikbosch, line 104: This seems to cause a problem for ddsidl - we get a struct CellVoltage within a module CellVoltage which causes problem for idlc.

To me this seems like a bug in ddsidl - the VSS is correct, ddsidl needs to handle it. Off course we could avoid the problem by selecting a different name, but that is not the long term solution as I see it.

SebastianSchildt:
Meeting 05/23

 - V2X maybe still not a good term (it is more associated with communication)
    - Need a different term maybe: EXTERNAL_ENTITY.
 - `CellVoltage.CellVoltage`: It does not seem "good style" to have a signal the same name as a branch, if it can be avoided
 - Should the signal maybe be `CellVoltageS`(plural, we did the some Error)

erikbosch:
Meeting notes: PR updated, please review

nickrbb:
In Charging.Mode, the values "MANUAL" and "GRID" are no longer allowed values as they were in v4.0 of this signal. Is their removal a backwards incompatible change?

If so, should we include them in 4.1 but state in the Deprecated field that these two values will be removed from version 5.0 and to use "TRIGGERED" and "EXTERNAL_ENTITY" instead?

spec/Powertrain/TractionBattery.vspec
```diff
@@ -296,13 +326,21 @@ Charging.ChargePlugType:
...
+               PROFILE means charging is controlled by profile downloaded to vehicle.
+               EXTERNAL_ENTITY means charging/discharging is controlled by the external entity connected to the vehicle.
+               This includes GRID-controlled charging (e.g. ISO 15118), but also other cases where vehicle is
+               connected to an arbitrary load that is powered by the vehicle.
                MANUAL means manually initiated (plug-in event, companion app, etc).
-               TIMER means timer-based.
                GRID means grid-controlled (e.g. ISO 15118).
-               PROFILE means controlled by profile download to vehicle.
-  comment:     The mechanism to provide a profile to the vehicle is currently not covered by VSS.
+  deprecation: V4.1 - MANUAL and GRID are deprecated, please use AUTOMATIC/TRIGGERED or EXTERNAL_ENITY instead.
```
erikbosch, line 341: Added MANUAL+GRID plus this deprecation text. Note that we here use the `deprecation` keyword in a somewhat new meaning as the signal itself is not deprecated but rather certain parts of it. We could alternatively just mention deprecation in free text

erikbosch:
Meeting notes: Merge

PR #604, opened 2023-05-10
Improving documentation on how to use Makefile

erikbosch:
Current status:

- In this repo we do not mention how to use the Makefile
- We also do not mention what dependencies that are needed to be able to run `make`

(Most information is available in vss-tools repo, but not visible here)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,80 @@
+# Building and Using the VSS Standard Catalog
+
+As a VSS user you have two options - either your project directly consumes the source files (`*.vspec`) in this repository,
+or you first convert it to some other format using [VSS Tools](https://github.com/COVESA/vss-tools).
+
+## Using the VSS source files
+
+If you intend to write your own VSS parser there are two files that you need to use as input.
+The root source file for the VSS standard catalog is [VehicleSignalSpecification.vspec](spec/VehicleSignalSpecification.vspec).
+It include all other `*.vspec` files in this repository (except overlays/profiles).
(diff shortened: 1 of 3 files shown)


SebastianSchildt:
Ok on content, Not sure "CONTIBUTING.md"  is the place I would normally expect "technical" documentation. I would put it in Readme or BUILD.md or something

Also, I think in the long run there should be no Makefile here, and this should just be the "catalogue", where any instructions to build should reference to vss-tools, or be a two-liner "pip install some tools" "vpsec2x xxx" . But as we are not there yet, maybe in the meantime improve what is here

erikbosch:
Meeting notes: 

- Please review
- Sebastian: Maybe do not have all documentation in CONTRIBUTING

erikbosch:
> Also, I think in the long run there should be no Makefile here, and this should just be the "catalogue", where any instructions to build should reference to vss-tools, or be a two-liner "pip install some tools" "vpsec2x xxx" . But as we are not there yet, maybe in the meantime improve what is here

In some way VSS (i.e. this repo) needs to specify how the standard catalogue shall be verified/tested in Continuous Integration, and as long as we create artifacts as part of every release it must in some way be specified how they are created. Today this is done by [wiki](https://github.com/COVESA/vehicle_signal_specification/wiki/Release-Instructions-and-Checklist) plus Makefile. VSS-tools as an upstream repo does not know how we in this repo wants to use VSS-tools. But in one way you are right, we should try to limit tooling-related topics here as much as possible. I will see how I can refactor this PR

PR #609, opened 2023-05-17
Clarify that root for types must be a branch

erikbosch:
Related to https://github.com/COVESA/vss-tools/pull/282


erikbosch:
Fixed in 4.0, but shall better be discussed at meeting to agree if any . Two things to discuss:

- OK limitation that root must be a branch? (If not some rework needed in e.g. protobuf tool)
- OK limitation to only have one (1) type root? The drawback is if you need to integrate "private" signals from multiple suppliers, then they might use different name for the type root. If we keep limitation, is there a need to recommend a root name.

PR #610, opened 2023-05-17
Introduce Time Remaining Signals

JonathanForce:
### Description
It would be interesting to know the remaining time of use of a vehicle according to its energy source.

### On Which vehicles ?
Any constructors could be interested by this signals. We think that they are generic enough to be part of the VSS tree.

I will be there next Tuesday for the VSS meeting. We can talk about it.

### DIFF
--- a/spec/Powertrain/FuelSystem.vspec
+++ b/spec/Powertrain/FuelSystem.vspec
@@ -60,6 +60,12 @@ Range:
   unit: m
   description: Remaining range in meters using only liquid fuel.
 
+TimeRemaining:
+  datatype: uint32
+  type: sensor
+  unit: s
+  description: Time remaining in seconds before the fuel tank is empty.
+
 InstantConsumption:
   datatype: float
   type: sensor
--- a/spec/Powertrain/Powertrain.vspec
+++ b/spec/Powertrain/Powertrain.vspec
@@ -15,6 +15,12 @@ Range:
   unit: m
   description: Remaining range in meters using all energy sources available in the vehicle.
 
+TimeRemaining:
+  datatype: uint32
+  type: sensor
+  unit: s
+  description: Time remaining in seconds before all energy sources available in the vehicle are empty.
+
 Type:
   datatype: string
   type: attribute
--- a/spec/Powertrain/TractionBattery.vspec
+++ b/spec/Powertrain/TractionBattery.vspec
@@ -196,6 +196,12 @@ Range:
   unit: m
   description: Remaining range in meters using only battery.
 
+TimeRemaining:
+  datatype: uint32
+  type: sensor
+  unit: s
+  description: Time remaining in seconds before the battery is empty.
+
 Charging:
   type: branch
   description: Properties related to battery charging.


SebastianSchildt:
Thanks for the PR, and looking forward to seeing you in the meeting.

I was wondering, where exactly those information is used/presented to the user (stalked you a bit and think I have seen things like snow mobiles, jet skis.. Unfortunately I have no hands one experience with that :)). Obviously time remaining depends on engine load, so is this meant to be at "full throttle", or rather some instantaneous thing as in "if you keep going exactly like this you have xxx time left"?)

JonathanForce:
Thanks for reply,

With our vehicles the customers don't do distance. They go to a lake, a track (for SSV or 4W) and they drive around for a whole afternoon, so distance doesn't mean anything in this context.

spec/Powertrain/Battery.vspec
```diff
@@ -196,6 +196,12 @@ Range:
   unit: m
   description: Remaining range in meters using only battery.
 
+Time:
```
erikbosch, line 199: I like the new signals - but should they possibly be called `TimeRemaining` or similar? If we just call them `Time` it might be difficult to know what they are intended for if just looking at the name.

JonathanForce: Yes, we can call it TimeRemaining.
I wanted to apply the same logic as for the "range".

Range:
  datatype: uint32
  type: sensor
  unit: m
  description: Remaining range in meters using only battery.

But then, shouldn't Range be RangeRemaining or RemainingRange ? Because Range could be the maximum distance you can travel with a full battery and RemainingRange the distance you can travel with the current charge level. I say this in order to keep consistency, this would be a change to be made in another branch of course.

What do you think about this?

SebastianSchildt:
Discussion 
 - can be confusing as it depends on "driving style"/environment (range as well as time)
 - is "time" or "Time remaining" is better?
    - "Remaining helps"
    - not consistent with Range. Change both?
 - should range/time not be grouped? Only need one subscription possibly?
 
Please review/gives comments.

JonathanForce:
> Discussion
> 
> * can be confusing as it depends on "driving style"/environment (range as well as time)
> * is "time" or "Time remaining" is better?
>   
>   * "Remaining helps"
>   * not consistent with Range. Change both?
> * should range/time not be grouped? Only need one subscription possibly?
> 
> Please review/gives comments.

I start :-)

- It will be up to the implementer to provide reliable data according to the different driving parameters. In the same way as the range finally.
- TimeRemaining seems to me more appropriate. 
- I will not change the range signal for the moment.
- As for grouping them, we would have to see what this could do and what it would mean for the subscription.

SebastianSchildt:
Generally I'd also lean towards calling this TimeRemaining , and maybe in this PR not touching Range ( I do think we should have a discussion whether we want to name it RangeRemaining for consistency reasons, but I think this can be a seperate discussion PR)

erikbosch:
Meeting notes:

- Jonathan to update signals to TimeRemaining
- Then ok to merge

erikbosch:
@JonathanForce - there is a conflict, most likely because the file Battery.vspec has been renamed to TractionBattery.vspec. Could you possible rebase on latest master and do a force push. Then i can take a quick look and if it looks ok I will merge

JonathanForce:
@erikbosch Thank you, i did the rebase :-)

erikbosch (approved):

PR #619, opened 2023-06-06
Identification extension (LicensePlate and VehicleExteriorColor)

jakubdrabik:
During our adoption of VSS we noticed missing license plate number and vehicle exterior color. 

VehicleExteriorColor could further be accompanied by VehicleExteriorColorSpecification with more details.

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -101,6 +101,12 @@ VehicleIdentification.DateVehicleFirstRegistered:
   type: attribute
   description: The date in ISO 8601 format of the first registration of the vehicle with the respective public authorities.
 
+VehicleIdentification.LicensePlate:
+  datatype: string
+  type: attribute
+  description: The license plate of the vehicle.
+  comment: Depending on the context, this attribute might not be up to date or might be misconfigured, and therefore should be considered untrustworthy in the absence of another method of verification.
+
 VehicleIdentification.MeetsEmissionStandard:
   datatype: string
   type: attribute
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -142,6 +142,11 @@ VehicleIdentification.VehicleSpecialUsage:
   type: attribute
   description: Indicates whether the vehicle has been used for special purposes, like commercial rental, driving school.
 
+VehicleIdentification.VehicleExteriorColor:
+  datatype: string
+  type: attribute
+  description: The main color of the exterior within the basic color palette (eg. red, blue, black, white, ...).
+
 VehicleIdentification.VehicleInteriorColor:
   datatype: string
   type: attribute


spec/Vehicle/Vehicle.vspec
```diff
@@ -136,6 +141,11 @@ VehicleIdentification.VehicleSpecialUsage:
   type: attribute
   description: Indicates whether the vehicle has been used for special purposes, like commercial rental, driving school.
 
+VehicleIdentification.VehicleExteriorColor:
```
erikbosch, line 144: Do we need to more explicit on what we expect this signal to contain, for example stating that it shall be the OEM name of the color, rather than color code, and that the names used are not standardized. For example, a possible name could be  "Carnelian red metallic".

erikbosch: I think it makes sense

jakubdrabik: I'll provide a better description of the property

erikbosch:
Meeting notes:

- Shall we possibly also add OEM marketing name and OEM color code?
- Adnan: Would be good to have all
- Jakub: Better to keep it as is, add the other later.
- Is there is a risk to have the license plate signal configurable by the user - could we add a comment stating the risk.
- Jakub - I could mention the risk
- Erik - Maybe something for me and Adnan to discuss in Data expert Group - how is various attributes handled, are they defined by OEM, vehicle or owner, ....

erikbosch:
Concerning colors there were a meeting chat message asking if we possibly should use colors from https://www.farb-tabelle.de/de/farbtabelle.htm.

My personal view is that using those standardized colors does not make sense, as they neither correspond to the colors supported by the vehicle registry, not the OEM marketing name of the color. The Hex/RGB code could possibly be interesting to represent, but I do not know if it would be sufficient for a garage to do a good paint job.

By the way, I found the 10 colors used in the german vehicle registry. So a bit fewer than the 24 we have in Sweden:

Es stehen insgesamt 10 Farbcodes zur Verfügung, die folgenden Farben zugeordnet werden:

0 = Weiß 1 = Gelb 2 = Orange 3 = Rot 4 = Lila bzw. Violett 5 = Blau 6 = Grün 7 = Grau 8 = Braun 9 = Schwarz

jakubdrabik:
> Concerning colors there were a meeting chat message asking if we possibly should use colors from https://www.farb-tabelle.de/de/farbtabelle.htm.
> 
> My personal view is that using those standardized colors does not make sense, as they neither correspond to the colors supported by the vehicle registry, not the OEM marketing name of the color. The Hex/RGB code could possibly be interesting to represent, but I do not know if it would be sufficient for a garage to do a good paint job.
> 
> By the way, I found the 10 colors used in the german vehicle registry. So a bit fewer than the 24 we have in Sweden:
> 
> Es stehen insgesamt 10 Farbcodes zur Verfügung, die folgenden Farben zugeordnet werden:
> 
> 0 = Weiß 1 = Gelb 2 = Orange 3 = Rot 4 = Lila bzw. Violett 5 = Blau 6 = Grün 7 = Grau 8 = Braun 9 = Schwarz

To give it another perspective, this is the list of options from [mobile.de](https://services.mobile.de/refdata/colors)

jakubdrabik:
@nickrbb The information might as well be simply out of date. I suggest this:
```
comment: Depending on the context, this attribute might not be up to date or misconfigured, and therefore should be considered untrustworthy in the absence of another method of verification.
```

What do you think?

nickrbb:
I don't understand why you think this is sensitive information (see my explanation as to why I don't think it is in my first paragraph of my posting above). Please can you elaborate?

jakubdrabik:
> I don't understand why you think this is sensitive information (see my explanation as to why I don't think it is in my first paragraph of my posting above). Please can you elaborate?

Even though it's prominent, considering the same system (inside/outside the vehicle) would also keep information about the user, it could potentially lead to risks like privacy invasion etc. However, this is very specific to use cases and probably not a concern of VSS itself. I updated the PR with just the first sentence.

nickrbb:
Looks OK to me now, thanks.

adobekan (approved):
Looks good.

erikbosch:
Meeting notes:

- Ok to merge if build succeeds

PR #622, opened 2023-06-19
Add MapScaleMode to Navigation branch

wonsuk73:
- Need to add MapScaleMode to support the map scale mode switching function to scale fixed mode or auto scale mode in navigation

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -120,12 +120,16 @@ Navigation.GuidanceVoice:
   description: Navigation guidance state that was selected.
   comment: ETC indicates a voice alternative not covered by the explicitly listed alternatives.
 
-Navigation.IsAutoScaleMode:
+Navigation.Map:
+  type: branch
+  description: All map actions
+
+Navigation.Map.IsAutoScaleModeUsed:
   datatype: boolean
   type: actuator
-  description: Used to select auto scale mode.
-  comment: If it is True then auto scale mode is used.
-           If it is False then fixed mode is used.
+  description: Used to select auto-scaling mode. This feature dynamically adjusts the zoom level of the map to provide an optimal view based on the current speed of the vehicle
+  comment: If true, then auto-scaling mode is used.
+           If false, then manual-scaling mode is used.
 
 HMI:
   type: branch
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -127,7 +127,7 @@ Navigation.MapScaleMode:
   description: Map Scale Mode state that was selected.
   comment: SCALE FIXED MODE indicates a map scale is fixed.
            AUTO_SCALE_MODE indicates a map scale is set automatically according to the speed, road class and guidance distance.
-           
+
 HMI:
   type: branch
   description: HMI related signals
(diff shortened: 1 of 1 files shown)


spec/Cabin/Infotainment.vspec
```diff
@@ -120,6 +120,14 @@ Navigation.GuidanceVoice:
   description: Navigation guidance state that was selected.
   comment: ETC indicates a voice alternative not covered by the explicitly listed alternatives.
 
+Navigation.MapScaleMode:
+  datatype: string
+  type: actuator
+  allowed: ['SCALE_FIXED_MODE', 'AUTO_SCALE_MODE']
```
erikbosch, line 126: Do we expect more alternatives in the future? if not an alternative solution could be a boolean signal like `IsAutoScaleMode` (and if false then fixed mode is used). But this approach work for me as well.

wonsuk73: @erikbosch Thanks for valuable comment! I think IsAutoScaleMode is enough. I am going to revise the PR based on the consensus.

wonsuk73: @erikbosch Could you review updated one? Any comment on this?

nickrbb:
The name of the signal and accompanying description are somewhat lacking. Previously the signal name indicated it was for scaling of maps, but now we don't have such an indication. I'd suggest renaming the signal to Navigation.IsMapScaleModeAuto or even Navigation.Map.IsScaleModeAuto if we foresee other signals for navigational maps coming in the future and therefore would benefit by creating a branch here. Regardless, please elaborate the description in the same manner e.g. Used to select between auto-scaling mode and manual-scaling mode of navigation map.

A minor nit: remove the two instances of "it is" from the comment i.e. If true, then..., If false, then...

wonsuk73:
@nickrbb Thanks for valuable comments! I agree to make Navigation.Map branch to cover other map signals. I revised as below. Is it okay to you? 

<pre><code>
Navigation.Map:
  type: branch
  description: All map actions

Navigation.IsAutoScaleMode:
  datatype: boolean
  type: actuator
  description: Used to select auto-scaling mode.
  comment: If true, then auto-scaling mode is used.
                    If false, then manual-scaling is used.
</code></pre>

spec/Cabin/Infotainment.vspec
```diff
@@ -120,6 +120,13 @@ Navigation.GuidanceVoice:
   description: Navigation guidance state that was selected.
   comment: ETC indicates a voice alternative not covered by the explicitly listed alternatives.
 
+Navigation.IsAutoScaleMode:
```
adobekan, line 123: what if we add term "used" into the leaf name? like IsAutoScaleModeUsed

erikbosch:
Meeting notes: Please review/discuss

nickrbb:
@wonsuk73 Signal name looks better, but please also update the description to something more informative and meaningful.

wonsuk73:
@nickrbb Thanks for feedback! Could you review updated one?

<pre><code>
Navigation.Map:
  type: branch
  description: All map actions

Navigation.Map.IsAutoScaleMode:
  datatype: boolean
  type: actuator
  description: Used to select auto-scaling mode. This feature dynamically adjusts the zoom level of the map 
               to provide an optimal view based on the current speed of the vehicle.
  comment: If true, then auto-scaling mode is used.
           If false, then manual-scaling mode is used.
</code></pre>

adobekan (approved):
Looks good

nickrbb:
OK with me now, thank you.

erikbosch:
Meeting notes: Ok to merge

PR #624, opened 2023-06-26
Add additional mirror signals

erikbosch:
Inspired by the [Android signals](https://android.googlesource.com/platform/hardware/interfaces/+/master/automotive/vehicle/2.0/types.hal) MIRROR_LOCK and MIRROR_FOLD.

**Lock Signal**

This signal might reflect the quite common design vehicle where you have common up/down/left/right buttons and a separate switch to decide if it shall affect left/right/none mirror. I.e. in "none" both left and right are locked

**Fold signal**

It can be discussed if we rather would need a percent-signal, where 100% either means fully folded or fully unfolded. Might be useful if you want to monitor progress or control speed.

### DIFF
--- a/spec/Body/ExteriorMirrors.vspec
+++ b/spec/Body/ExteriorMirrors.vspec
@@ -30,3 +30,13 @@ IsHeatingOn:
   datatype: boolean
   type: actuator
   description: Mirror Heater on or off. True = Heater On. False = Heater Off.
+
+IsLocked:
+  datatype: boolean
+  type: actuator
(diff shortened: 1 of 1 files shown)


erikbosch:
Meeting notes:

- Please review until next week
- Possibly add note that True means partially or fully folded mirror
- Ulf: Could be an "allowed value" with 3 values folded/folding/unfolded, i.e. one state for moving or partially folded

ppb2020:
Align with PR#265 for description of IsOpen for Door. (That is, use same approach regarding description of fully folded/fully opened and fully unfolded/fully closed.

erikbosch:
Meeting notes: OK to merge

PR #630, opened 2023-07-10
Add SteeringWheel.Heating

erikbosch:
Inspired by HVAC_STEERING_WHEEL_HEAT in https://android.googlesource.com/platform/hardware/interfaces/+/master/automotive/vehicle/2.0/types.hal

### DIFF
--- a/spec/include/ItemHeatingCooling.vspec
+++ b/spec/include/ItemHeatingCooling.vspec
@@ -1,0 +1,14 @@
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+
+
+HeatingCooling:
+  datatype: int8
+  type: actuator
+  min: -100
+  max: 100
+  unit: percent
+  description: Heating or Cooling requsted for the Item.
+               -100 = Maximum cooling, 0 = Heating/cooling deactivated, 100 = Maximum heating.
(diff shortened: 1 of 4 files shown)


erikbosch:
Meeting notes:

- Please review
- Nick: Maybe call it something with both heat and cooling.
- Pierre: Check if it can be aligned with seat heating/cooling (reusable signal)

erikbosch:
AP: Erik to update PR

SebastianSchildt:
Meeting 08/15 

 - Is "TemperatureControl" a good name (it sounds like HVAC, but this signal is not very usable for that (relative scale))
 - Is "Control" is a good suffix
 - Maybe something something with "Relative" e.g. RelativeTemperature

spec/include/ItemHeatingCooling.vspec
```diff
@@ -0,0 +1,14 @@
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+
+
+HeatingCooling:
```
erikbosch, line 7: Changed `TemperatureControl` (in previous PR version) to `HeatingCooling`, I think that I prefer that over `RelativeTemperature` but we can discuss that at next meeting

erikbosch:
Meeting notes:

- Erik changed name to "HeatingCooling"
- Daniel: Would not use term "temperature" if unit is not celsius, possibly better with "intensity"
- Sebastian: Possibly "HeatingIntensity"
- Erik: Please review/discuss until next week

erikbosch:
Meeting notes: merge

Issue #634, opened 2023-07-19
Process for handling major and minor changes


erikbosch:
We need to agree on a process for how to handle major vs minor releases.

**Background**

Now and then we see a need to do backward incompatible changes, like changing name of a signal or changing semantic meaning. That should only happen in a major release, but we do not want to keep minor changes waiting until we plan for a major release. The solution would be to work with two branches, compared to today when we only use one. 

But then we need to find a good practice for handling those branches.

adobekan:
What if we focus more on deprecation. Until major release we keep both names, if we talk about changing the name of the signal/branch and then with major release we remove old name.

erikbosch:
Deprecation works good for changes like:

- Renaming a signal (Copy the old and rename it, mark the old as deprecated)
- Removing a signal (Just mark it as deprecated)

Deprecation does not work (easily) for:
- Changing type, unit, description or semantics of a signal (while keeping the same name)

A typical example would be if we would like to change unit of `Vehicle.Speed` to `m/s` (instead of `km/h`). Another example is if we would like to do a refactoring like in #625 where signals keep their names, but there might be semantic changes. But maybe we should think of alternative solutions to having multiple branches, as that would require quite some work to keep branches in sync.

A thought that came to my mind is if we could support multiple versions of a signal in the same vspec. Taking the hypothetical unit change for Vehicle.Speed. If we would allow a syntax like below it would be possible for us to support multiple VSS-versions from the same file. I.e. let the tooling support a VSS version parameter and use it to select which definition to use.

Then of course, we would still need to decide what to release when we do a release. Taking the hypothetical example below, we maybe do not think the change in itself merits us releasing v5.0, so we could possibly before next AMM release v4.1 (using the old definition) and a v5.0a1 as a pre-release using the new definition (Alpha release, based on https://peps.python.org/pep-0440/). If a signal/branch has no `vss-version` attribute then it is assumed included in all versions currently supported. We most likely do not want to support more than 2-3 versions at a time. I.e. at some time we will consider v4.X as no longer maintained and then we can cleanup the *.vspec files accordingly





```

Speed:
  vss-version: 4
  datatype: float
  type: sensor
  unit: km/h
  description: Vehicle speed.

Speed:
  vss-version: 5+
  datatype: float
  type: sensor
  unit: m/s
  description: Vehicle speed.
```

adobekan:
[...]

PR #637, opened 2023-08-15
Change datatype for HVAC temeprature 

SebastianSchildt:
It was defined as int, most cars are drive have 0.5°C steps to set temperature, which can not be mapped to the current VSS model.

Changing this to float in the model fixed it, and then it is up to an app/user to represent this with a lower granularity if requried.

### DIFF
--- a/spec/Cabin/SingleHVACStation.vspec
+++ b/spec/Cabin/SingleHVACStation.vspec
@@ -19,7 +19,7 @@ FanSpeed:
   description: Fan Speed, 0 = off. 100 = max
 
 Temperature:
-  datatype: int8
+  datatype: float
   type: actuator
   unit: celsius
   description: Temperature


SebastianSchildt:
Meeting Notes 8/23

 - are datatype changes for major release only? 
 - see also https://github.com/COVESA/vehicle_signal_specification/issues/634
 - before merge we need to decide whether "master" is running towards 5.0 now, do we need a 4.x branch as well?

erikbosch:
I would say that a datatype change in most cases shall be considered as a backward incompatible changeIf a client use a signal with float type then it is reasonable that the server also handles it as a float type.

erikbosch:
Meeting notes:

- OK to merge from content perspective
- Shall be for the next "major release", await merging until minor/major process clarified

erikbosch:
Merging to master, shall not be included in 4.X (and this v4.1 release)

erikbosch (approved):

PR #644, opened 2023-09-12
Fix instance documentation

### LINKED ISSUE #641: Adapt documention to use of DriverSide/PassengerSide
Some time ago we changed https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Cabin/Cabin.vspec and other files to use DriverSide and PassengerSide instead of Left/right, but we forgot to update documentation like https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/rule_set/instances.md. That shall better be addressed.

erikbosch:
Fixes #641

Updates the text concerning instances, needed as we now use DriverSide/PassengerSide. As we did not had the source for the old image I had to create new ones, and is using PlantUML for this. The resulting images are shown below:

image


erikbosch:
Meeting notes:

- ok to merge

Issue #280, opened 2021-05-03
Unicode support


gunnarx:
Placeholder for discussion on allowable characters in the YAML definitions.

- Today: Which are the allowable characters, for example in signal names (right now only letters, numbers and "." encoded in ASCII seems assumed)?
- Future: What are future rules taking into account non-english languages, as well as the general consensus in many programming environments to use unicode file formats (UTF8 would be a good small step of course).
- Also, make sure to compare to YAML specs so that the result is still YAML compliant.

crea7or:
Since JSON and YAML are UTF-8 by default there are not so many choices.

oliveirarleo:
Well, I think more support is good, but this may interfere with some things already defined, for example `.` and `#` are already reserve.

Another thing is that someone will be able to use emojis and other characters that may not be desirable.

PR #651, opened 2023-09-18
Make naming rules more strict

erikbosch:
This is a proposal for discussion.

**Background**

As of today we more or less say that anything that is valid Yaml is valid VSS, which easily gives problem in downstream tools trying to generate code from VSS. An example exists in https://github.com/COVESA/vss-tools/pull/284, and there have been discussions in the past like in #280. 

**Idea**

Already today we have some recommendations written in documentation, and vss-tools at least partially validate them and give a warning, and if you request "strict" checking you might get an error. But should we possibly be more strict and change many of them to rules. That would make life easier for vss-tools, but possibly also for APIs and ontology representation of VSS. Then the "core vss-tool functionality" can do the check, so that exporters does not need to be able to handle VSS-style specifications that does not follow the rules. 

Then we have two alternatives, either free-text rules like here, or "inheriting" rules from some other definition (XML? RDF?). Or describing them in some other format. Suggestions are welcome. Most likely we cannot enforce rules that consider keywords in "all" potential output formats, but perhaps we can consider the most popular output formats.  

**Way forward**

First agree on if we want to add additional rules. Then if so, define them and implement them in vss-tools. Preferably, any changed rule should not affect signals currently being part of standard catalog.

Feedback is welcome.

### DIFF
--- a/docs-gen/content/rule_set/basics.md
+++ b/docs-gen/content/rule_set/basics.md
@@ -103,14 +103,31 @@ Vehicle.Navigation.CurrentLocation:
 
 It is recommended for servers, which are implementing protocols for the vehicle signal specification, to serve old and new nodes during the deprecation period described above.
 
+## Naming Rules
+
+Different languages and environments have different rules for identifiers.
+To simplify usage of VSS in as many languages and environments as possible the following rules apply.
+
+* The allowed character set for node names (branch and data nodes) is `A-Z`, `a-z` and `0-9`.
+* Node names must not start with a digit.
+* Names are case sensitive, but it is not allowed to have two nodes that only differ by case,
+  i.e. it is not allowed to have both `Vehicle.Aaa` and `Vehicle.AaA`.
+* It is not allowed to reuse the same node name as a parent or ancestor node for a child node,
+  i.e., it is not allowed to have a child `A` if the parent or any ancestor is called `A`
+* For string literals only the characters `A-Z`, `0-9` and `_` are allowed,
+  and the value must start with `A-Z`. See the [chapter](/vehicle_signal_specification/rule_set/data_entry/allowed/) for specifying allowed values.
+
+It is recommended that VSS Tools shall give an error if the rules above are not followed.
+
 ## Style Guide
 
-The VSS specification must adhere to YAML syntax. To keep the standardized VSS specification in this repository consistent the following style guide is provided.
+For a contribution to VSS standard catalog the following recommendations must be followed:
 
 ### Naming Conventions
 
-The recommended naming convention for node elements is to use camel case notation starting with a capital letter. It is recommended to use only
-`A-Z`, `a-z` and `0-9` in node names. For boolean signals it is recommended to start the name with `Is`.
+* The recommended naming convention for node elements is to use camel case notation starting with a capital letter.
+* For boolean signals it is recommended to start the name with `Is`.
+* For acronyms in node names only the first letter in the acronym shall be capitalized, like `Vehicle.Adas.Abs.IsAbsEnabled`
 
 Examples:
 
@@ -118,7 +135,6 @@ Examples:
 SomeBranch.AnotherBranch.MySignalName
 Vehicle.Cabin.Door.Row1.Left.IsLocked
 ```
-Naming convention for string literals can be found in the [chapter](/vehicle_signal_specification/rule_set/data_entry/allowed/)for specifying allowed values.
 
 ### Line Length
 
--- a/docs-gen/content/rule_set/data_entry/allowed.md
+++ b/docs-gen/content/rule_set/data_entry/allowed.md
@@ -27,7 +27,7 @@ in a list.  Only values can be assigned to the data entry, which are
 specified in this list.
 
 The `datatype` specifier gives the type of the individual elements of the `allowed`
-list. In general `allowed:` is valid for all datatypes, including integer- and float-based types, unless othwerise specified.
+list. In general `allowed:` is valid for all datatypes, including integer- and float-based types, unless otherwise specified.
 
 ## Recommendation on String values
 


docs-gen/content/rule_set/basics.md
```diff
@@ -103,22 +103,37 @@ Vehicle.Navigation.CurrentLocation:
...
 
+## Naming Rules
+
+Different languages and environments have different rules for identifiers.
+To simplify usage of VSS in as many languages and environments as possible the following rules apply.
+
+* The allowed character set for node names (branches and signals) is `A-Z`, `a-z` and `0-9`
+* Node names must not start with a digit
+* Names are case sensitive, but it is not allowed to have two nodes that only differ by case,
+  i.e. it is not allowed to have both `Vehicle.Aaa` and `Vehicle.AaA`
```
ppb2020, line 114: i.e. -> i.e.,

docs-gen/content/rule_set/basics.md
```diff
@@ -103,22 +103,37 @@ Vehicle.Navigation.CurrentLocation:
 
 It is recommended for servers, which are implementing protocols for the vehicle signal specification, to serve old and new nodes during the deprecation period described above.
 
+## Naming Rules
+
+Different languages and environments have different rules for identifiers.
+To simplify usage of VSS in as many languages and environments as possible the following rules apply.
+
+* The allowed character set for node names (branches and signals) is `A-Z`, `a-z` and `0-9`
```
ppb2020, line 111: branches and signals -> branch and data nodes

Perhaps? To align with terms used elsewhere?

docs-gen/content/rule_set/basics.md
```diff
@@ -103,22 +103,37 @@ Vehicle.Navigation.CurrentLocation:
...
+* Node names must not start with a digit
```
ppb2020, line 112: Period at end of sentence?

docs-gen/content/rule_set/basics.md
```diff
@@ -103,22 +103,37 @@ Vehicle.Navigation.CurrentLocation:
...
+* Names are case sensitive, but it is not allowed to have two nodes that only differ by case,
+  i.e. it is not allowed to have both `Vehicle.Aaa` and `Vehicle.AaA`
+* It is not allowed to reuse the same node name as a parent node in a child node,
```
ppb2020, line 115: This reads like if you mean the immediate parent, as opposed to all parents.

erikbosch:
meeting notes:

- Sebastian: Text in PR not totally consistent, if it is a rule it shall give an error, not optional to give an error. Some items in rules should better be guidelines
- Erik: But if we for instance allow all unicode characters, what do we then expect from tools; should tools then handle them in a consistent way? Or ok to give error?

docs-gen/content/rule_set/basics.md
```diff
@@ -103,22 +103,37 @@ Vehicle.Navigation.CurrentLocation:
...
 ## Style Guide
 
-The VSS specification must adhere to YAML syntax. To keep the standardized VSS specification in this repository consistent the following style guide is provided.
+For a contribution to VSS standard catalog the following recommendations must be followed:
 
 ### Naming Conventions
 
-The recommended naming convention for node elements is to use camel case notation starting with a capital letter. It is recommended to use only
-`A-Z`, `a-z` and `0-9` in node names. For boolean signals it is recommended to start the name with `Is`.
+The recommended naming convention for node elements is to use camel case notation starting with a capital letter.
```
ppb2020, line 128: As well, it seems to be common practice that acronyms are in uppercase in node names (which, by the way, makes it a bit awkward to read when the acronym is in the middle of a node name).

Given that acronyms are indeed "words", my personal preference is for acronyms in code, etc. to follow the same case use as codes. For example, Vehicle.Adas.Abs.IsEnabled.

docs-gen/content/rule_set/basics.md
erikbosch: Will change it. We are a bit inconsistent in documentation, like in https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/. We sometimes talks about signals, sometimes about data entries.

docs-gen/content/rule_set/basics.md
erikbosch: Adding a period to all bullet points

docs-gen/content/rule_set/basics.md
erikbosch: Will try to rephrase it

docs-gen/content/rule_set/basics.md
erikbosch: Adding that to the PR to get comments

SebastianSchildt:
I personally think in 2023 it is a bit to restrictive to say 
 > The allowed character set for node names (branch and data nodes) is `A-Z`, `a-z` and `0-9`.

Many (most?) modern languages do allow some form of unicode in identifiers, e.g. https://doc.rust-lang.org/reference/identifiers.html , so while for greatest compatibility with all outputs it might be desirable in standard catalogue, I would not limit us to "7 bit ASCII" here. 

I would leave this as a recommendation,  and only require no whitespaces.

docs-gen/content/rule_set/basics.md
SebastianSchildt: Personally, in my eyes, reading "Adas" or "Abs" looks weird as opposed to "ADAS" or "ABS". In case it turns out I am in the minority here, I can live with it :)

argerus:
> I personally think in 2023 it is a bit to restrictive to say
> 
> > The allowed character set for node names (branch and data nodes) is `A-Z`, `a-z` and `0-9`.

> Many (most?) modern languages do allow some form of unicode in identifiers, e.g. https://doc.rust-lang.org/reference/identifiers.html
> , so while for greatest compatibility with all outputs it might be desirable in standard catalogue, I would not limit us to "7 bit ASCII" here.

Even in a modern language such as Rust, you are only allowed to use the **alpha-numeric** characters of unicode in the identifiers (+ underscore).

I suppose that C++23 restricts them to alpha-numeric characters of ASCII + underscore.

Now, it's perfectly possible to use paths with arbitrary characters in both of these languages, just not as identifiers. But when such paths are used in other contexts, i.e. as identifiers in protocols, access tokens, config files etc, allowing arbitrary special characters (or even alpha-numeric unicode) will most likely lead to issues.

> I would leave this as a recommendation, and only require no whitespaces.

Wouldn't it make more sense to require alpha-numeric characters in that case (unicode or otherwise)?

But honestly, if the point is to promote interoperability, aiming for a common denominator by restricting what characters are allowed in an identifier is probably not such a bad idea.

Allowing _the content_ addressed by these identifiers to contain other things than 7bit ascii, now that's something that should be expected in 2023.

ppb2020:
Perhaps on thing to think of: restrict the set of characters in the COVESA catalog to what works in all formats prodiuced by the tooling, but allow other characters such that users of overlays that do not require this level of compatibility are not hampered by the limitation. Say, add a --strict option to that, when used, will check and guarantee that the produced artifacts will for in any format (IDL, JSON, etc.). We validate any submission here using --strict, but anyone defining their own signals can opt not to use --strict and use whatever they want, as long as the can generate the atalog in whatever set of formats they use.

PR #656, opened 2023-09-20
Remove UUID from release artifacts

erikbosch:
This is a proposal to remove UUIDs from the default assets that we produce/attach as part of releases, like in https://github.com/COVESA/vehicle_signal_specification/releases/tag/v4.0. The rationale is that they make the resulting artifacts bigger and at least of today the UUIDs not that frequently used.

This is however no urgent change and we may leave this PR open until we have agreed on the "new" UUID handling that some people are working on, as that may affect how useful or not it is to include UUIDs in release artifacts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,4 +184,7 @@ have been removed from the [VSS syntax](docs-gen/content/rule_set/data_entry/sen
 
 ## Planned Changes VSS 5.0
 
--
+### No UUID in relese artifacts
+
+From VSS 5.0 there will be no UUIDs in [release artifacts](https://github.com/COVESA/vehicle_signal_specification/releases)
+and the commands in the `Makefile` used for both CI and release artifacts will no longer include the `--uuid`argument.
--- a/Makefile
+++ b/Makefile
@@ -21,37 +21,38 @@ optional_targets: clean protobuf ttl
 TOOLSDIR?=./vss-tools
 
 json:
-	${TOOLSDIR}/vspec2json.py -I ./spec --uuid -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).json
+	${TOOLSDIR}/vspec2json.py -I ./spec -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).json
 
 franca:
-	${TOOLSDIR}/vspec2franca.py -v $$(cat VERSION)  -I ./spec --uuid -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).fidl
+	${TOOLSDIR}/vspec2franca.py -v $$(cat VERSION)  -I ./spec -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).fidl
 
 yaml:
-	${TOOLSDIR}/vspec2yaml.py -I ./spec --uuid -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).yaml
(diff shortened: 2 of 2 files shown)


ppb2020:
In an earlier implementation, we were using UUID as unique indexes into tables representing a signal catalog in our product. We have since moved to an enhanced form of UUIDs that we have created ourselves, and thus no longer rely on UUIDs. However, the fact that we did use them would imply that other implementations may be using them as well. It may not be possible to determine if UUIDs are in use in some implementation as of today. Of course, removing them is one way to find out. Is this what we are considering here?

erikbosch:
> In an earlier implementation, we were using UUID as unique indexes into tables representing a signal catalog in our product. We have since moved to an enhanced form of UUIDs that we have created ourselves, and thus no longer rely on UUIDs. However, the fact that we did use them would imply that other implementations may be using them as well. It may not be possible to determine if UUIDs are in use in some implementation as of today. Of course, removing them is one way to find out. Is this what we are considering here?

I agree that it is impossible to know with 100% certainty how our release artifacts are used, and what changes we can do without causing problems for anyone. The only way to get some indication is to ask. Well we could off course remove, release and see what reactions we get but better to get comments before doing anything. BMW have said that they will contribute a proposal for updating the UUID mechanism, so even if we keep UUIDs in the release artifacts they might look different than today if we change how they are generated.

If we agree on that it is best to keep including UUIDs in our release artifacts I have no problems closing this PR. Theoretically we could release artifacts both with and without UUIDs, but then number of release artifacts will be doubled. If we just release one variant (with or without UUID) it would till be relatively easy for someone to generate the other artifact themselves.

ppb2020:
I was thinking more of deprecating them first before removal.

erikbosch:
> I was thinking more of deprecating them first before removal.

Tes, that is also a possibility. That would in practice mean that it would be added as a note in https://github.com/COVESA/vehicle_signal_specification/blob/master/CHANGELOG.md, like "From VSS 6.0 UUIDs will not be included in release artifacts"

erikbosch:
Meeting notes:

- Please review, do we want to keep UUIDs or not in release artifacts?
-

erikbosch:
No we have the proposal for static ID in https://github.com/COVESA/vss-tools/pull/305. But that one only adds a new tool, it does not change current UUID generation mechanism. That solution (with consecutive IDs) is also from some aspects very different from generating a UUID or some other unique identifier by some clever hash mechanism, shown for example by Blackberry. Maybe we need both mechanisms in VSS-tool, i.e. keep support for unique identifiers, but possibly change or make it configurable how they can be generated.

Meeting discussion:

- We can postpone uuid argument discussion until later, for 4.1 we just add the new static id tool
- Adnan: Think it is fine if we remove UUID from release artifacts.
- Decision: Keep uuid in release artifacts for 4.1, remove for 5.0
- Discuss later if --uid argument shall be kept, possibly used for integration of Blackberry solution. 
- Erik to check with Pierre how they generate their uid

ppb2020:
BlackBerry manages signal definitions using JSON files. We post-process the files generated by vss-tools to compute and add the EUUIDs as an element called "x-euuid". The same tool adds read and write permissions to each node also as elements: "x-read-permission" and "x-write-permission". These are used to control access to each node in the signal catalog. Applications that access nodes must have the corresponding permission, otherwise access is denied.

The tool visits each node in the signal catalog (tree), takes the node path name and concatenates to it the value of each semantically relevant element present fort that node, then hashes this string into a 64 bit value using FNv1

erikbosch:
Thanks @ppb2020 , so then my conclusion is that you do not need the vss-tools "--uuid" feature. But still, the question is if anyone else use it. Maybe we could add a deprecation warning if anyone use `--uuid` and encourage them to add an issue if they want to feature to be kept. If no one steps forward we could in the long term remove it.

erikbosch:
Haha, I forgot this one when adding #721, they do more or less the same, keeping the new one.

PR #657, opened 2023-09-22
Adding signals for width and turn diameter

erikbosch:
This is inspired by some signals in Android VHAL with no corresponding signals in VSS; In `INFO_EXTERIOR_DIMENSIONS` there are fields for width with and without mirrors and also turn diameter.

We could possibly make use of 3 widths

- Excluding mirrors
- With mirrors folded
- With mirrors in normal position

Here we have SAE definitions we could use. A possible problem is that we already have a Width signal, but that signal does not specify what width it concerns. This PR boldly assumes that the existing signal refers to the width without mirrors. It can be discussed if this is a reasonable assumption. (If so this could be considered a minor change, if not this is a major change)

For turn diameter two measurement can be used, wall-to-wall and curb-to-curb. The SAE standard only give a definition for wall-to-wall:

_"Diameter of the minimum cylindrical envelope in which the
vehicle can drive a circle with the front wheels at maximum
lock (between walls). If different between left and right turn,
the larger one is reported"._


erikbosch:
Meeting notes:

- Please review
- Nick: Better to duplicate signal instead of changing meaning, like `WidthExcludingMirrors`, Keep `Width`, mark obsolete

SebastianSchildt:
I don't think in VSS we want both turning radii.  More practical seems to have "wall-to-wall", see also Wikipedia, just calling that TuringDiameter, and write in description the specific definition used

https://en.wikipedia.org/wiki/Turning_radius

erikbosch:
Meeting notes:

- Please review, potential merge at next meeting

erikbosch:
Meeting note: merge

Issue #667, opened 2023-10-05
Datatype/Unit for Timestamps


erikbosch:
As of today we represent Timestamps in COVESA as string signals where we in free text specify that the value is a timestamp and shall be formatted according to ISO 8601. This is however difficult to interpret for tooling, there it would be better to have a formal notation that this signal actually contains a timestamp. I see two "easy" ways to solve this:

- One is to invent a new datatype "timestamp", "iso_8601_timestamp" or similar, where we in documentation says that this datatype by default should be a string formatted according to ISO 8601. 
- Another possibility is to use `unit` for this purpose ( and keep `string` as type).

Both methods would help SDKs/APIs/GUIs to know that this should be a timestamp so they can offer a nice date-picker and/or a timestamp-validator. If using a downstream project like VISS or KUKSA you should still use an ISO 8601 string when exchanging data.

Any thoughts?

https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types/#timestamps

_Timestamps are in VSS typically represented as strings, formatted according to ISO 8601. Timestamps shall be expressed in UTC (Coordinated Universal Time), with special UTC designator (“Z”). Time resolution SHALL at least be seconds, with subsecond resolution as an optional degree of precision when desired. The time and date format shall be as shown below, where the sub-second data and delimiter is optional._

ppb2020:
Unit is typically used to tell the user how to interpret a value. Thus, I think "iso8601" as a unit would make sense. I would remove the restriction that the timestamp be in UTC, as ISO 8601 allows the specification of a time zone in teh string, which means that it is always possible to convert back to UTC.

One issue with this approach, though, is that it also requires the dataype to be a string. AFAIK, there no such enforcement today (that is, no enforcement of a specific datatype for any unit). Should this be considered?

This downside is resolved by the first proposed solution, as a datatype of "ISO8601" or "ISO_8601" could be enforced to be a string with a specific format. Here, I would also allow time zone specification as per ISO 8601

This aligns with some thoughts I have been having of proposing a new datatype called, say, "reference", where the datatype would be a string that is expected to contain a node path name. For example:

```
DriverPosition:
  datatype: reference
  type: attribute
  default: "Vehicle.Cabin.Seat.Row1.Pos1"
  description: The position of the driver seat.
```

This is similar to a datatype of "iSO8601", where the value is expected to be a string in a specific format that can be parsed and validated.

erikbosch:
[...]

Issue #588, opened 2023-05-02
Use of "unit" - make it mandatory?


erikbosch:
Based on discussion in meeting 2023-05-02:

- Shall we make units mandatory to make sure that we add a unit where necessary
- Then we need an explicit unit "none" or "unitless" to avoid warnings when there is no unit.

erikbosch:
We regularly close old issues and pull requests that appear to be abandoned or completed, where there have not been any activity for a long time. If you think your item is still relevant please feel to add a comment and re-open it. You are also very welcome to bring up the topic at one of our weekly VSS meetings.

PR #669, opened 2023-10-13
Refactoring unit file

erikbosch:
This is a draft for discussion only, partially inspired by #667, partially inspired by discussions at the AMM.

At the AMM we discussed if signals shall have a unit (like km/h) or just a domain/dimension (like speed). With the latter approach it would be possible to send around values with different units, as long as the unit-information is included the receiver can convert it to whatever the user wants. On the other hand, in some APIs/solutions like VISS and KUKSA gRPC unit information is never explicitly included when you get an observation, you just get a value which is assumed to be in the unit defined for that signal

Based on that I come up with some ideas:

-  Should we possibly specify in the VSS unit file how different units like `km/h` and `m/s` relate to each other? Even if it might be easy to understand for a human how to convert between them it might be more difficult for a computer. One way could be to say the we use km/h as default and all other speed units must specify how you they relate to km/h. That would allow for VSS-based solutions to offer solutions where you can get the signal value in whatever VSS unit you want, like `get('Vehicle.Speed', 'm/s')`
- For time it is infeasible to specify how to convert between iso8061 and unix timestamp in the unit file itself, so there one may need to specify that custom (hardcoded) conversion methods must be implemented
- What fields/terms do we actually need for units? Shall we possibly remove "label", it is not used and does not seem to provide much value if we have both name and description.
- Shall we possibly specify supported datatypes for each unit? If we do so the tooling could give an error or warning if you use another datatype. Like combining iso8061 with a numeric type, or a unix timestamp with string or a small integer.

I would like to know your opinion on if this seems to be reasonable direction for improving/extending the unit file. If we agree on that then we can start looking into details. That CI will fail for this PR is expected - as of today "label" is a required property, so if we agree to make it optional we must do a minor change in vss-tools.

_Note: I think there are a lot of things that could be discussed. Like shall we keep the term "domain", or would it be better to call it "dimension" or "quantity" (term used in the NIST document). Similarly would "time" + "duration" be better than "point in time" + "time"._

### DIFF
--- a/docs-gen/content/rule_set/data_entry/data_unit_types.md
+++ b/docs-gen/content/rule_set/data_entry/data_unit_types.md
@@ -28,7 +28,7 @@ But in protocols not explicitly specifying data units (like [VISS](https://raw.g
 it is expected that `Vehicle.Speed` is sent and received as `km/h` (without scaling or offset).
 
 In some cases it is natural to omit the data unit. This concerns typically signals where datatype `string` is used,
-but also signals where the value just represents a number, like in the example below:
+but also signals where the value just represents a number (dimensionless quantities), like in the example below:
 
 ```
 Vehicle.Cabin.DoorCount:
@@ -74,67 +74,133 @@ Vehicle.Powertrain.Transmission.ClutchWear:
 
 ## List of supported Data Unit
 
-This list composed with definition according to International Units (SI) and few automotive specific units: [Specification](https://www.iso.org/standard/30669.html), [Wikipedia](https://en.wikipedia.org/wiki/International_System_of_Units)
-
-
-Unit          | Domain                          | Description
-:-------------|:--------------------------------|:-------------
-mm            | Distance                        | Distance measured in millimeters
-cm            | Distance                        | Distance measured in centimeters
-m             | Distance                        | Distance measured in meters
-km            | Distance                        | Distance measured in kilometers
-inch          | Distance                        | Distance measured in inches
-km/h          | Speed                           | Speed measured in kilometers per hours
-m/s           | Speed                           | Speed measured in meters per second
-m/s^2         | Acceleration                    | Acceleration measured in meters per second squared
-cm/s^2        | Acceleration                    | Acceleration measured in centimeters per second squared
-ml            | Volume                          | Volume measured in milliliters
-l             | Volume                          | Volume measured in liters
-cm^3          | Volume                          | Volume measured in cubic centimeters
-celsius       | Temperature                     | Temperature measured in degree celsius
-degrees       | Angle                           | Angle measured in degrees
-degrees/s     | Angular Speed                   | Angular speed measured in degrees per second
-W             | Power                           | Power measured in watts
-kW            | Power                           | Power measured in kilowatts
-PS            | Power                           | Power measured in horsepower
-kWh           | Energy Consumption              | Energy consumption measured in kilowatt hours
-g             | Weight                          | Mass measured in grams
-kg            | Weight                          | Mass measured in kilograms
-lbs           | Weight                          | Mass measured in pounds
-V             | Electric Potential              | Electric potential measured in volts
-A             | Electric Current                | Electric current measured in amperes
-Ah            | Electric Charge                 | Electric charge measured in ampere hours
-ms            | Time                            | Time measured in milliseconds
-s             | Time                            | Time measured in seconds
-min           | Time                            | Time measured in minutes
-h             | Time                            | Time measured in hours
-day           | Time                            | Time measured in days
-weeks         | Time                            | Time measured in weeks
-months        | Time                            | Time measured in months
-years         | Time                            | Time measured in years
-UNIX Timestamp| Time                            | Unix time is a system for describing a point in time. It is the number of seconds that have elapsed since the Unix epoch, excluding leap seconds.
-mbar          | Pressure                        | Pressure measured in millibars
-Pa            | Pressure                        | Pressure measured in pascal
-kPa           | Pressure                        | Pressure measured in kilopascal
-stars         | Rating                          | Rating measured in stars
-g/s           | Mass per time                   | Mass per time measured in grams per second
-g/km          | Mass per distance               | Mass per distance measured in grams per kilometers
-kWh/100km     | Energy Consumption per distance | Energy consumption per distance measured in kilowatt hours per 100 kilometers
-ml/100km      | Volume per distance             | Volume per distance measured in milliliters per 100 kilometers
-l/100km       | Volume per distance             | Volume per distance measured in liters per 100 kilometers
-l/h           | Flow                            | Flow measured in liters per hour
-mpg           | Distance per Volume             | Distance per volume measured in miles per gallon
-N             | Force                           | Force measured in newton
-Nm            | Torque                          | Torque measured in newton meters
-rpm           | Rotational Speed                | Rotational speed measured in revolutions per minute
-Hz            | Frequency                       | Frequency measured in hertz
-ratio         | Relation                        | Relation measured as ratio
-percent       | Relation                        | Relation measured in percent
-... | ... | ...
-
-[VSS-Tools](https://github.com/COVESA/vss-tools) require that a unit file is available when transforming *.vspec files.
-It can be specified by the `-u` parameter, and if not given the tools will search for a file `units.yaml`
+The VSS syntax does not in itself specify what units can be used, the unit attribute as declared for signals in *.vspec files is optional and can contain an arbitrary string value.
+[VSS-Tools](https://github.com/COVESA/vss-tools) however require that all units used are defined.
+Units are defined by including them in a unit file with syntax as described below.
+One or more unit files can be specified by the `-u` parameter and, if not given, the tools search for a file `units.yaml`
 in the same directory as the root *.vspec file.
 
-The VSS standard catalog is based on [units.yaml](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml),
-i.e. only units from that file can be used.
+For the VSS standard catalog the VSS-project has defined a set of units that can be used for signals in the VSS standard catalog.
+This list is composed of definitions according to International Units (SI) and few automotive-specific units:
+[Specification](https://www.iso.org/standard/30669.html), [Wikipedia](https://en.wikipedia.org/wiki/International_System_of_Units)
+
+The VSS list of units for the standard catalog exists in [units.yaml](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml).
+
+
+## Unit file syntax
+
+Unit files follow the syntax defined below:
+
+```
+[
+ <vss-unit-identifier>: # Typically unit abbreviation, like km/h or mm, but
+    definition: <string>
+    [unit: <string>] # Full name of unit, optional, if not given assumed to be equal to vss-unit-identifier
+    quantity: <string> # Quantity of the unit.
+    [allowed-datatypes] : ['numeric', 'string', uint8', ...]] # Allowed datatypes in VSS standard catalog
+    [deprecation: <reason>]
+]*
+```
+
+The VSS term `quantity` corresponds roughly to the term `quantity` as defined in for example ISO 80000.
+Typical example are `length`, `mass` and `velocity` that all can be expressed in SI-units.
+However, from a VSS perspective quantities do not need to correspond to physical quantities.
+It could be an arbitrary term, but it is generally expected that it is possible to convert between values
+using different units but defined with the same quantity.
+
+Example:
+
+```
+m:
+  definition: Length measured in meters
+  unit: meter
+  quantity: length
+  allowed-datatypes: ['numeric']
+
+mm:
+  definition: Length measured in millimeters
+  unit: millimeter
+  quantity: length
+  allowed-datatypes: ['numeric']
+```
+
+As `m` and `mm` are defined with the same quantity it is expected that you can convert a value from `m` to `mm`.
+
+The `allowed-datatypes` attributes can be used to specify which VSS datatypes can contain values of a particular unit.
+For most units the symbolic datatype `numeric` meaning any float or integer datatype can be used.
+As an example, both `uint8` and `float` can be used to represent a length value.
+If using `uint8` you will have range restrictions, but that might be acceptable for some signals.
+For some units more specific datatype restrictions are relevant. Some examples:
+
+* A date/time expressed in ISO 8061 format can only be represented as a string
+* A UNIX Timestamp signal must be at least 32 bit unsigned to be able to handle date/time after year 2038.
+
+The `deprecation` keyword can be used to indicate that a specific unit may be removed in the future.
+Tooling shall preferably give a warning if a signal uses a deprecated unit or the unit used belongs to a deprecated quantity.
+The reason should preferably list when and why the unit is deprecated, a hypothetical example is given below:
+
+```
+
+  inch:
+    definition: Distance measured in inches
+    unit: inch
+    quantity: distance
+    deprecation: V5.0 replaced by 'in'
+  in:
+    definition: Distance measured in inches
+    unit: inch
+    quantity: distance
+```
+
+
+The unit syntax has recently been changed. To simplify transition to the new syntax it is recommended that
+tooling also supports unit files using the old syntax described below.
+
+```
+units:
+  [
+    <vss-unit-identifier>: # Typically unit abbreviation, like km/h or mm, but
+      label: <string> # Replaced with "unit"
+      description: <string> # Replaced with "definition"
+      quantity: <string>
+  ]*
+```
+
+## Quantity file syntax
+
+
+Defing quantities is recommended, but currently optional for backward compatibility reasons.
+If tooling supports quantity files it can verify that all units provided in unit files
+use defined quantities.
+
+``
+
+  [
+    <vss-quantity-identifier>: # Identifier preferably taken from a standard, like ISO 80000
+      definition: <string>
+      [remark: <string>] # remark as defined in for example ISO 80000
+      [comment: <string>]
+      [deprecation: <reason>]
+  ]*
+
+```
+
+
+The VSS list of quantitiess for the standard catalog exists in [quantities.yaml](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/quantities.yaml).
+
+## Defining custom units
+
+It is possible to define custom units in a unit file.
+Assume for instance you want to have a signal showing remaining range in [furlong](https://en.wikipedia.org/wiki/Furlong).
+Then you could add an additional unit `furlong`. No need to specify `unit` or `symbol` as they equals the default (i.e. "furlong").
+As this is unit not commonly used and not described in any standards, it might be relevant to describe how it can be converted to other units.
+That is however only informative, as it a custom unit a downstream implementation supporting unit conversion may not support automatic conversion
+of furlong to other units.
+
+```
+units:
+  furlong:
+    definition: Length measured in furlong, 1 furlong equals 201.1680 m
+    quantity: length
+    allowed-datatypes: ['numeric']
+```
--- a/spec/units.yaml
+++ b/spec/units.yaml
@@ -1,238 +1,389 @@
 # This file defines units supported in the VSS standard catalog
-units:
-  mm:
-    label: millimeter
-    description: Distance measured in millimeters
-    domain: distance
-  cm:
-    label: centimeter
-    description: Distance measured in centimeters
-    domain: distance
-  m:
-    label: meter
-    description: Distance measured in meters
-    domain: distance
-  km:
-    label: kilometer
-    description: Distance measured in kilometers
-    domain: distance
-  inch:
-    label: inch
-    description: Distance measured in inches
-    domain: distance
-  km/h:
-    label: kilometer per hour
-    description: Speed measured in kilometers per hours
-    domain: speed
-  m/s:
-    label: meters per second
-    description: Speed measured in meters per second
-    domain: speed
-  m/s^2:
-    label: meters per second squared
-    description: Acceleration measured in meters per second squared
-    domain: acceleration
-  cm/s^2:
-    label: centimeters per second squared
-    description: Acceleration measured in centimeters per second squared
-    domain: acceleration
-  ml:
-    label: milliliter
-    description: Volume measured in milliliters
-    domain: volume
-  l:
-    label: liter
-    description: Volume measured in liters
-    domain: volume
-  cm^3:
-    label: cubic centimeters
-    description: Volume measured in cubic centimeters
-    domain: volume
-  celsius:
-    label: degree celsius
-    description: Temperature measured in degree celsius
-    domain: temperature
-  degrees:
-    label: degree
-    description: Angle measured in degrees
-    domain: angle
-  degrees/s:
-    label: degree per second
-    description: Angular speed measured in degrees per second
-    domain: angular speed
-  W:
-    label: watt
-    description: Power measured in watts
-    domain: power
-  kW:
-    label: kilowatt
-    description: Power measured in kilowatts
-    domain: power
-  PS:
-    label: horsepower
-    description: Power measured in horsepower
-    domain: power
-  kWh:
-    label: kilowatt hours
-    description: Energy consumption measured in kilowatt hours
-    domain: energy consumption
-  g:
-    label: gram
-    description: Mass measured in grams
-    domain: mass
-  kg:
-    label: kilogram
-    description: Mass measured in kilograms
-    domain: mass
-  lbs:
-    label: pound
-    description: Mass measured in pounds
-    domain: mass
-  V:
-    label: volt
-    description: Electric potential measured in volts
-    domain: electric potential
-  A:
-    label: ampere
-    description: Electric current measured in amperes
-    domain: electric current
-  Ah:
-    label: ampere hours
-    description: Electric charge measured in ampere hours
-    domain: electric charge
-  ms:
-    label: millisecond
-    description: Time measured in milliseconds
-    domain: time
-  s:
-    label: second
-    description: Time measured in seconds
-    domain: time
-  min:
-    label: minute
-    description: Time measured in minutes
-    domain: time
-  h:
-    label: hour
-    description: Time measured in hours
-    domain: time
-  day:
-    label: days
-    description: Time measured in days
-    domain: time
-  weeks:
-    label: weeks
-    description: Time measured in weeks
-    domain: time
-  months:
-    label: months
-    description: Time measured in months
-    domain: time
-  years:
-    label: years
-    description: Time measured in years
-    domain: time
-  UNIX Timestamp:
-    label: UNIX Timestamp
-    description: Unix time is a system for describing a point in time. It is the number of seconds that have elapsed since the Unix epoch, excluding leap seconds.
-    domain: point in time
-  mbar:
-    label: millibar
-    description: Pressure measured in millibars
-    domain: pressure
-  Pa:
-    label: pascal
-    description: Pressure measured in pascal
-    domain: pressure
-  kPa:
-    label: kilopascal
-    description: Pressure measured in kilopascal
-    domain: pressure
-  stars:
-    label: stars
-    description: Rating measured in stars
-    domain: rating
-  g/s:
-    label: grams per second
-    description: Mass per time measured in grams per second
-    domain: mass per time
-  g/km:
-    label: grams per kilometer
-    description: Mass per distance measured in grams per kilometers
-    domain: mass per distance
-  kWh/100km:
-    label: kilowatt hours per 100 kilometers
-    description: Energy consumption per distance measured in kilowatt hours per 100 kilometers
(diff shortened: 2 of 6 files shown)


spec/units.yaml
```diff
@@ -1,238 +1,236 @@
...
+    default-unit: s
 units:
   mm:
-    label: millimeter
     description: Distance measured in millimeters
     domain: distance
+    # Multiply this value by 1000 to get default unit (meters)
+    multiplier: 1000
+    # List of allowed datatypes. Numeric is short-cut for all int/uint types
+    allowed-datatypes: ['numeric']
```
erikbosch, line 32: The idea here is to give a warning or error if someone defines a VSS signal that has the unit `mm` but where the datatype is `string`, `boolean` or some struct type. Array types based on any numeric type should be accepted.

adobekan:
here is nice ref document.
https://www.govinfo.gov/content/pkg/GOVPUB-C13-f10c2ff9e7af2091314396a2d53213e4/pdf/GOVPUB-C13-f10c2ff9e7af2091314396a2d53213e4.pdf

spec/units.yaml
```diff
@@ -1,238 +1,272 @@
...
     domain: pressure
   kPa:
-    label: kilopascal
     description: Pressure measured in kilopascal
     domain: pressure
+    multiplier: 1000
+  psi:
+    description: Pressure measured in pounds per square inch
+    domain: pressure
+    multiplier: 6894.7572931783
```
erikbosch, line 209: Would be 6894.757 if we would follow the recommendation to use 7 significant digits in the document referenced by Adnan

Kostadin-Ivanov:
I see that using such appendices with unit conversion and other guides could be useful, but I agree with Sebastian that the VSS, being a standard should only define default unit and provide list with, let's say other allowed units.

This way each VSS user / application should be able to easy validate such signals and will make sure to be VSS compliant etc.

With regards to messages and data transport between different VSS compliant edges (applications), it would be sufficient that sender to declare its unit + value, and the receiver would either read it as it is (in case both use same unit), or convert it to its own unit for the corresponding signal (attribute/measurement).

Like I said, having unit conversions rules as a reference guides will be very handy for both user reads and tooling  / implementation, but not critical since every user can have their own way to read (convert) the data depending on their use case.

SebastianSchildt:
> Concerning only using dimension and default unit - we have today many signals that use the domain/dimension "distance" but very different units, like TraveledDistance (km), Length (mm) and Axle.WheelDiameter (inch). Theoretically all could be represented by meter, but that would be a quite big difference and then likely we would need to make sure that all use float as datatype. Also, especially for wheel/tire-related data metric values are uncommon also in Europe. You rarely see someone bragging about their 48 cm wheels, but sometimes about their 19 inch wheels. By that reason I think it makes sense to keep the possibility for signals to use a different unit than the default unit for that dimension

I don't disagree, I just think _dimension_ and _(default)_unit_ are a property of a signal, but there is no need for a unit file to "list all combinations". So today, Vehicle.Speed is defined as "km/h", and a dimension you might only learn indirectly (i.e. our own tooling ignores it), what I would propose, - in case we do want to make this explicit - that every unit we have in the unit file needs a dimension, and that this can exported in vss-tools as well.

We _might_ think about whether later just defining a dimension is enough, and a unit not needed, but I feel today for most practical purposes/tech stacks, you would like to have a  (default) unit as well.

erikbosch:
Some points to discuss tonight:

- Shall we add `iso8601`as unit - so far no one objects
- Shall we remove `label` - so far no one objects
- Shall we define default unit and scaling for each domain - different opinions

SebastianSchildt:
Some general information on "dimensions" and how even thy might need to be combined:

https://www.me.psu.edu/cimbala/Learning/General/units.htm
https://www.sciencetopia.net/physics/dimensional-formula-of-speed

Kostadin-Ivanov:
> * Daniel: We could possibly rename "unit" to "default_unit"
> 
> * Daniel: VSS to define units useful as default for signals, then users can add other as needed

If there will be a default unit, should VSS then specify the set of allowed units for corresponding measurement so user can pick one of these and don't mess up with some other - invalid unit?

For example: to mix speed units (km/h, mph etc) with acceleration (m/s*s) ones?

erikbosch:
> > * Daniel: We could possibly rename "unit" to "default_unit"
> > * Daniel: VSS to define units useful as default for signals, then users can add other as needed
> 
> If there will be a default unit, should VSS then specify the set of allowed units for corresponding measurement so user can pick one of these and don't mess up with some other - invalid unit?
> 
> For example: to mix speed units (km/h, mph etc) with acceleration (m/s*s) ones?

That is in one way at least implicitly defined today in https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml

Our signal `Vehicle.Speed` is today defined with the unit `km/h`. In the unit file one can find that it is defined with domain `speed`, so a downstream project using VSS (and supporting units) could then verify that `Vehicle.Speed` is sent/received/set using either `km/h` or `m/s`. 

```
  km/h:
    label: kilometer per hour
    description: Speed measured in kilometers per hours
    domain: speed
  m/s:
    label: meters per second
    description: Speed measured in meters per second
    domain: speed
```

erikbosch:
I updated the PR to reflect discussions in the VSS meeting. I also changed approach, documenting a possible new syntax in the documentation. It may be easier to read at https://github.com/boschglobal/vehicle_signal_specification/blob/erik_units/docs-gen/content/rule_set/data_entry/data_unit_types.md

There are no longer any changes in the `units.yaml` file, instead I just highlight some problems or discussion topics I see. So if reviewing, focus on the markdown file.

erikbosch:
We have an old idea in #588 - should unit be mandatory? I.e. require `unit: none` to explicitly state that no unit is used?

Kostadin-Ivanov:
I guess that units are mainly for the leaf nodes, i.e. branches or standalone trees, might not need to have a **unit** value.
Unless there is some other scenario / use case?

erikbosch:
> I guess that units are mainly for the leaf nodes (signals, properties, attributes etc.). Branches or standalone trees, might not need to have a **unit** value. Unless there is some other scenario / use case?

Yes, today we only list `unit` as optional keyword/member for attribute/sensor/actuator.

https://covesa.github.io/vehicle_signal_specification/rule_set/branches/
https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/sensor_actuator/
https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/attributes/

Kostadin-Ivanov:
Thank you for the quick response, Erik.

Ah, now, after I read your previous comment:

> We have an old idea in #588 - should unit be mandatory? I.e. require `unit: none` to explicitly state that no unit is used?

I got its point. Namely, to have "unit: none" on each branch / node in the VSS which does not have unit.

Now on the real question :) 
Since [vsstree.py in vss-tools](https://github.com/COVESA/vss-tools/blob/master/vspec/model/vsstree.py) is handling the defaulting of not provided entry for branch **unit** to **none**, maybe it would be good to keep it like this, so to keep VSS less wordy. 

I mean having a mechanism to handle default values in VSS nodes fields would keep the VSS with just right kind of information, which is required to be presented. 
Otherwise, we might end up with many "defaulted" fields under each described branch, signal etc., which will just add more, less significant, content to the specification. Especially to indicate some **none** fields, which might not be that relevant to the described branch / node / signal.

doganulus:
Before seeing this pull request, I have opened this discussion: https://github.com/COVESA/vehicle_signal_specification/discussions/680

Seems related.

nickrbb:
I think it's good to be re-looking at this aspect. However, I wonder if we should perhaps be looking to be more flexible when it comes to units in VSS. Right now, we define a "default" unit, which is mostly (although not always) in metric. However, there is a large proportion of the world that doesn't use metric (or whatever default  is specified) and so the default unit will not suit them. Instead, why don't we extend VSS signal definitions to include measurement units as part of the signal path? For example, for the vehicle.speed signal, we would extend it to include those units specified in the speed domain e.g. vehicle.speed.mph, vehicle.speed.kmh. This way, a VSS "client" can request the unit it prefers and the VSS "server" can take care of any needed conversion and provide the requested signal in the requested unit. If no unit is provided, then maybe the default unit can be used.

To a certain extent (i.e. by making the unit now explicit in signals), it also takes care of the case where a VSS server implementation provides signals in a different unit to that specified as the default in VSS (which is something not prohibited by VSS) but clients don't know this without prior knowledge of the system they are running on.

Kostadin-Ivanov:
With current VSS implementation we have:

Vehicle.Speed.datatype
Vehicle.Speed.type
Vehicle.Speed.unit
Vehicle.Speed.description

which is enough sufficient to build an aspect model for the corresponding signal / measurement. 

If the implementation, goes with:
Vehicle.Speed.mph
Vehicle.Speed.kmh
and discards the Vehicle.Speed.unit, then what will happen if we need to handle: meter per second (ms) or foot per second (fps)?
Will there be required to declare:
Vehicle.Speed.### for each supported unit type?

On my opinion, having default metric units makes the VSS definition consistent. 
The units.yamls file provides the option to the user to load VSS with imperial or other types of units, for corresponding domain. In this case Vehicle.Speed. 
Then, VSS user can build their aspect models with clearly defined units so their models' user will have to conform with the declared interface / contract.

nickrbb:
> With current VSS implementation we have:
> 
> Vehicle.Speed.datatype
> Vehicle.Speed.type
> Vehicle.Speed.unit
> Vehicle.Speed.description

You seem to be mixing paradigms here. We have "vehicle.speed" as the signal name/path. Datatype, Type, Unit, and Description are attributes or metadata for the signal.


> which is enough sufficient to build an aspect model for the corresponding signal / measurement.

Agreed, but that's not the issue I was highlighting.

 
> If the implementation, goes with: Vehicle.Speed.mph Vehicle.Speed.kmh and discards the Vehicle.Speed.unit, then what will happen if we need to handle: meter per second (ms) or foot per second (fps)? Will there be required to declare: Vehicle.Speed.### for each supported unit type?

Not entirely sure I follow you, but there would be a need to specify all allowed units for each signal. But is this any different to adding allowed units into the units.yaml file as we do today? Also, we don't have to be explicit and add them to the signal name, we could just allow any unit to be appended to a signal name that is specified in the associated domain as specified in units.yaml.

 
> On my opinion, having default metric units makes the VSS definition consistent.

Metric is not always the default (see Vehicle.Chassis.Axle.Row1.WheelDiameter, Vehicle.Chassis.Axle.Row1.WheelWidth), and as I previously stated, the default unit we specify in VSS isn't always the default used in certain parts of the real world.


> The units.yamls file provides the option to the user to load VSS with imperial or other types of units, for corresponding domain. In this case Vehicle.Speed. Then, VSS user can build their aspect models with clearly defined units so their models' user will have to conform with the declared interface / contract.

Right, but you're assuming here that VSS users "know" what unit is provided in a response to a request for a signal. My point was that in applications written to be cross-platform/system-independent, how do they know the unit and/or how can they request a unit that they will understand and can process?

Kostadin-Ivanov:
Hello Nick,
thank you for the detailed response.

With regards to:
> You seem to be mixing paradigms here. We have "vehicle.speed" as the signal name/path. Datatype, Type, Unit, and Description are attributes or metadata for the signal.
I understand that datatype, type, unit and description are attributes of the Vehicle.Speed signal. I was just following the suggested paths: Vehicle.Speed.mph and Vehicle.Speed.kmh, which at first reading, I understood that they are supposed to hold a value of the measured speed in mph and kmh respectively. 

I agree on the note that there should be a list with allowed units for the speed domain, as defined in the units.yaml file, which then should be included in the generated VSSNode object, when the VSS tree is expanded (parsed) for further use. 

In relation to the default units - being either metrics or imperial - I'd say that in order to keep consistent specification, VSS should be defined in either of these default units systems. When user expands them, it will be up to the user to switch to other than the default units system.
Personally, not that I am used to or I am the expert to debate which system is better, but the metric one seems to be the better default candidate to me. These irregular relations like 12 inches for a foot, 3 feet for a yard are not my fitting in my calculator :) 
Anyway, these are just my thoughts and as I said I am not the expert to say which system is better.

And, on the last topic about:
> Right, but you're assuming here that VSS users "know" what unit is provided in a response to a request for a signal. My point was that in applications written to be cross-platform/system-independent, how do they know the unit and/or how can they request a unit that they will understand and can process?
Sorry that I've missed one of "The Four Agreements", which would perfectly fit in many different areas, including software engineering. Namely: "don't make assumptions" :) 
I checked some of the APIs which we generate, using VSS, and Speed was simply provided as a **number** without any notice of the related unit, which is definitely a miss in the API as you mentioned.

As per above, maybe it would worth to re-discuss design and modeling of such VSS signals / nodes, where unit type is kind of "a must to be included in the API".

Maybe if the Speed or similar signal is provided in the API as:
```
Speed: {
   value: 50.5,
   unit: "kmh"
}
```
then users and providers can easy declare the system (metric, imperial) in which they require / provide corresponding signal.
In order to achieve this, I guess that the Vehicle.Speed might not just declared as a signal, but rather as a branch, which has an object of Speed with related unit etc. 
This here is just my guessing so I am not 100% if this would be best way to describe a signal like Speed it in VSS as an object with value and unit, which then can result in an API entity like the one above.

erikbosch:
Concerning the metric vs. imperial topic. The exceptions we have today in VSS concerns the few areas where inches are still used in Europe - for describing wheel/tire/rim sizes. There it is typically useless to describe the wheel/rim diameter and width in millimeters, unless you happen to use [Michelin TRX](https://en.wikipedia.org/wiki/Michelin_TRX) wheels. Could be useful for calculation purposes to use a different unit, but not if you are to buy new tires and need to tell the garage what size you need.

Kostadin-Ivanov:
This makes sense.

However, there might still be needed some sort of keeping on tract the unit of provided measurement / requested signal. Further more, the API provider should declare somehow the unit of returned value. Otherwise there might be some confusions on the client side and vice versa.

For stateful API communications, it might be possible to exchange units and other static information on the contract part, but for stateless communications, it would more likely be needed to provide the unit along with the value.

In terms of reads, maybe is a bit easier, because user can specify the required unit for corresponding signal, and then service provider will do corresponding conversions. However the service provider (backend) will still need to know the unit of requested signal value.

erikbosch:
I rebased the PR and did a small update - I added a `deprecated` keyword. I think this could be useful if we in VSS_project wants to remove a unit or rename a unit. Like if we in the VSS-project would agree that `kph` is a better identifier than `km/h`. Then we could create a new unit `kph` and mark the old `km/h` as deprecated until next major release, or longer if needed. That would give signal catalogs reasonable time to be updated before the old unit is removed. It could also give downstream projects a reasonable timeframe to protest if VSS-project wants to remove a unit because we think it is not used, but there actually are users. A typically example could be the unit `stars`.

Anyhow - this is my conclusion of the discussion so far:

- Some would like VSS to just specify "domain" for signals - Client/Servers could attach unit-information when they send values.
- On the other hand - some existing APIs like VISS and KUKSA gRPC does not support sending unit-information. This may change in the future, but for now we must have a "default" unit for each signal as you otherwise cannot interpret data you get with VISS and KUKSA gRPC


So for now a slightly refactored unit-file (compared to how it looks like [today](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml)) with more well defined domains/quantities seems to be a step in the right direction, regardless of what path we want to go later. So I would like to get feedback on the proposed unit file syntax in the `*.md` file. If there is something you see as a step in the wrong direction. There have been ideas to add conversion factors or relationship to SI units, but that could be added later when we see a need for it.

Kostadin-Ivanov:
Hello Erik,
I agree on the deprecated keyword.
I am not sure how exactly this will be provided and how it should be handled. I just have few questions:

1. Would this lead to some kind of a user notification when VSS tree gets expanded / parsed or users will have to handle the **deprecated** keyword on their own?
2. If there will be user notification about **deprecated** entry, field, unit etc. would it be possible to have some additional information?
2.1 As per example with `km/h` being replaced by `kph` - maybe a warning saying that: 
      _"Unit 'km/h' is deprecated and will be replaced by 'kph'"_
      or something similar?

Other than this, the rest sounds OK to me.

erikbosch:
I refactored the deprecation part to make it more similar to the [corresponding keyword for signals](https://covesa.github.io/vehicle_signal_specification/rule_set/basics/#deprecation-since-version-21)  where we propose to suggest information like

`deprecation: V2.1 moved to Vehicle.CurrentLocation`

... then it would be quite easy in code to have something like:

```
if self.unit.deprecation != "":
  print "Warning: Signal {self.name} use deprecated unit {self.unit}: {self.unit.deprecation}"
```

doganulus:
This may be a stylistic choice, but what about two separate `quantities.yml` and `units.yml` without top-level `quantities` and `units` keys?

This means:
1) Flatter file structure
2) Saving one-level access when the file is loaded into a variable.

Unless we want one `config.yml` to rule them all.

erikbosch:
> This may be a stylistic choice, but what about two separate `quantities.yml` and `units.yml` without top-level `quantities` and `units` keys?
> 
> This means:
> 
>     1. Flatter file structure
> 
>     2. Saving one-level access when the file is loaded into a variable.
> 
> 
> Unless we want one `config.yml` to rule them all.

Works fine for me.

erikbosch:
See also #680

erikbosch:
> I certainly don't want to hold up this PR, but I think it would be good to continue the discussion on the possible enhancement of specifying unit types using some kind of branch/sensor hybrid solution. Should I start a new Issue?

Feel free to create a new issue, then we can see if it fits here or somewhere else

erikbosch:
Based on the discussion so far I have tried to come up with something "doable" and I have refactored the PR accordingly. It now consist of:

- A slightly changed format for the unit file including refactoring of current file
- A format for domain file including an initial version
- Introduction of iso8601 as unit
- Some minor change of domain names.

The changes would require changes in tools, I have a draft at https://github.com/COVESA/vss-tools/pull/307 which can handle both the old and new unit format. It does not add any new functionality like consuming the domain file. So that this PR fails in CI right now is expected, if we "approve" the new syntax we shall first merge the vss-tools PR, then update submodule link.

My suggestion when you start reviewing is to focus on structural parts, actual definition of "domains" is not that important now, that we can improve later. It is more important to agree on that we keep the term "domain" and that it can cover all types of "domains" that we have as of today.

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
-Nm            | Torque                          | Torque measured in newton meters
-rpm           | Rotational Speed                | Rotational speed measured in revolutions per minute
-Hz            | Frequency                       | Frequency measured in hertz
-ratio         | Relation                        | Relation measured as ratio
-percent       | Relation                        | Relation measured in percent
-... | ... | ...
-
-[VSS-Tools](https://github.com/COVESA/vss-tools) require that a unit file is available when transforming *.vspec files.
-It can be specified by the `-u` parameter, and if not given the tools will search for a file `units.yaml`
+The VSS syntax does not in itself specify what units that can be used, the unit attribute as declared for signals in *.vspec files is optional and can contain an arbitrary string value.
```
ppb2020, line 77:
```suggestion
The VSS syntax does not in itself specify what units can be used, the unit attribute as declared for signals in *.vspec files is optional and can contain an arbitrary string value.
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+[VSS-Tools](https://github.com/COVESA/vss-tools) however require that all units used are defined.
+Units are defined by including them in an a unit file with syntax as described below.
```
ppb2020, line 79:
```suggestion
Units are defined by including them in a unit file with syntax as described below.
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+One or more unit files can be specified by the `-u` parameter, and if not given the tools will search for a file `units.yaml`
```
ppb2020, line 80:
```suggestion
One or more unit files can be specified by the `-u` parameter and, if not given, the tools search for a file `units.yaml`
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
 in the same directory as the root *.vspec file.
 
-The VSS standard catalog is based on [units.yaml](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml),
-i.e. only units from that file can be used.
+For the VSS standard catalog the VSS-project has defined a set of units that can be used for signals in the VSS standard catalog.
+This list is composed with definitions according to International Units (SI) and few automotive specific units:
```
ppb2020, line 84:
```suggestion
This list is composed of definitions according to International Units (SI) and few automotive-specific units:
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+[Specification](https://www.iso.org/standard/30669.html), [Wikipedia](https://en.wikipedia.org/wiki/International_System_of_Units)
+
+The VSS list of units for the standard catalog exists in [units.yaml](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml).
+
+
+## Unit file syntax
+
+Unit files have the syntax defined below:
```
ppb2020, line 92:
```suggestion
Unit files follow the syntax defined below:
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+    [unit: <string>] # Full name of unit, optional, if not given assumed to be equal to vss-unit-identifier
+    domain: <string> # Domain of the unit.
+    [allowed-datatypes] : ['numeric', 'string', uint8', ...]] # Allowed datatypes in VSS standard catalog
+    [deprecation: <reason>]
+]*
+```
+
+The VSS term `domain` corresponds roughly to the term `quantity` as defined in for example ISO 80000.
+Typical example are `length`, `mass` and `velocity` that all can be expressed in SI-units.
+Domains do however not from a VSS perspective need to correspond to physical quantities.
```
ppb2020, line 107:
```suggestion
However, from a VSS perspective domains do not need to correspond to physical quantities.
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+It could be an arbitrary term, but it is in general expected that it is possible to convert between values
```
ppb2020, line 108:
```suggestion
It could be an arbitrary term, but it is generally expected that it is possible to convert between values
```

erikbosch:
Meeting notes:

- Please review, may be merged next week if no comments
- Sebastian: I prefer "quantity" rather than "domain"

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+  allowed-datatypes: ['numeric']
+
+mm:
+  definition: Length measured in millimeters
+  unit: millimeter
+  domain: length
+  allowed-datatypes: ['numeric']
+```
+
+As `m` and `mm` are defined with the same domain it is expected that you can convert a value from `m`to `mm`.
```
ppb2020, line 127:
```suggestion
As `m` and `mm` are defined with the same domain it is expected that you can convert a value from `m` to `mm`.
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+
+The `allowed-datatypes` attributes can be used to specify which VSS datatypes that can contain values of a particular unit.
```
ppb2020, line 129:
```suggestion
The `allowed-datatypes` attributes can be used to specify which VSS datatypes can contain values of a particular unit.
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+For most units the symbolic datatype `numeric` meaning any float or integer datatype can be used.
+As an example, both `uint8` and `float` can be used to represent a length value.
+If using `uint8` you will have range restrictions, but that might be acceptable for some signals.
+For some units more specific datatype restrictions are relevant. Some examples:
+
+* A date/time expressed in ISO 8061 format can only be represented in string
```
ppb2020, line 135:
```suggestion
* A date/time expressed in ISO 8061 format can only be represented as a string
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+* A UNIX Timestamp signal must be at least 32 bit unsigned to be able to handle date/time after year 2038.
+
+The `deprecation` keyword can be used to indicate that a specific unit may be removed in the future,
```
ppb2020, line 138:
```suggestion
The `deprecation` keyword can be used to indicate that a specific unit may be removed in the future.
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+Tooling shall preferably give a warning if a signal use a deprecated unit or the unit used belongs to a deprecated quantity. The reason should preferably list when and why the unit is deprecated, a hypothetical example is given below:
```
ppb2020, line 139:
```suggestion
Tooling shall preferably give a warning if a signal uses a deprecated unit or the unit used belongs to a deprecated quantity. The reason should preferably list when and why the unit is deprecated, a hypothetical example is given below:
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+
+```
+
+  inch:
+    definition: Distance measured in inches
+    unit: inch
+    domain: distance
+    deprecation: V5.0 replaced with in
```
ppb2020, line 147:
```suggestion
    deprecation: V5.0 replaced by 'in'
```

docs-gen/content/rule_set/data_entry/data_unit_types.md
```diff
@@ -74,67 +74,132 @@ Vehicle.Powertrain.Transmission.ClutchWear:
...
+
+
+The VSS list of domains for the standard catalog exists in [domains.yaml](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/domains.yaml).
+
+## Defining custom units
+
+It is possible to define custom units in a unit file.
+Assume for instance you want to have a signal showing remaining range in [furlong](https://en.wikipedia.org/wiki/Furlong).
+Then you could add an additional unit `furlong`. No need to specify `unit` or `symbol` as they equals the default (i.e. "furlong").
+As this is unit not commonly used and not described in any standards it might be relevant to describe how it can be converted to other units.
```
ppb2020, line 195:
```suggestion
As this is unit not commonly used and not described in any standards, it might be relevant to describe how it can be converted to other units.
```

spec/domains.yaml
```diff
@@ -0,0 +1,77 @@
...
+  definition: TBD
+power:
+  definition: Scalar product of force F acting to a body and its velocity (ISO 80000-4:2019)
+work:
+  definition: Process quantity describing the total action of a force F along a continuous curve
+              in three-dimensional space with infinitesimal displacement dr,
+              as a line integral of their scalar product (ISO 80000-4:2019)
+mass:
+  definition: Property of a body which expresses itself in terms of inertia
+              with regard to changes in its state of motion as well as
```
ppb2020, line 26:
```suggestion
              related to changes in its state of motion as well as
```

ppb2020 (commented):
Looks good. A few suggested textual changes.

ppb2020:
As to 'quantity' vs. 'domain', I'd go with a standard usage of the term. For 'quantity', I found:

https://www.iso.org/obp/ui/#iso:std:iso:80000:-1:ed-1:v1:en
https://en.wikipedia.org/wiki/International_System_of_Quantities

Hence, I think quantity would be the correct term to use here.

erikbosch:
PR updated to use "quantity" instead of "domain". The companion PR https://github.com/COVESA/vss-tools/pull/307 updated as well. There are some CI check failures at the moment, partially caused by the dependency to vss-tools. They will be fixed before merging.

PR #671, opened 2023-10-19
For discussion: Add support for definition

erikbosch:
At the AMM we briefly discussed "description" vs "definition". I had a mail conversation with Alan to get some more input, see response below. My conclusions based on the answer and the discussion at the AMM:

- So far all seem to prefer "definition" over "description"
- No-one seem to have a need for both
- But "migrating" from "description to definition is not simple as we have many signals

Based on this I propose:

- We add support for `definition` as an optional member of a signal declaration. That way we can start adding definitions if we like.
- When (if ever) we have `definition` for most signals we can decide on time plan to make definition mandatory and description optional
- Add `definition` as an allowed/expected attribute in vss-tools, but for now no other explicit changes. this means `definition` will be included in Yaml/JSOn output, but ignored for other outputs

This PR propose a possible change for documentation. If we agree to do the change a corresponding change in vss-tools must be implemented (low effort). This PR also include some unrelated changes to make sure documentation and actual behavior in vss-tools match.

**Info received from Alan**

Hi Erik,

The important distinction between definition and description is that a definition is a more formal specification that includes the necessary and sufficient conditions.  It's the king of specification you find in a dictionary.  To state the obvious, that provides unambiguous semantics.  But the real reason it's critical is that those necessary and sufficient conditions translate to formal axioms in an ontology and can be used for inferencing and automation.

A description, on the other hand, provides any kind of information that helps humans conceptualize the entry.  It can include examples, physical properties, etc.  

If I'm defining human being, then I can say that it's a culture-bearing primate classified in the genus Homo, especially the species H. sapien.  It meets the necessary conditions because there can be no human being who is not a culture-bearing primate classified in the genus Homo.  It includes the sufficient conditions because every a culture-bearing primate classified in the genus Homo is a human being.

If I'm describing human being, then I can say it has superior mental development, power of articulate speech, and upright stance.  However, we are finding that bonobos and even some monkeys can fit that description.  So it's not a definition.

Typically, we use definitions for reference data - data whose consistency, semantic precision, and operational impact require centralized curation prior to use.  We use descriptions for master data and transactions - data for which changes in state are managed as a routine course of business.  I would classify the objects, properties, and signals in VSS as the former.

In sum, it's not just changing the name but ensuring that the content contains the necessary and sufficient conditions.

### DIFF
--- a/docs-gen/content/rule_set/data_entry/allowed.md
+++ b/docs-gen/content/rule_set/data_entry/allowed.md
@@ -20,8 +20,6 @@ SteeringWheel.Position:
 
 ```
 
-If `allowed` is set, `min` or `max` cannot be defined.
-
 The `allowed` element is an array of values, all of which must be specified
 in a list.  Only values can be assigned to the data entry, which are
 specified in this list.
--- a/docs-gen/content/rule_set/data_entry/sensor_actuator.md
+++ b/docs-gen/content/rule_set/data_entry/sensor_actuator.md
@@ -14,17 +14,32 @@ entry example is given below:
 ```YAML
 Speed:
   type: sensor
-  description: The vehicle speed.
+  definition: Vehicle speed relative to the X-axis of the intermediate axis system as defined by ISO 8855 section 2.13.
+  description: Vehicle speed relative to road surface. Positive if vehicle is moving forward, negative if vehicle is moving backward.
   comment: For engine speed see Vehicle.Powertrain.CombustionEngine.Engine.Speed.
   datatype: float
   unit: km/h
   min: 0
   max: 300
+  allowed: ...
 ```
 
-**```Drivetrain.Transmission.Speed```**
-Defines the dot-notated name of the data entry. Please note that
-all parental branches included in the name must be defined as well.
+**```<signal name>```**
+Defines (parts of) the name of the data entry.
+The full name/path of a VSS signal, considering parent branches, must be unique,
+but the name specified in the **.vspec* file do not necessarily need to be unique.
+
+The VSS **.vspec* format supports [#include](/vehicle_signal_specification/rule_set/includes/) directives.
+When including a **.vspec* file it is possible to define a branch prefix that shall be used on entries in the
+included file. If the definition above exists in a signal included like:
+
+```YAML
+#include Vehicle/Vehicle.vspec Vehicle
+```
+
+... then the full name of the signal using VSS dot-notation will be `Vehicle.Speed`
+
+*Note: VSS implementations does not necessarily need to use VSS dot-notation to reference signals, they can use alternative mechanisms to uniquely reference signals!*
 
 **```type```**
 Defines the type of the node. This can be `branch`,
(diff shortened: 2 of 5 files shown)


docs-gen/content/rule_set/data_entry/allowed.md
```diff
@@ -20,8 +20,6 @@ SteeringWheel.Position:
 
 ```
 
-If `allowed` is set, `min` or `max` cannot be defined.
```
erikbosch, line 23: This is not enforced in vss-tools. In standard catalog as of today we only use `allowed` for strings, and there min/max does not make sense, but nothing that forbids it.

docs-gen/content/rule_set/data_entry/sensor_actuator.md
```diff
@@ -14,17 +14,32 @@ entry example is given below:
 ```YAML
 Speed:
   type: sensor
-  description: The vehicle speed.
+  definition: Vehicle speed relative to the X-axis of the intermediate axis system as defined by ISO 8855 section 2.13.
```
erikbosch, line 17: This is a good exercise, trying to find a definition for vehicle speed. I looked a bit in ISO 8855 trying to find correct words, but the result is not that intuitive. it tries to capture:

- If the vehicle is sliding sideways down an icy hill speed shall be 0, as we only are interested in the "forward" component
- We are interested in the speed relative to the surface, i.e. it shall be 0 if the vehicle is stuck with spinning wheels in mud
- We are not interested in the speed relative to vehicle x-axis, if the rear of the vehicle is heavily loaded there might be small difference between speed mentioned along vehicle x-axis and the intermediate x-axis

But I am no vocabulary expert, this just shows that it isn't that easy. Then we can also discuss if we want to specify what a signal ideally should contain, or rather what it is allowed to contain. Most likely we consider it OK that speed is reported wrong if one or more wheels are stuck in mud.

docs-gen/content/rule_set/data_entry/allowed.md
ppb2020: In the same manner that "default" only applies to attributes, but nothing forbids assigning a default to other signals.

docs-gen/content/rule_set/data_entry/sensor_actuator.md
ppb2020: Indeed, this isn't easy. A vehicle on the back of a two truck clearly has a speed as would be reflected by GPS, but I would bet most car will report a Vehicle.Speed of 0 in that case. That leads to thoughts I have been having on the need to have an accurate description of each signal, where in this specific case of Vehicle.Speed the description would be something like "speed of the vehicle when ignition is turned on and the vehicle is self-propelled" with a comment such as "refer to GPS speed" for a vehicle speed in situation where the actual movement of the vehicle is of interest (such as when being towed or slipping down an icy hill with brakes fully on)".

erikbosch:
Meeting notes:

- Adnan: Uncertain if we are able to provide accurate definitions
- Daniel: As a consequence of property + object introduction we could possibly consult official dictionary
- Daniel: In VSS signals are both property and object, defining that combination is tricky, easier if you have separate property and object
- Erik: Conclusion seems to be that not much value in adding definition as keyword, as we likely will not add actual definitions. Could be relevant if we do changes to have property+object. Put on hold until someone wants to contribute definitions?
- Adnan: Does Ford have definitions they would be willing to contribute?
- Paul: Will check
- Pleas review/discuss

erikbosch:
Closing this one for now as there does not seem to be much support for it

PR #673, opened 2023-10-24
Support negative torque for combustion engine

ppb2020:
Some combustion engine vehicles report a negative torque under braking conditions. This cannot be properly reflected by the current definition of engine torque as it is an unsigned integer.

This change proposes modifying the combustion engine torque definition from an unsigned integer to an integer.

Signed-off-by: Pierre Pierre Blais ppblais@BlackBerry.com

### DIFF
--- a/spec/Powertrain/CombustionEngine.vspec
+++ b/spec/Powertrain/CombustionEngine.vspec
@@ -223,12 +223,11 @@ Power:
 # Current Torque
 #
 Torque:
-  datatype: uint16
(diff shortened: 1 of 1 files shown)


erikbosch:
It is still the DCO checks that fail as it expects something like `Signed-off-by: Erik Jaegervall <erik.jaegervall@se.bosch.com>` to be part of the commit message (i.e. not the pull request message)

erikbosch (approved):

PR #676, opened 2023-10-26
Align Temperature Signals

erikbosch:
As of today we have a mix of float and int16 for temperature signals, with no real rationale for the difference. This PR changes the int16 signals to instead use float, so after this change all signals with Celisus unit use float.

Only intended for VSS 5.0 as this is a backward incompatible change.


ppb2020 (approved):
Looks good. Approved.

PR #688, opened 2023-11-14
Align usage of the word type.

### LINKED ISSUE #683: Types and datatypes
I would like to share my observations about VSS types and datatypes. The current classification is as follows:

```yaml
type: [sensor, actuator, attribute, property, struct, branch]
datatype: [boolean, string, numbers, arrays]
```

I have two questions here:
- First, the difference between types `attribute` and `property` is unclear to me. Let me try to summarize my understanding below. Is that right?
  - `sensor` denotes a dynamic (temporal) signal not directly controllable by the system -- fast change
  - `actuator` denotes a dynamic (temporal) signal directly controllable by the system -- fast change
  - `attribute` denotes a dynamic? configuration variable -- thus, runtime-changeable but not often.
  - `property` denotes a static? configuration parameter? -- runtime-constant

- Second, I understand that `struct` is a relatively new keyword for the VSS project. My initial assessment would place the structural elements of `struct` and `branch` under `datatype` as I could not find a functional description for `struct` as above. The same is true for `branch` where the difference is that structs are packed, and branches are not. Therefore, what would say about the following classification that moves them under datatypes? 

```yaml
type: [sensor, actuator, attribute, property]
datatype: [boolean, string, numbers, arrays, struct, branch]
```

Even more, taking full liberty to ignore all necessary backward compatibility issues (probably the project cannot do very easily), I would prefer the key `type` instead of `datatype` for compatibility with other specification formats, mainly JSON Schema, and prefer another term, like `kind` or `class`, to denote its usage in the system.

```yaml
kind: [sensor, actuator, attribute, property]
type: [null, boolean, string, numbers, tuple, list, struct, branch]
```

And, a straightforward mapping to JSON types may be valuable as I would map `tuple` and `list` to JSON arrays as well as `struct` and `branch` to JSON objects. But I admit I don't have much prior experience with other formats that VSS can be converted.

erikbosch:
This PR intends align terms used in documentation

* If we refer to a VSS data type, the term "datatype" is used
* If we refer to the VSS node attribute "type" that specifies if a node is a branch, sensor, actuator, struct, ... we use the term "type"

Fixes #683

### DIFF
--- a/docs-gen/content/rule_set/data_entry/allowed.md
+++ b/docs-gen/content/rule_set/data_entry/allowed.md
@@ -26,8 +26,8 @@ The `allowed` element is an array of values, all of which must be specified
 in a list.  Only values can be assigned to the data entry, which are
 specified in this list.
 
-The `datatype` specifier gives the type of the individual elements of the `allowed`
-list. In general `allowed:` is valid for all datatypes, including integer- and float-based types, unless othwerise specified.
+The `datatype` specifier gives the datatype of the individual elements of the `allowed`
+list. In general `allowed:` is valid for all datatypes, including integer- and float-based datatypes, unless othwerise specified.
 
 ## Recommendation on String values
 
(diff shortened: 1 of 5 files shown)


ppb2020 (approved):
Good cleanup. Approved.

erikbosch:
Meeting notes: Please review

Issue #642, opened 2023-09-11
Need for change of instance handling


erikbosch:
Now and then we discuss if the current instance solution is sufficient. We have some issues and pull requests currently opened:

#564 suggests to be able to have instances on individual signals
#375 suggests a more flexible naming mechanism 
https://github.com/COVESA/vss-tools/pull/293 has a solution to produce unexpanded output

Instances sometimes also comes up as a topic when discussing ontologies. The intention with this issue is to discuss if we see any changes required in the VSS-syntax to better support ontologies and Vehicle APIs. some ideas follow below:

## Idea on how to change syntax

I am thinking if we should change to a different method to represent instances. As I see it instance declarations like stating number of rows in standard tree has limited value as it is vehicle specific and even may change for a specific vehicle, like if you remove some seats from your minivan. That VSS standard catalog states that there shall be two rows of seats with 3 seats in each does not really make sense. An alternative approach would be that VSS just states that a seat has two instance-related const attributes, row and position, of integer and string types. Which instances that exist is deployment related - you could either give it in VSS-specified syntax as a customization or overlay, or let the server handle it dynamically. 

So instead of stating that there IS 2 rows we maybe just state that there is a row attribute and a Pos attribute. For Row we do not state any limits, for Pos we give allowed values. This is what a client needs to know to be able to create correct requests. We could even support to both have a Left/Right-approach as well as Driver/Passenger-approach and let the server manage the mapping.

Another reason for changing syntax for defining instances is that the current syntax is not that well defined, vss-tools use patterns to guess if a range or a list is given, so you cannot for example combined list and prefix.

```
Vehicle.Seat:
  type: branch
  description: Branch A.AA.
  instance:
    definition:
        - id: 'Row' # Id, would typically be used as argument name in API methods, like "getAA(row: uint8, pos: string)"
          datatype: uint8 # Defining type
          prefix: 'Row' # Prefix to use in extended names, like "Row1"
        - id: 'Pos'
          datatype: string
          allowed: ['LEFT','RIGHT','CENTER','DRIVER_SIDE', 'PASSENGER_SIDE']
```

[...]

PR #695, opened 2023-11-30
Instance discussion - Driver/Passenger vs Left/Right

erikbosch:
There have been some criticism on the VSS change some time ago to use DriverSide/PassengerSide as reference rather than Left/Right. I can see values in supporting both notations, or that at least OEMs can decide which notation they prefer.

This PR discuss/propose a possible solution. In short it is based on having something like this in VSS:

```
Seat:
  type: branch
  instances:
    - Row[1,2]
    - ["DriverSide","Middle","PassengerSide", "Left","Right"]
```

We can also use this PR to capture the discussion/decision on whether we actually need to do something.

Some previous discussion available in #642

### DIFF
--- a/instances.md
+++ b/instances.md
@@ -1,0 +1,335 @@
+# Handling of instances - the problem and possible solutions
+
+## Introduction
+
+* Do we see a need to support both DriverSide/PassengerSide and Left/Right notation?
+* If so, do we want signals in tree to be duplicated, i.e. both `Vehicle.Seat.Row1.Left.Position` and
+  `Vehicle.Seat.Row1.DriverSide.Position` should exist?
+* Or rather a mechanism that to indicate that for example `Vehicle.Seat.Row1.DriverSide` for this particular vehicle
+  is a synonym for `Vehicle.Seat.Row1.Left``
+
+This file show some possibilities to solve the problem.
+
+A technically feasible solution (IF we want to change) could be:
+
+* First just expand available names so you can reference by both DriverSide/PassengerSide and Left/Right notation.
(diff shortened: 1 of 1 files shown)


alanmapleburl-au:
Here is a perspective from an ontologist.  TL/DR: I support - ["DriverSide","Middle","PassengerSide", "Left","Right"].  It uses anyOf logic and provides the knowledge required rather than using oneOf logic to constrain it.

In Basic Formal Ontology, which is a philosophically rigorous upper ontology, The lateral or y-axis position (Left, Middle, and Right) would be deemed "qualities."  They are only dependent on the thing they describe, such as the seat position.  They will always be able to describe the seat position regardless of processes or circumstances. 

Driver and Passenger, on the other hand, would be deemed "roles."  They not only depend on the thing that they describe, but also on some process--driving a vehicle in this case.  The problem becomes clear when you consider a fully autonomous vehicle in which there are passenger seats but no driver position.  

In a fully-articulated model, the position of the driver controls would be a property of the vehicle, not a seat.  Everything necessary could be reported by the vehicle.driverPosition and seat.position.  Given the tree model and the notion of allowed values, if we're going to constrain with oneOf, then [Left, Middle, Right] is the stronger option.  But I see no reason that we can't have it both ways by including both the lateral position and the occupant role.

erikbosch:
My conclusion so far:

- No-one really seems to object supporting both left/right and driverside/passengerside
- The question is if we need to add syntax to VSS so that implementations (like VISS server, Kuksa) automatically knows that they are synonyms, or if we leave that to implementations

Maybe I will create a PR for a "simple" solution

erikbosch:
Closing this one for now as there likely will be no progress within the next months

PR #686, opened 2023-11-09
Update screen mirroring in Infotainment.vspec

spshin3:
Add screen mirroring in infortainment.

Mirror phone screen to car screen.

PR #699, opened 2023-12-13
Update screen mirroring in Infotainment.vspec

spshin3:
Resolve "erik's comment" and a DCO problem from PR #686 .

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -267,4 +267,27 @@ SmartphoneProjection.SupportedMode:
   allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
 
+SmartphoneScreenMirroring:
+  type: branch
+  description: All smartphone mirroring actions.
+
+SmartphoneScreenMirroring.Active:
+  datatype: string
+  type: actuator
+  allowed: [ 'NONE', 'ACTIVE', 'INACTIVE' ]
+  description: Mirroring activation info.
+  comment: NONE indicates that mirroring is not supported.
+
+SmartphoneScreenMirroring.Source:
+  datatype: string
+  type: actuator
+  allowed: ['USB', 'BLUETOOTH', 'WIFI']
+  description: Connectivity source selected for mirroring.
+
+SmartphoneScreenMirroring.Resolution:
+  datatype: string
+  type: attribute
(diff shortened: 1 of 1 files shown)


erikbosch:
This PR addresses screen mirroring, we have since before smartphone projection, see below.

I am no expert in this area, but what is the difference? I.e. what use-cases and technologies are supposed to be covered by the screen mirroring signals? Just want to make sure that we do not duplicate functionality. Is it intended for Chromecast and similar "one way" technologies? (Android Auto, Apple Carplay and MirrorLink already mentioned in projection signals).

If it concerns different but similar technologies that cannot be used in parallel we may continue just extending the `SmartphoneProjection` branch as needed


```

SmartphoneProjection:
  type: branch
  description: All smartphone projection actions.

SmartphoneProjection.Active:
  datatype: string
  type: actuator
  allowed: [ 'NONE', 'ACTIVE',  'INACTIVE' ]
  description: Projection activation info.
  comment: NONE indicates that projection is not supported.

SmartphoneProjection.Source:
  datatype: string
  type: actuator
  allowed: ['USB', 'BLUETOOTH', 'WIFI']
  description: Connectivity source selected for projection.

SmartphoneProjection.SupportedMode:
  datatype: string[]
  type: attribute
```

spec/Cabin/Infotainment.vspec
```diff
@@ -267,4 +267,27 @@ SmartphoneProjection.SupportedMode:
...
+
+SmartphoneScreenMirroring.Source:
+  datatype: string
+  type: actuator
+  allowed: ['USB', 'BLUETOOTH', 'WIFI']
+  description: Connectivity source selected for mirroring.
+
+SmartphoneScreenMirroring.Resolution:
+  datatype: string
+  type: attribute
```
erikbosch, line 289: Is this the physical resolution offered by the HMI, i.e. static, or is it the current resolution as agreed by receiver (Infotainment) and sending device? 

One can in general discuss if the ScreenMirroring/SmartphoneProjection signals shall be sensor or actuators. I assume the actual setup/connection/pairing not will be managed by VSS so that VSS only is used for showing current status, and then all could be "sensors", but I have no problems calling them "actuators" either, in the end it all depends on implementation. But if this signal represents current resolution rather than static/constant display resolution the it should be "sensor" rather than "attribute".

erikbosch:
@spshin3 - I converted this PR to a draft as I see that some rework is needed. For example calrifying the different between "SmartphoneScreenMirroring" and "SmartphoneProjection". Will you have time to continue work on this PR in the next weeks?

Issue #635, opened 2023-07-31
Deprecation and Removal of OBD branch?


erikbosch:
A topic that comes up now and then is removal of the [OBD](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/OBD/OBD.vspec) branch. Sometimes the rationale is that COVESA should not care about diagnostics data at all, sometimes that there is no reason to duplicate signals, if we need signals concerning for examples oxygen sensors and trouble codes they should exist in the "normal" tree.

This issue intends to collect feedback on the topic as input for decision. Is there someone that is strongly against removing the OBD branch? If so, why? Could it be mitigated by for example adding new signals in other trees?

To start the discussion - would this be a feasible way forward?

- In VSS 5.0 we mark the OBD branch as deprecated
- In VSS 6.0 we remove it.
- Until VSS 5.0 (or latest at VSS 6.0) everyone that wants/needs information that currently exist only in OBD-branch must create corresponding signals in some other branch.

ppb2020:
Seems like a good plan to me.

nickrbb:
I agree, removing duplication of signals is a good idea. I would suggest we make efforts to port all of the signals though, as VSS has a much wider consumer base than it does active contributor base. It would also make VSS a much better alternative to OBD (since it would host a superset of OBD signals).

Furthermore, it might also be a worthwhile effort though to provide a mapping in some form or another (possibly just a readmeOBD.md file) of OBD signals to their VSS equivalents, at least in v6, to help e.g. OBD experts who are also VSS newbies. This could prevent people bringing in contributions later on introducing a specific OBD branch and signal that already exists in VSS but which is not so easy to find/map.

SebastianSchildt:
It might make sense, to still provide OBD as an "offical" overlay, as it is still in widespread use, also for simple telemetry applications.

Other than that I'd think that maybe DTC codes should be somehow migrated, as those are really the basis for many "first-shot" diagnostics, but those might be split into the different branches.

adobekan:
> It might make sense, to still provide OBD as an "offical" overlay, as it is still in widespread use, also for simple telemetry applications.
> 
> Other than that I'd think that maybe DTC codes should be somehow migrated, as those are really the basis for many "first-shot" diagnostics, but those might be split into the different branches.

+1 We should definitely move them to an overlay.

SebastianSchildt:
Meeting 8/23

[...]

PR #701, opened 2024-01-04
Deprecating OBD-branch

erikbosch:
This is a follow up to https://github.com/COVESA/vehicle_signal_specification/issues/635. Intention is to have them deprecated and remove them first in VSS 6.0 Until then anyone can create PRs with replacement signals as needed. This PR contains replacement signals for diagnostics.

As a first point we should we discuss if this is a good approach for removing the OBD branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -197,4 +197,30 @@ A new version of vss-tools is used, with support for static IDs and jsconschema.
 
 ## Planned Changes VSS 5.0
 
--
+### OBD Branch deprecated
+
+The `Vehicle.OBD` branch is now deprecated. The plan is to remove it in VSS 6.0.
+The background is a decision that VSS standard catalog shall not contain a one-to-one representation of the OBD standard.
+Instead, VSS standard catalog may contain corresponding information elsewhere. As an example, instead of
+`Vehicle.OBD.EngineSpeed` (PID `0C`), the VSS signal `Vehicle.Powertrain.CombustionEngine.Speed` can be used.
+Note that not all signals in `Vehicle.OBD` has "duplicates", especially in the area of combustion engine control
+(like Oxygen sensor lambda and voltage readings) VSS currently does not have any counterparts in other parts of the tree.
+
+
+What to do if you as of today use signals from the OBD file
+
+* Check if any suitable replacement signal exist
+  Example: if you use `Vehicle.OBD.Speed`, consider using `Vehicle.Speed` instead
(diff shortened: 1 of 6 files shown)


ppb2020:
IIRC, our reasoning for deprecating OBD was that the signals were generally duplicates of other signals. I would have expected that some signals are not duplicated and would need to be added elsewhere in the catalog as part of the removal. I don't see other signals added...

I would also have expected that some ODB-specific signals might remain, for example a signal indicating that a device is connected to the OBD port. Or one where some "commands" can be sent in specifically to the OBD using certain pins on the diagnostic connector, which could be represented as signals specific to OBD... I don't know if OBD specifies this, however.

erikbosch:
> IIRC, our reasoning for deprecating OBD was that the signals were generally duplicates of other signals. I would have expected that some signals are not duplicated and would need to be added elsewhere in the catalog as part of the removal. I don't see other signals added...
> 
In this PR I add signals for DTCCount and DTCList as we do not have DTCs elsewhere and several people have said that they could be useful. But you are correct, there are more signals that is not represented in VSS, especially in the area of combustion engine status (fuel trim, timing advance, lambda sensor readings and so on). But the question is if those signals are useful in VSS. 

I can see multiple approaches here:

- We could assume that everything is needed and create new signals for all that currently only exist in OBD. That could however be a quite tedious tasks - like how to represent lambda sensor banks for different engine configurations.
- We could use this PR to discuss what signals we think are needed and only add replacement signals for those that someone request, as part of this PR
- We could mark all as deprecated, and take the decision on which to create elsewhere later.

My original idea is to use the last approach, but allow long time for replacement signals to be created, i.e. remove OBD signals first in VSS 6.0. But your points are good, and we should align on the approach.

SebastianSchildt (approved):
I think this is good. Putting deprecation, but already migrating DTC (which is more or less the only point in OBD I have seen continuously being used"

 I do agree with @ppb2020 that we should double check we are not "loosing" any potentially useful information from the OBD branch, but as deprecation does not imply immediate removall, we might want to do this in a seperate issue/PR?

erikbosch:
Meeting notes:

- Sebastian: We could possibly add OBD as an overlay, for people to grab. There may still be users that want to keep using OBD data. Possibly add overlay file already now.
- AP: Erik to add info to changelog file. Add overlay, add to makefile

SebastianSchildt (approved):

erikbosch:
MoM: OK to merge

PR #702, opened 2024-01-04
Adding support Digital Key management

erikbosch:
Digital Keys are becoming common for vehicles and there is even standardization ongoing within the Car Connectivity Consortium (e.g. CCC Digital Key Release 3). This is a complex area and it is likely so that VSS is not suitable to handle all aspects of digital keys as the procedures for registering keys may vary considerably and may involve aspects like token/certificate evaluation and physical steps that must be performed within a vehicle. There are however areas where VSS representation could be useful. One potential use-case is to offer the driver/owner an overview of which digital keys that have been defined for the vehicle and the status of them. VSS could potentially be used to trigger certain management steps, like adding or removing a key. Note however that VSS should not in itself be responsible for security critical functionality, like validating that request have been signed by the trusted primary key.

The purpose of this PR is to initiate a discussion on if this makes sense to add to VSS,

### DIFF
--- a/spec/Vehicle/DigitalKey.vspec
+++ b/spec/Vehicle/DigitalKey.vspec
@@ -197,16 +197,16 @@ DigitalKey.KeyData.KeyCapability:
 
 DigitalKey.KeyData.StartTime:
   type: sensor
-  datatype: uint64
-  description: Start time of validity for the key.
-  unit: unix-time
+  datatype: string
+  description: Start time of validity for the key, formatted according to ISO 8601 with UTC time zone.
+  unit: iso8601
   comment: If time is before DigitalKey.KeyData.StartTime then the key cannot be used, even if state is ACTIVE.
 
 DigitalKey.KeyData.EndTime:
   type: sensor
-  datatype: uint64
-  description: End time of validity for the key.
-  unit: unix-time
+  datatype: string
+  description: End time of validity for the key, formatted according to ISO 8601 with UTC time zone.
+  unit: iso8601
   comment: If time is after DigitalKey.KeyData.StartTime then the key cannot be used, even if state is ACTIVE.
            After DigitalKey.KeyData.EndTime state of key may be changed to EXPIRED.
 
--- a/spec/Vehicle/DigitalKey.vspec
+++ b/spec/Vehicle/DigitalKey.vspec
@@ -1,0 +1,236 @@
+# Copyright (c) 2023 Contributors to COVESA
+#
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+#
+# This file specifies VSS signals that can be used to access and manage digital access keys for a vehicle.
+#
+# ************************ Feature Description **************************
+#
+# The digital key representation in VSS is based on the following feature:
+# - The vehicle key storage may store 1:n keys.
+# - The vehicle may interact with multiple keys at the same time.
+# - The capabilities associated with each key may differ, e.g., one key may give access to features like
+#   unlocking the vehicle and start the HVAC, while another one may be used to start the engine
+# - There is at most one valid "primary key" present in the vehicle key storage.
+#   - The primary key by default has all capabilities that are specified for a digital key.
+#   - It can be used to create additional "secondary keys". The creation process is outside the scope of this spec.
+# - The primary key registration process is outside the scope of this spec. It may require additional methods of authorization, e.g.,
+#   through a designated physical master key that is held by the vehicle owner.
+# - The secondary key registration process is for now outside the scope of this spec.
+#   - Secondary key invalidation is based on a custom implementation.
+#
+# The intention of the VSS signals in this file is to view and manage already existing keys.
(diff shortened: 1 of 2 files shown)


ppb2020:
I am in agreement that we may need to define some signals in this area. We mkay not be able to define what these signals are at this time, but we can start defining where they should reside in the tree. I can certainly see signals indicating whether the user has decided to block or enable sharing digital keys.

tspreck:
@tguild you onto this one also _ I know we were discussing the CCC activities the other day and how we could do something in the CV BoF group for fleet management & digital keys

spec/Vehicle/DigitalKey.vspec
```diff
@@ -0,0 +1,236 @@
...
+#
+# Each key (primary key or secondary keys) has an associated state.
+# The state indicates for example if the key is correctly setup and valid.
+#
+# For more information see DigitalKey.KeyData.KeyState
+#
+# There are several reasons why a key state may change. Note that the implementation is outside the scope of this spec.
+#
+# Examples (informative):
+# - If the primary key paring state changes to UNPAIRED all existing secondary keys may be terminated or removed.
```
ppb2020, line 64: paring -> pairing

spec/Vehicle/DigitalKey.vspec
```diff
@@ -0,0 +1,236 @@
...
+# - If the vehicle identity (DigitalKey.VehicleId) changes all existing keys may be terminated or removed.
```
ppb2020, line 65: changes -> changes,

spec/Vehicle/DigitalKey.vspec
```diff
@@ -0,0 +1,236 @@
...
+#
+#
+# Key Capabilities and Key Presence
+#
+# Keys may have different capabilities. Some keys can be used to drive the vehicle while others
+# only can be used for specific purposes or with specific restrictions.
+# The VSS model does not make any assumptions on what capabilities exist, but includes a signal
+# DigitalKey.KeyData.KeyCapability (of string datatype) which can be used to show capabilities
```
ppb2020, line 73: which -> that

spec/Vehicle/DigitalKey.vspec
```diff
@@ -0,0 +1,236 @@
...
+  description: Identifier used to identify the vehicle for digital key purposes.
+  comment: The vehicle identifier used for key purposes is not necessarily the same as any identifier used
+           for other purposes. For privacy reasons, it is likely not the same as the VIN.
+           The identifier may change during the lifetime of the vehicle, for example at owner change
+           to effectively invalidate all existing digital keys.
+
+DigitalKey.PairingState:
+  datatype: string
+  type: sensor
+  allowed: ['UNPAIRED', 'PAIRED', 'PAIRING','BLOCKED']
```
ppb2020, line 96: ,'BLOCKED' -> , 'BLOCKED"

spec/Vehicle/DigitalKey.vspec
```diff
@@ -0,0 +1,236 @@
...
+DigitalKey.WirelessCapabilities:
+  type: attribute
+  datatype: string
+  allowed: ['NFC', 'NFC_BLE', 'NFC_BLE_UWB']
+  description: Specifies which wireless capabilities the digital key system in the vehicle supports.
+               NFC indicates NFC Forum Analog and Digital Technical Specification 2.1.
+               BLE indicates Bluetooth Core Specification 5.0 or later.
+               UWB indicates IEEE Standard for Low-Rate Wireless Networks (802.15.4z-2020).
+
+DigitalKey.BluetoothLeAddress:
```
ppb2020, line 112: Presumably, this attribute is only present or set for a BLE DigitalKey, correct? What would be returned if this attribute is not set? All zeroes? Should a comment mention this?

Similar comment for other fields in this definition.

spec/Vehicle/DigitalKey.vspec
```diff
@@ -0,0 +1,236 @@
...
+  type: sensor
+  datatype: string
+  description: String indicating capabilities of the key.
+               Allowed values and interpretation is implementation-specific.
+
+DigitalKey.KeyData.StartTime:
+  type: sensor
+  datatype: uint64
+  description: Start time of validity for the key.
+  unit: unix-time
```
ppb2020, line 202: This should be UTC and not "wall clock" time. Same for other times herein.

ppb2020 (commented):
Some relatively minor comments and nit picks.

spec/Vehicle/DigitalKey.vspec
erikbosch: Historically we typically do not include any "undefined" value for VSS signals. So that if you ask for `Vehicle.VehicleIdentification.VehicleExteriorColor` or `Vehicle.Speed` and your VSS server currently does not know the vehicle color or the vehicle speed, the API should return "not available", "not found", "None" or whatever mechanism the API has to indicate that no value exist.

(This could off course be discussed for this particular case; if we would benefit from have a specific value for undefined/unknown)

spec/Vehicle/DigitalKey.vspec
erikbosch: That is implied by our definition of `unix-time` in https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml#L213:

`Number of non-leap seconds which have passed since 00:00:00 UTC on Thursday, 1 January 1970`

erikbosch:
Temporarily removed

Issue #525, opened 2023-01-18
UNIX timestamp s vs ms


adobekan:
Hi All,

What would be the good way on handling UNIX timestamp in ms and s as well? 

Right now we have following description
<html><head></head><body>

UNIX Timestamp | Time | Unix time is a system for describing a point in time.  It is the number of seconds that have elapsed since the Unix epoch,  excluding leap seconds.
-- | -- | --

</body></html>

erikbosch:
Strictly speaking we do not specify encoding or datatype for the VSS units. So there is nothing that prevents us from using `UNIX Timestamp` together with float, and tooling will not even complain if you use it together with string or boolean. So indirectly we could get milliseconds by using float representation. (We actually do not use the UNIX timestamp unit at all today in standard VSS catalog)

But I have no problem if we would like to add a `timestamp_ms` unit. At the same time we should possibly rename `UNIX Timestamp` to something like `timestamp_s` to get a common pattern and get rid of the blank.

A related topic - if we start changing timestamp type, should we then also consider creating a unit for "ISO 8601 time". We have today some signals with datatype string, no unit, but in free text stated that the string shall be encoded according to ISO 8601. This could possibly be handled by a unit `time_iso_8601`. Then we could possibly also in the unit file specify which datatypes that are allowed for a unit. E.g. specify that `timestamp_s` shall be allowed for `datatype: uint, int, float` (i.e. all except boolean and string), and `time_iso_8601` be allowed only for `datatype: string`. That would make it possible for tools to give errors if someone tries to define a timestamp signal using boolean as datatype.

( By the way, Neil Puthuff also brought up some ideas on timestamp handling with additional fields, see https://docs.google.com/document/d/1dHHMIlIZpsbnVsTiiza4Suua-vVbfR1xm11KNKdAm38/edit#)

erikbosch:
> This is how Apache IoTDB handles it https://iotdb.apache.org/UserGuide/Master/Data-Concept/Data-Type.html#timestamp but of course specifying (input, output) is different from internal representation.

As a side note - I think the we can improve our documentation on the meaning of datatype/unit in VSS. We have a short sentence in [documentation](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_unit_types/):

[...]

PR #703, opened 2024-01-04
Add battery precondition signals

erikbosch:
Electrical vehicles may have a system that prepares the traction battery for driving or charging by making sure that the battery has an optimal temperature for driving or charging. This PR proposes a set of signals for this purpose. The idea is that the signals should be quite generic and fit various type of implementations of a battery precondition system.


ppb2020:
Looks like a good start.

Regarding times, shall we require they be in UTC and ISO 8601 format?

I haven't worked through use cases, but the definitions should take into account what happens when "plans change". That is, the battery was being conditioned for a specific time and the vehicle is no longer parked or plugged in, etc. Indeed, the logic will be implemented in the vehicle, but we may need to have signals that represent the state or cause for interruption of the conditioning.

spec/Powertrain/TractionBattery/BatteryConditioning.vspec
```diff
@@ -0,0 +1,112 @@
...
+
+
+# Proposed to use unix time rather than time remaining (seconds) in VSS.
+# That does not prevent that time remaining is used internally,
+
+BatteryConditioning.StartTime:
+  type: actuator
+  datatype: uint64
+  description: Start time for battery conditioning, if Mode is not INACTIVE.
+  unit: unix-time
```
erikbosch, line 93: Pierre Pierre brought up the topic of time representation. 

We now have two time units; `unix-time` (to be used with uint32/uint64/int64) and `iso8601` (to be used with string). Historically we have favored `iso8601` over unix time, but we do not have a deliberate decision. Technically we could use iso8601 here as well, so what do we prefer?

SebastianSchildt: I think we had several discussion wrt ti time https://github.com/COVESA/vehicle_signal_specification/issues/667 , https://github.com/COVESA/vehicle_signal_specification/issues/525  and favoring iso is a more "recent" historical decisions.

I think personally some of you guys convinced me to prefer iso instead of unix timestamps in the past :D In any case it is what we are recommending in the documentation: https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types/#timestamps  and I would agree in terms of VSS data model (an implementation can still use faster/more primitive timestamps for internal representation)

erikbosch:
MoM:

- Some parts to be removed from PR, like "START OF PART NOT TO BE MERGED"
- AP: Erik to update
- Will be merged next week unless comments/remarks at next meeting

erikbosch:
Removed some comments not intended for merging from PR

SebastianSchildt (approved):

erikbosch:
MoM : Merge

PR #710, opened 2024-01-30
Feature/structure engine fluids

kowend:
Several engine oil and coolant related attributes where spread in the main branch of the combustion engine.
The engine oil and coolant have been put under a branch.
Additional coolant lifespan and level (ISO 7000 2429)  have been added under the coolant branch.

### DIFF
--- a/spec/Powertrain/EngineCoolant.vspec
+++ b/spec/Powertrain/EngineCoolant.vspec
@@ -6,26 +6,19 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-#
-# Engine coolant
-#
-EngineCoolant:
-  type: branch
-  description: Signals related to the engine coolant
-
-EngineCoolant.Capacity:
+Capacity:
   datatype: float
   type: attribute
   description: Engine coolant capacity in liters.
   unit: l
 
-EngineCoolant.Temperature:
+Temperature:
   datatype: float
   type: sensor
   unit: celsius
   description: Engine coolant temperature.
 
-EngineCoolant.Level:
+Level:
   datatype: string
   type: sensor
   allowed: [
@@ -35,7 +28,7 @@ EngineCoolant.Level:
     ]
   description: Engine coolant level.
 
-EngineCoolant.LifeRemaining:
+LifeRemaining:
   datatype: int32
   type: sensor
   description: Remaining engine coolant life in seconds.
--- a/spec/Powertrain/EngineCoolant.vspec
+++ b/spec/Powertrain/EngineCoolant.vspec
@@ -1,0 +1,45 @@
+# Copyright (c) 2016 Contributors to COVESA
+#
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+#
+# Engine coolant
+#
+EngineCoolant:
+  type: branch
+  description: Signals related to the engine coolant
+
+EngineCoolant.Capacity:
+  datatype: float
+  type: attribute
+  description: Engine coolant capacity in liters.
+  unit: l
+
+EngineCoolant.Temperature:
+  datatype: float
+  type: sensor
+  unit: celsius
+  description: Engine coolant temperature.
+
+EngineCoolant.Level:
+  datatype: string
+  type: sensor
+  allowed: [
+    'CRITICALLY_LOW',  # Critically low, immediate action required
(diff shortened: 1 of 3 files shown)


erikbosch:
Topics to discuss:

- This is right now considered as a "major" change (to be included in a major release only) as some signals will change name
- Alternatively, one could for now "keep" the old names (in addition to the new ones) but mark them as deprecated

kowend:
I think marking them as deprecated is fine, and leave them in. Is there any standard/rule for removing deprecations?
I can imagine we don't want too many deprecations left in the specs.

kowend:
Ah nevermind about the rule for deprecation. I found the description in the basics.md. But knowing the version for deprecation would still be useful. Would that be v5.0?

erikbosch:
> Ah nevermind about the rule for deprecation. I found the description in the basics.md. But knowing the version for deprecation would still be useful. Would that be v5.0?

Possibly. Historically we have selected "next release" as "deprecated from", but right now we have not really decided if we will release a 4.2 or if next release will be 5.0. I would say that 5.0 is the best option for now for main, if we decide to create a 4.2 then we as maintainers should anyway check all deprecation comments and adjust version if needed, like replacing 5.0 with 4.2 for the maintenance branch.

kowend:
Clear! I'll add the deprecation attributes. 
I've realised that the electrical motor could also have a cooling compartiment. At least the coolant temperature is mentioned as an optional. 
I've moved the coolant branch into a separate spec file.

spec/Powertrain/EngineCoolant.vspec
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2016 Contributors to COVESA
```
ppb2020, line 1: 2016 -> 2024?

spec/Powertrain/EngineCoolant.vspec
```diff
@@ -0,0 +1,38 @@
...
+  unit: celsius
+  description: Engine coolant temperature.
+
+Level:
+  datatype: string
+  type: sensor
+  allowed: [
+    'CRITICALLY_LOW',  # Critically low, immediate action required
+    'LOW',             # Level below recommended range, but not critical
+    'NORMAL',          # Within normal range, no need for driver action
```
ppb2020, line 27: Should this also have a HIGH and CRITICALLY_HIGH level as for EngineOil.Level?

ppb2020 (commented):
A few minor comments.

erikbosch:
MoM:

- Please review

spec/Powertrain/EngineCoolant.vspec
erikbosch: It would at least not hurt. I actually do not know what sensors modern high-end vehicles have for coolant; like can they detect and report that there is too much coolant?

ppb2020: I don't know enough about the mechanisms, but one possibility may be that the same kind of sensor could be used for both, for example. As well, what I am proposing seems to provide consistency that could be reused for other "level" sensors (washer fluid, transmission fluid, etc.). We do have another issue tackling common signals. Maybe we should consider a fluid-level signal that can be re-used for all these kinds of sensors?

kowend: I did consider reusing this and have the same enums. Technically I don't know what a typical sensor for the coolant could do; From a functional perspective I don't know if it makes sense to have HIGH and CRITICALLY_HIGH. For the engine oil I can imagine that having the right amount between 2 thresholds is important. But with coolant I don't think there is an upper threshold (other than you'll probably get wet shoes).
ISO 7000 2429, which is meant for the coolant level telltale describes 'low' or 'out of specified values'. In that case it would maybe make more sense to have something like  'NORMAL', 'LOW', 'CRITICAL'.

spec/Powertrain/EngineCoolant.vspec
erikbosch: In one way 2016 can make sense, as some of the signals actually have been copied from another file and maybe they were added in 2016, I have not checked.  So even if the file is created 2024, the content is older

erikbosch:
MoM:

- Pierre Pierre not present
- Merge decision next week unless comments/remarks

erikbosch:
MoM:

- Pierre on vacation, keep open. He is back next week

erikbosch:
MoM:

- Possible merge decision next week
- @ppb2020 could you check if it is ok

ppb2020:
I have no further comments. I am fine with the responses that were provided, with a note that I am still slightly in favour of reusing the set of enum values, but not enough to block merge.

erikbosch:
MoM:

- Merge

PR #716, opened 2024-02-08
Update documentation

### LINKED ISSUE #700: Enforce "style" for VSS standard catalog
Now and then the topic on what identifiers that shall be allowed in VSS comes up. As of today we are quite flexible in documentation, we sort of say that as long as it is valid Yaml we support it. This makes however tooling more complex if it is to handle identifiers with odd characters or allowed/default values containing blanks. We do state recommendations in some places like for [allowed values](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/allowed/#recommendation-on-string-values) and [node names](https://covesa.github.io/vehicle_signal_specification/rule_set/basics/#naming-conventions) and we have some support in vss-tools to give warnings or even fail if using strict mode.

But we do not enforce strict mode as a criteria for standard catalog, so as of today we have signals like this that does not follow the style guide (Values like "01" are not recommended):

```
PidsA:
  datatype: string[]
  type: attribute
  allowed: ["01","02","03","04","05","06","07","08","09","0A","0B","0C","0D","0E","0F","10","11","12","13","14","15","16","17","18","19","1A","1B","1C","1D","1E","1F","20"]
  description: PID 00 - Array of the supported PIDs 01 to 20 in Hexadecimal.
```

Some ideas:

- Remove the term "recommended" from VSS documentation. Instead separate between what is valid VSS syntax and what is acceptable for VSS standard catalog.
- Require that generic VSS tooling (the tooling that parse vspec files) can handle any VSS syntax
- But for generators/exporters only require that they can handle what is acceptable for VSS standard catalog
- Change Continuous Integration so that we use strict mode and give error if characters/style not allowed in standard catalog is detected.

Example:

_A signal name `Vehicle.Räksmörgås` is valid in VSS as it is a valid Yaml name. VSS-tools generic functionality shall support it. But we should not accept it in standard catalog as it does not follow the rules for the standard catalog. Exporters/Generators does not need to support `Vehicle.Räksmörgås`, but if so they should give an error._

But if we are to change like this we need to change the PId signals as well, but they may anyway be removed by #635

### LINKED ISSUE #705: Ranges for Float and Double
What is the source for float/double range at https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types/

Ranges are possible different compared to AUTOSAR

```

float | floating point number | -3.4e -38 | 3.4e 38
double | double precision floating point number | -1.7e -300 | 1.7e 300
```

Not fully compatible with for example:

[IEEE 754](https://en.wikipedia.org/wiki/IEEE_754) 32-bit base-2 floating-point
https://en.wikipedia.org/wiki/Double-precision_floating-point_format

Shall we possibly base limits on IEEE754 binary 32 and binary64

### LINKED ISSUE #706: Supported encoding for string values
VSS does not rely specify what character encodings we support like UTF-8, UTF-16, ASCII, ...

We need to discuss:

- What shall we specify in VSS, what do we leave for APIs
- AUTOSAR use UTF-8
- We say in *.vspec files that we support what is allowed by Yaml

erikbosch:
Clarifies string/float/double.
Reformat recommendatations to state that they also are rules for VSS catalog Remove enum recommendation

Fixes #705
Fixes #706
Fixes #700


docs-gen/content/rule_set/basics.md
```diff
@@ -109,17 +111,25 @@ The VSS specification must adhere to YAML syntax. To keep the standardized VSS s
...
+* Full path of node names must be unique also when using case insensitive comparison.
+  It is for example not allowed to have both `Vehicle.Abc` and `Vehicle.ABC` in the standard catalog
+  even if that is valid VSS as VSS is case sensitive.
 
 Examples:
 
 ```
-SomeBranch.AnotherBranch.MySignalName
 Vehicle.Cabin.Door.Row1.Left.IsLocked
+SomeOtherTree.AnotherBranch.MySignalName
```
SebastianSchildt, line 129: Not sure for what is this an example, doesn't it state above every branch need to start with "Vehicle"?

docs-gen/content/rule_set/data_entry/data_types.md
```diff
@@ -57,6 +62,8 @@ The time and date format shall be as shown below, where the sub-second data and
 YYYY-MM-DDTHH:MM:SS.ssssssZ
 ```
 
+VSS also supports the unit `unix-time` that can be used together with datatype `uint32`/`uint64`/`int64`to indicate that the signal value correspond to number of non-leap seconds which have passed since 00:00:00 UTC on Thursday, 1 January 1970.
```
SebastianSchildt, line 65: Is this relevant here, "unix-time" being a `unit` (not a datatype), and we are also not explicitly talking about all the other units in the standard catalogue unit file.

erikbosch:
MoM: Please review

docs-gen/content/rule_set/basics.md
```diff
@@ -109,17 +111,25 @@ The VSS specification must adhere to YAML syntax. To keep the standardized VSS s
...
-The recommended naming convention for node elements is to use camel case notation starting with a capital letter. It is recommended to use only
-`A-Z`, `a-z` and `0-9` in node names. For boolean signals it is recommended to start the name with `Is`.
+Node names in VSS must follow the restrictions specified in [Yaml](https://yaml.org/spec/1.2.2/#chapter-5-character-productions) i.e. must only use the printable subset of the Unicode character set.
+
+Tools and Generators supporting VSS may however put additional restrictions on which characters that are allowed. For maximum portability, node names in the VSS standard catalog must fulfill the following rules
+
+* Node names in the VSS standard catalog shall use camel case notation starting with a capital letter.
+* Node names in the VSS standard catalog shall use only `A-Z`, `a-z` and `0-9` in node names.
+* Boolean signals must start with `Is`.
+* Full path of node names must be unique also when using case insensitive comparison.
```
nickrbb, line 121: Lines

docs-gen/content/rule_set/basics.md
```diff
@@ -109,17 +111,25 @@ The VSS specification must adhere to YAML syntax. To keep the standardized VSS s
...
+  It is for example not allowed to have both `Vehicle.Abc` and `Vehicle.ABC` in the standard catalog
+  even if that is valid VSS as VSS is case sensitive.
```
nickrbb, line 123: Lines 121 to 123 are confusing. Can we shorten it to something like:
Node names are case insensitive for comparison operations, therefore the full path of node names must be unique e.g. `Vehicle.Abc` and `Vehicle.ABC` are the same in comparison and therefore the presence of both is prohibited.

docs-gen/content/rule_set/basics.md
erikbosch: In the VSS standard catalog the root name is `Vehicle`, but you may have a different tree using VSS syntax and then you can select whatever root name you want. Like if you use HIM to specify that multiple trees shall be used. I tried to make it more visible by renaming it from `SomeBranch` to `SomeOtherTree`. But that is sort of a special case so I am considering just removing this line.

docs-gen/content/rule_set/data_entry/data_types.md
erikbosch: But with that argument we could remove the whole `Timestamps` section from `data_types.md` as we do not have any specific "timestamp" datatype. I think it makes sense to somewhere describe the two methods we support for reporting timestamps, but maybe it fits better in the unit section of documentation.

erikbosch: Will refactor #717 a bit and then we can remove it here

erikbosch:
PR updated to address comments

erikbosch:
MoM: Merge

PR #717, opened 2024-02-09
Improving unit documentation

### LINKED ISSUE #690: Support of multiple measurement units in VSS signals
In PR#669, a discussion was started on the issue of different units of measurement being used by different customers. To recap, current we define a "default" unit, which is mostly, although not always, in metric. However, there is a large proportion of the world that doesn't use metric (or whatever default is specified) and so the default unit will not suit them. A common example is that of vehicle speed, where seemingly one half of the world's countries use kilometres per hour and the other half use miles per hour. This results in unit conversion always having to be done at entities that requested the signal and received a response in a different unit to the one that they prefer/require e.g. kilometres per hour to miles per hour.

In a client-server model, this effectively means all clients having to implement code to do such conversion, whereas it would be more efficient to have the server do the conversion needed and save duplicate code amongst client applications. BlackBerry has also observed that some in the industry do not follow the COVESA specified default unit and instead use their own preferred unit instead. This requires client applications to "know" that the signal information returned in a response contains a unit of measure different to that defined in VSS.

Therefore, it is proposed to enhance VSS to accommodate different preferences (potentially on a per regional basis) of measurement units rather than instil a single unit of measure on the whole industry, as we effectively do today. To accomplish this, it is proposed to enhance signal definitions to include multiple measurement unit types as part of the VSS signal path name. This then enables VSS to accommodate for a plethora of different unit types, and potentially also simplify client pre-configured "knowledge" of systems and unit conversions.

As an example for this, consider the Vehicle.Speed signal. This could be extended to include all possible units specified in the defined speed "quantity" (previously called "domain") e.g. Vehicle.Speed.mph, Vehicle.Speed.kmh. Or to put it another way, we introduce branches to signals, providing one branch per possible measurement unit type as defined in the units.yaml file. This could even be automated via vss-tools e.g. every time vss-tools finds a measurement unit domain indication, it could replace it with branches associated with each measurement unit; Vehicle.Speed.[quantity:speed] -> Vehicle.Speed.kmh, Vehicle.Speed.mph.

This way, a VSS "client" can request the unit it prefers and the VSS "server" can take care of any needed conversion and provide the requested signal in the requested unit. If no unit is provided, then we could also allow the default unit (as we already specify) be used instead. This would also allow for some backwards compatibility.

Comments welcome.

erikbosch:
Based on discussion in #690, clarify that use of of alternative units is not a topic for VSS but rather APIs

Fixes #690

### DIFF
--- a/docs-gen/content/rule_set/data_entry/data_units.md
+++ b/docs-gen/content/rule_set/data_entry/data_units.md
@@ -27,6 +27,12 @@ and a transport protocol may send speed in another unit, possibly also involving
 But in protocols not explicitly specifying data units (like [VISS](https://raw.githack.com/w3c/automotive/gh-pages/spec/VISSv2_Core.html))
 it is expected that `Vehicle.Speed` is sent and received as `km/h` (without scaling or offset).
 
+VSS does not offer any syntax for defining alternative units for a specific signal.
+The VSS project has specified for all units a corresponding quantity e.g. `velocity` for `km/h`.
+An API may use the quantity information to identify alternative units that can be used to represent the signal.
+The VSS list of units does not specify conversion factors, so even if VSS list both `km/h` and `m/s` as units for
+the quantity `velocity`, it is not specified how they relate to each other.
+
 In some cases it is natural to omit the data unit. This concerns typically signals where datatype `string` is used,
 but also signals where the value just represents a number (dimensionless quantities), like in the example below:
 
@@ -37,6 +43,7 @@ Vehicle.Cabin.DoorCount:
   default: 4
   description: Number of doors in vehicle.
 ```
+
 ### Logical Data Units
 
 VSS supports `percent` as data unit, typically with 0 to 100% as the allowed range.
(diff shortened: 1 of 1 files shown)


erikbosch:
MoM: Please review

nickrbb:
Thanks for creating a PR to resolve Issues #690 in the way we concluded earlier. Just for clarification, do we want to indicate that the use of an alternative unit can require additions to the units file? For instance, miles per hour (mph) is a unit that some regions use for Vehicle.speed but this is undefined in units.yaml.

Also, please find below some very minor nits and suggested resolutions:
1. "VSS does in itself not offer any syntax..." -> "VSS does not offer any syntax..."
2. "The VSS project has for all units specified also specified a corresponding quantity,lime `velocity` for `km/h`." -> "The VSS project has specified for all units a corresponding quantity e.g. 'velocity' for 'km/h'."
3. "The VSS list of units does however not specify conversion factors, ..." -> "The VSS list of units does not specify conversion factors, ..."

erikbosch:
Thanks for the comments. I think it could make sense to add a section on what to do if you need or want a new unit like `mph`. So far our approach for the unit file has been quite flexible - we do not just include units actually used by signals in the VSS standard catalog, but also other units that "possibly" could be make sense in future additions to the standard catalog or in custom extensions. I will try to add something.

erikbosch:
Refactored a bit to address the comments of @nickrbb

nickrbb:
Thanks @erikbosch for the updates, all looks good. I might create a new PR to add mph as a unit to the units file then, but as previously discussed and concluded, we will leave it to the API to determine its use in added or modified signals.

PR #718, opened 2024-02-09
Left/Right Instance Discussion

erikbosch:
This is a follow up proposal to the items listed below, showing a potential solution:

- Duplicating instances os we have both Left/Right and DriverSide/PassengerSide
- Leaving mapping to implementation

Concerning implementation, I see several alternatives.

- If an OEM exclusively use Left/Right or Driver/Passenger references, they can in overlay remove the labels they do not need, or just not provide values for the unused "dimension"
- If both are used it can either be managed within the server by some config, or you have a LHD/RHD-specific mapper that manage mapping where needed.

Would a change like in this PR be an acceptable solution?

https://github.com/COVESA/vehicle_signal_specification/pull/695
https://github.com/COVESA/vehicle_signal_specification/discussions/639
https://github.com/COVESA/vehicle_signal_specification/issues/642


erikbosch:
MoM: Please review, will need longer discussion if this is how we want things to be

erikbosch:
Putting it on hold/draft for now, in the sense that there seems to be more to discuss

docs-gen/content/rule_set/instances.md
```diff
@@ -10,17 +10,39 @@ quite often there is a need to repeat branches and data entries
...
+to reference particular instances. There is not necessarily a 1:1 relation between
+VSS instance identifiers and physical instances. A typical example could be instance identifiers
+for doors, windows and seats. For a left-hand drive vehicle, the driver seat may be referenced both
+as `DriverSide` and as `Left`. VSS does not provide any syntax to specify that VSS instance `A`
+(for this particular vehicle) actually refers to the same physical instances as `C`.
+It is up to the VSS implementation and deployment to manage this relationship, for example
+by a configuration file.
+
+
+As an example is shown below for doors. In VSS four instances of doors are specified for each row,
```
ppb2020, line 25: Review phrasing. Not clear what is meant by "As an example is shown below."

"In VSS" -> "In VSS,"

docs-gen/content/rule_set/instances.md
```diff
@@ -10,17 +10,39 @@ quite often there is a need to repeat branches and data entries
...
+but there are at most two physical doors on each side.
+`DriverSide`and `PassengerSide`can be considered as aliases for `Left`and `Right`, or vice versa.
```
ppb2020, line 27: `can -> ` can

as aliases -> aliases

`and -> ` and

docs-gen/content/rule_set/instances.md
```diff
@@ -10,17 +10,39 @@ quite often there is a need to repeat branches and data entries
...
 
 <!-- Image source in docs-gen/image_source/instance_tree.puml -->
 ![Example tree](/vehicle_signal_specification/images/instance_tree.png)
 
-When expanded this corresponds to:
+
+If expanded (for instance by [VSS-tools](https://github.com/COVESA/vss-tools)) this corresponds to:
```
ppb2020, line 33: ) -> ),

docs-gen/content/rule_set/instances.md
```diff
@@ -10,17 +10,39 @@ quite often there is a need to repeat branches and data entries
 branches and data entries in the specification an instance-concept is supported.
 Instances remove the need of repeating definitions, by defining at the node itself how often it occurs in
 the resulting tree. They are meant as a short-cut in the specification and
-interpreted by the tools. As an example is shown below for doors:
+interpreted by the tools.
+
+The instance names defined in VSS can be seen as labels
```
ppb2020, line 15: I think this makes sense, but I would like to spend some more time thinking this through as to how this would be implemented in practice.

I immediately see two approaches:
1. Use an overlay to (effectively) delete the entries that are not implemented (that is, replace ["DriverSide", "PassengerSide", "Left", "Right"] by ["Left", "Right"] or ["DriverSide", "PassengerSide"]).
2. Support them as aliases.

For 1., there are portability issues (an app written to look for DriverSide would not find it on vehicles that use a ["Left", "Right"] definition, and thus would need to be adapted). There is also the argument that if we expect an overlay to be used here, then going with one or the other of ["Left", "Right"] or ["DriverSide", "PassengerSide"] makes some minimal portability guarantee (at least for vehicles that use the definition as it is).

For 2., any code that reports the value for this signal would need to know that it needs to report (or store) this value twice and not just once as it normally does for other signals. Another issue is that 

As well, this approach embeds the notion of left-hand vs. right-hand driver side in a bit of an awkward way. Thinking out loud, would it be better for this notion of driver side not to be embedded in this signal, but carried independently, leaving this signal to only have ["Left", "Right"] as instances, with the application relaying on the independent driver position information to cheese between the two instances?

Lastly, where does this leave us with vehicles that do not have a driver side? That is, where the driver is in the middle?

ppb2020 (commented):
Some thoughts on this topic.

erikbosch:
Closing this one for now, may reopen later

PR #721, opened 2024-02-12
Exclude UUID from release artifacts

erikbosch:
This is a proposal to remove UUIDs from VSS 5.0 release artifacts. Background is that we have not yet heard about anyone actually using them, and if someone need them they can anyway generate them themselves.

Possible discussion points:

- If we agree on that we long term want to remove it, but think 5.0 is too soon then we could instead mention it as a planned change for 6.0
- If we agree on removing it we can discuss if we (long term) want to keep `--uuid` at all, or if we already now at 5.0 (or 6.0) shall mark it as deprecated, i.e. plan to remove it one major release later (i.e. 6.0 or 7.0)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -197,6 +197,18 @@ A new version of vss-tools is used, with support for static IDs and jsconschema.
 
 ## Planned Changes VSS 5.0
 
+### UUID removed from release artifacts
+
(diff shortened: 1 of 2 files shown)


erikbosch:
MoM:

- Merge
- Possibly deprecate/remove `--uuid` (separate PR)

PR #730, opened 2024-03-11
Enforce strict check in CI

erikbosch:
This means that CI checks will fail if a signal name does not fulfill the rules for being included in VSS standard catalog


erikbosch:
Mom:

- Erik presented it
- Please review, merge decision next week if no open topics

erikbosch:
MoM: Merge

PR #737, opened 2024-04-12
Add screen mirroring in Infotainment.vspec

spshin3:
This PR is rework of PR #699 .
The type of SmartphoneScreenMirroring.Resolution is changed to 'sensor'.

The differences of smartphone screen mirroring with smartphone projection as follow:
Screen Mirroring, such as through AirPlay, mirrors all activities of a device’s screen, allowing every detail to be displayed on a secondary screen. On the other hand, existing Projection specifically targets in-vehicle use, selectively providing access to certain applications in smartphone like navigation and music players.
Specifically, Screen Mirroring enables the casting of applications not typically offered by Projection, such as 'YouTube', to a vehicle's display. Therefore, Screen Mirroring and Projection are parallel technologies, each serving distinct use cases without overlapping functionalities.

### DIFF
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -267,4 +267,27 @@ SmartphoneProjection.SupportedMode:
   allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
 
+SmartphoneScreenMirroring:
+  type: branch
+  description: All smartphone mirroring actions.
+
+SmartphoneScreenMirroring.Active:
+  datatype: string
+  type: actuator
+  allowed: [ 'NONE', 'ACTIVE', 'INACTIVE' ]
+  description: Mirroring activation info.
+  comment: NONE indicates that mirroring is not supported.
+
+SmartphoneScreenMirroring.Source:
+  datatype: string
+  type: actuator
+  allowed: ['USB', 'BLUETOOTH', 'WIFI']
+  description: Connectivity source selected for mirroring.
+
+SmartphoneScreenMirroring.Resolution:
+  datatype: string
+  type: attribute
+  allowed: ['RES_720P', 'RES_1080P', 'RES_4K']
+  description: Display resolution for the mirrored content.
+
 #include include/PowerOptimize.vspec
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -270,8 +270,7 @@ SmartphoneProjection.SupportedMode:
 
 SmartphoneScreenMirroring:
   type: branch
-  description: All smartphone screen mirroring actions.
-  comment: Smartphone screen mirroring mirrors the whole screen of the Smartphone on the vehicle infotainment system.
+  description: All smartphone mirroring actions.
 
 SmartphoneScreenMirroring.Active:
   datatype: string
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -247,7 +247,6 @@ HMI.DisplayOffDuration:
 SmartphoneProjection:
   type: branch
   description: All smartphone projection actions.
-  comment: Smartphone projection exposes or controls specific applications on the Smartphone on the vehicle infotainment system.
 
 SmartphoneProjection.Active:
   datatype: string
@@ -261,6 +260,7 @@ SmartphoneProjection.Source:
   type: actuator
   allowed: ['USB', 'BLUETOOTH', 'WIFI']
   description: Connectivity source selected for projection.
+  comment: Smartphone projection exposes or controls specific applications on the Smartphone on the vehicle infotainment system.
 
 SmartphoneProjection.SupportedMode:
   datatype: string[]
--- a/spec/Cabin/Infotainment.vspec
+++ b/spec/Cabin/Infotainment.vspec
@@ -247,6 +247,7 @@ HMI.DisplayOffDuration:
 SmartphoneProjection:
   type: branch
   description: All smartphone projection actions.
+  comment: Smartphone projection exposes or controls specific applications on the Smartphone on the vehicle infotainment system.
 
 SmartphoneProjection.Active:
   datatype: string
@@ -270,7 +271,8 @@ SmartphoneProjection.SupportedMode:
 
 SmartphoneScreenMirroring:
   type: branch
-  description: All smartphone mirroring actions.
+  description: All smartphone screen mirroring actions.
+  comment: Smartphone screen mirroring mirrors the whole screen of the Smartphone on the vehicle infotainment system.
 
 SmartphoneScreenMirroring.Active:
   datatype: string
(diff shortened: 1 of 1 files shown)


spec/Cabin/Infotainment.vspec
```diff
@@ -267,4 +267,27 @@ SmartphoneProjection.SupportedMode:
   allowed: [ 'ANDROID_AUTO', 'APPLE_CARPLAY', 'MIRROR_LINK', 'OTHER' ]
   description: Supportable list for projection.
 
+SmartphoneScreenMirroring:
```
erikbosch, line 270: In https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Cabin/Infotainment.vspec#L247 we have 

```
SmartphoneProjection:
  type: branch
  description: All smartphone projection actions.
```

I think it would be good to add some more information on what is the difference between `SmartphoneProjection` and `SmartphoneScreenMirroring`.

My assumption (for now)

- `SmartphoneProjection` - Expose or control specific applications on the Smartphone on the vehicle infotainment system
- `SmartphoneScreenMirroring` - Mirror the whole screen of the Smartphone on the vehicle infotainment system

Or is there a better definition?

spshin3: Thanks erik. I think your definitions are good for me. 
I made comments for each of the function.

spec/Cabin/Infotainment.vspec
```diff
@@ -267,4 +268,28 @@ SmartphoneProjection.SupportedMode:
...
+  datatype: string
+  type: actuator
+  allowed: [ 'NONE', 'ACTIVE', 'INACTIVE' ]
+  description: Mirroring activation info.
+  comment: NONE indicates that mirroring is not supported.
+
+SmartphoneScreenMirroring.Source:
+  datatype: string
+  type: actuator
+  allowed: ['USB', 'BLUETOOTH', 'WIFI']
```
ppb2020, line 286: Consider using the following instead:
```suggestion
  allowed: ['USB', 'BLUETOOTH', 'WI-FI']
```

Although many folks tend to write "Wifi" or "WiFi", the correct spelling is "Wi-Fi": https://www.wi-fi.org/

spec/Cabin/Infotainment.vspec
```diff
@@ -267,4 +268,28 @@ SmartphoneProjection.SupportedMode:
...
+  description: Connectivity source selected for mirroring.
+
+SmartphoneScreenMirroring.Resolution:
+  datatype: string
+  type: sensor
+  allowed: ['RES_720P', 'RES_1080P', 'RES_4K']
```
ppb2020, line 292: Not clear if and how this is useful. If useful, there is possibly much more information that may be useful (screen size, etc.).

Pardon my lack of knowledge, but isn't resolution a factor of what the display supports as well as what the application wishes to use or requires? If so, how do we address it here?

spec/Cabin/Infotainment.vspec
erikbosch: I have no strong opinion but the literals used here correspond to what we have in https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Cabin/Infotainment.vspec#L258 

```
SmartphoneProjection.Source:
  datatype: string
  type: actuator
  allowed: ['USB', 'BLUETOOTH', 'WIFI']
  description: Connectivity source selected for projection.
```

So if we are to change to `WI-FI` we should also consider changing `Vehicle.Cabin.Infotainment.SmartphoneProjection.Source` (and then comes the question on how to handle backward compatibility, just adding `WI-FI` and keep `WIFI` for now, but state that `WIFI`may be removed later)

I suggest we discuss it at a meeting

spshin3: Thanks for your comment.

Personally, I don't have a strong opinion on this matter either. I also agree that following the global standards is important.

However, I'm little bit confused about whether it's appropriate to use "WI-FI" for data naming. When I see the android document, "WIFI" is used as a data type name, and "Wi-Fi" is used for descriptive text in the document. (https://source.android.com/docs/core/ota/modular-system/wifi)

Typically, hyphens ('-') are not used in data type names in VSS. Therefore, how about we keep using "WIFI" as it is?

spshin3: I think changing "WIFI" to "WI-FI" is not too difficult since we only need to update the naming in SmartphoneProjection. More important thing is making the right decision regarding the correct naming between WIFI and WI-FI.

spec/Cabin/Infotainment.vspec
spshin3: In my opinion, screen size is not an appropriate parameter for screen mirroring configuration. This is because even within the same screen size, smartphones can support various resolutions based on user preferences. The android base smartphone also supports the configuration of your resolution in 'configuration->display->screen resolution'.

The resolution setting is actually related to the display graphics processing load of the infotainment system, and typically, it can be adjusted to suit the user's preferences.

spec/Cabin/Infotainment.vspec
erikbosch: At the VSS meeting earlier this week we agreed that keeping `WIFI` is the best option for now, as we then do not need to change the other existing signal that use `WIFI`.

spec/Cabin/Infotainment.vspec
erikbosch: Would it be a possible way forward to clarify in the description that this signal concerns the resolution currently used, not maximum allowed resolution, possibly by changing to something like:

"Display resolution **currently used** for the mirrored content"

This off course does not prevent that more signals are added later if needed, like attributes for screen size, supported resolutions and so on, but that could be added later if/when needed.

spshin3:
I resolved the DCO error before the meeting. 
Thanks for your summary.

> MoM
> 
> * WI-FI/WIFI
> * PP: No major preferences, we could align with old  `WIFI
> * Stefan: Does not matter
> * Erik: Please review/discuss, continued discussion next meeting

erikbosch:
Trying to summarize my view on this PR.

- At last meeting we agreed to use `WIFI` rather than `WI-FI` to avoid changing name of other already existing signals using `WIFI`
- For the other signal, maybe we should clarify that the signal concerns currently used resolution to avoid ambiguity

At next meeting I suggest that we agree on remaining needed changes (if any) , so that we can merge it as soon as they have been fixed

spshin3:
> MoM
> 
> * WI-FI/WIFI
> * PP: No major preferences, we could align with old  `WIFI
> * Stefan: Does not matter
> * Erik: Please review/discuss, continued discussion next meeting

Thanks for your clarification.

In my personal opinion, resolving the ambiguity surrounding 'resolution' is challenging in our current signal.
This is because, we don't have a specific signal for display-related and resolutions with in our spec.

I think for this issue, we need to discuss adding the display-related signal before. 
Once this is added, we can clarify what 'resolution' means in the context of screen mirroring. 

Therefore, my suggestion for this PR is removing the resolution signal for this time.
And I will prepare another PR for the display related signal for the infotainment system.

Thanks.

spshin3:
> MoM:
> 
> * OK to merge if resolution signal removed
> * Richard Fernandes - Resolution can be complex due to DRM, we from AOSP could possibly contribute

Thanks for your feedback and the decision.
I just removed the resolution in my PR.

PR #746, opened 2024-06-13
Add rad/s

erikbosch:
COVESA VSS Standard Catalog already supports deg/s for angular speed, but in some areas rad/s is commonly used. This PR adds rad/s, so that it can be used both in the standard catalog as well as in user extensions.


nwesem:
Mom:
Adnan: ready to merge

PR #747, opened 2024-06-13
Update overlay example and tool dependency

erikbosch:
In many cases it is no longer required to specify type/datatype for overlays


nwesem:
MoM:
Adnan: please review and update description

erikbosch:
I updated the PR @adobekan but were there something more to be done?

erikbosch:
MoM:

- Niclas to check
- If no remark from Niclas ok to be merge

PR #750, opened 2024-06-24
New unit Lux and illuminance quantity

mikehaller:
For defining vehicle signal sensors which measure light intensity in Lux.

### DIFF
--- a/spec/units.yaml
+++ b/spec/units.yaml
@@ -398,3 +398,11 @@ Ohm:
   unit: Ohm
   quantity: resistance
   allowed-datatypes: ['numeric']
+
+# Illuminance
+
+Lux:
+  definition: Light intensity in Lux
+  unit: lux
+  quantity: illuminance
+  allowed-datatypes: ['numeric']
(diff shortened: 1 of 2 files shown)


erikbosch:
MoM:

- Please review, merge next week if no comments/remarks

spec/units.yaml
```diff
@@ -398,3 +398,11 @@ Ohm:
   unit: Ohm
   quantity: resistance
   allowed-datatypes: ['numeric']
+
+# Illuminance
+
+Lux:
```
erikbosch, line 404: Historically we have typically used the unit abbreviation/symbol as entry point here, also described at https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_units/

If we would follow that approach I think `lx` would be more correct, see https://en.wikipedia.org/wiki/Lux

erikbosch:
MoM:

- Erik to follow up with Mike

spec/units.yaml
erikbosch: @mikehaller - to be consistent with most other units I think `lx:` would be better than `Lux:` as we typically use the abbreviation/symbol of the unit. Any opinion?

erikbosch (approved):
LGTM

Issue #753, opened 2024-07-08
Allow instantiation of charging ports


jdacoello:
Right now, the charging port is modelled assuming the car has a unique port. However, there might be cases where multiple ports exist. Hence, the following suggestion is made:

# Current
Some properties of the port are modelled under the `Charging` branch. 
```YAML
Charging:
  type: branch
  description: Properties related to battery charging.
...
Charging.ChargePlugType:
  deprecation: V4.1 renamed to Charging.ChargePortType
  datatype: string[]
  type: attribute
  allowed: [
    'IEC_TYPE_1_AC',
    'IEC_TYPE_2_AC',
    'IEC_TYPE_3_AC',
    'IEC_TYPE_4_DC',
    'IEC_TYPE_1_CCS_DC',
    'IEC_TYPE_2_CCS_DC',
    'TESLA_ROADSTER',
    'TESLA_HPWC',
    'TESLA_SUPERCHARGER',
    'GBT_AC',
    'GBT_DC',
    'OTHER'
    ]
  description: Type of charge plugs (charging inlet) available on the vehicle.
               A charge plug type may occur multiple times in the list if there are multiple instances of that charge
               plug type.
               IEC types refer to IEC 62196,  GBT refers to  GB/T 20234.
  comment:     A vehicle may have multiple charging inlets.
               The signal Charging.ChargePlugPosition can be used to indicate position of the charge plug.
               IEC_TYPE_1_AC refers to Type 1 as defined in IEC 62196-2. Also known as Yazaki or J1772 connector.
               IEC_TYPE_2_AC refers to Type 2 as defined in IEC 62196-2. Also known as Mennekes connector.
               IEC_TYPE_3_AC refers to Type 3 as defined in IEC 62196-2. Also known as Scame connector.
               IEC_TYPE_4_DC refers to AA configuration as defined in IEC 62196-3. Also known as Type 4 or CHAdeMO connector.
               IEC_TYPE_1_CCS_DC refers to EE Configuration as defined in IEC 62196-3. Also known as CCS1 or Combo1 connector.
               IEC_TYPE_2_CCS_DC refers to FF Configuration as defined in IEC 62196-3. Also known as CCS2 or Combo2 connector.
               TESLA_ROADSTER, TESLA_HPWC (High Power Wall Connector) and TESLA_SUPERCHARGER refer to non-standardized charging
               inlets/methods used by Tesla.
               GBT_AC refers to connector specified in GB/T 20234.2.
[...]

Issue #749, opened 2024-06-24
Improve naming convention to distinguish instances from branches


jdacoello:
The current property naming convention can cause confusion when several branches exist, as it uses the same dotted notation for both branches and instances. 

# Simple scenario
For example, the property name `Vehicle.Cabin.Door.Row1.DriverSide.Window.Position` does not clearly differentiate between the `Row1.DriverSide` being an instance of the `Door` feature of interest, versus a branch in the property hierarchy.

To address this issue and improve the clarity of the property naming convention, the following two alternatives are proposed:

1. **Using Distinct Symbols to Enclose Instances**: Example: `Vehicle.Cabin.Door[Row1.DriverSide].Window.Position`. In this approach, a set of distinct symbols, such as square brackets `[ ]`, are used to enclose the instance information, making it visually distinct from the branch names. The specific symbols used (e.g., `[ ]`, `{ }`, `- -`etc.) can be determined based on team preference and overall consistency.
2. **Using a Separate Instance Identifier**: Example: `Vehicle.Cabin.Door.Instance[Row1.DriverSide].Window.Position` In this approach, a separate identifier, such as the keyword `Instance`, is used to explicitly indicate that the following information represents the instance of the feature of interest `Door`. The instance information is then enclosed in a set of distinct symbols, similar to the first solution. 

The goal of these proposals is to maintain the simplicity of the dot notation for branch names, while providing a clear visual distinction for the instance information. This will improve the overall readability and understanding of the property naming convention used in the codebase.

# Complex scenario
What about cases when multiple branches are instantiated to specify the property of a feature of interest? For example, a vehicle can have multiple instances of a `Door` (e.g., Rows and sides), and each one can also have multiple instances of `Speaker` (e.g., upper, middle, low, etc.). If we follow the current approach, we would end up with something like: `Vehicle.Cabin.Door.Row1.DriverSide.Speaker.Upper.Volume`. With the proposed approaches, we could have something nicer and easier to handle like:
1. `Vehicle.Cabin.Door[Row1.DriverSide]Speaker[Upper].Volume`
2. `Vehicle.Cabin.Door.Instance[Row1.DriverSide].Speaker.Instance[Upper].Volume`

Other examples can be, multiple buttons in a particular component, multiple lights in a particular position inside the cabin, multiple USB ports or button embedded in a seat, etc.

Any opinion or additional suggestion?

erikbosch:
[...]

PR #755, opened 2024-07-15
feat: Allow the specification of multiple charging ports.

jdacoello:
This PR addresses the issue described and discussed in #753 .

### DIFF
--- a/spec/Powertrain/TractionBattery.vspec
+++ b/spec/Powertrain/TractionBattery.vspec
@@ -236,6 +236,64 @@ Charging:
   type: branch
   description: Properties related to battery charging.
 
+Charging.ChargingPort:
+  type: branch
+  instances:
+    - ["FrontLeft", "FrontMiddle", "FrontRight", "RearLeft", "RearMiddle", "RearRight", "AnyPosition"]
+  description: Properties related to a particular charging port available in the vehicle.
+  comment: If a vehicle has a single charging port, then use the instance AnyPosition.
+
+Charging.ChargingPort.SupportedInletTypes:
+  datatype: string[]
+  type: attribute
+  allowed: [
+    'IEC_TYPE_1_AC',
+    'IEC_TYPE_2_AC',
+    'IEC_TYPE_3_AC',
+    'IEC_TYPE_4_DC',
+    'IEC_TYPE_1_CCS_DC',
+    'IEC_TYPE_2_CCS_DC',
+    'TESLA_ROADSTER',
+    'TESLA_HPWC',
+    'TESLA_SUPERCHARGER',
+    'GBT_AC',
+    'GBT_DC',
+    'OTHER'
+    ]
+  description: A list of the supported (i.e., available) charging inlets in a particular charging port.
+               IEC types refer to IEC 62196,  GBT refers to  GB/T 20234.
+  comment:     A vehicle may have multiple charging ports.
+               IEC_TYPE_1_AC refers to Type 1 as defined in IEC 62196-2. Also known as Yazaki or J1772 connector.
+               IEC_TYPE_2_AC refers to Type 2 as defined in IEC 62196-2. Also known as Mennekes connector.
+               IEC_TYPE_3_AC refers to Type 3 as defined in IEC 62196-2. Also known as Scame connector.
+               IEC_TYPE_4_DC refers to AA configuration as defined in IEC 62196-3. Also known as Type 4 or CHAdeMO connector.
+               IEC_TYPE_1_CCS_DC refers to EE Configuration as defined in IEC 62196-3. Also known as CCS1 or Combo1 connector.
+               IEC_TYPE_2_CCS_DC refers to FF Configuration as defined in IEC 62196-3. Also known as CCS2 or Combo2 connector.
+               TESLA_ROADSTER, TESLA_HPWC (High Power Wall Connector) and TESLA_SUPERCHARGER refer to non-standardized charging
+               ports/methods used by Tesla.
+               GBT_AC refers to connector specified in GB/T 20234.2.
+               GBT_DC refers to connector specified in GB/T 20234.3. Also specified as BB Configuration in IEC 62196-3.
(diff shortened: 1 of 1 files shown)


spec/Powertrain/TractionBattery.vspec
```diff
@@ -273,69 +331,28 @@ Charging.MaximumChargingCurrent.Phase3:
...
-Charging.ChargePortFlap:
-  deprecation: V4.1 - Replaced with Charging.IsChargePortFlapOpen
-  datatype: string
-  type: actuator
-  allowed: ['OPEN', 'CLOSED']
-  description: Status of the charge port cover(s), can potentially be controlled manually.
-               OPEN if at least one is open.
-
 Charging.IsChargePortFlapOpen:
+  deprecation: V4.3+ - Replaced with Charging.ChargingPort.IsFlapOpen to allow multiple charging ports.
```
SebastianSchildt, line 335: Shoudl we mark in the deprecation Note that it is instantiated now. Otherwise if somebody processes this automatically the results may be unexpected 

Suggestions

```
Charging.ChargingPort.*.IsFlapOpen
```

Or 

```
Charging.ChargingPort.<INSTANCE>.IsFlapOpen
```

SebastianSchildt:
I saw some (deprectaed in 4.1) elements deleted, is my understanding correct , that this is with the assumption this chance would only go into a 5.x? But in that case maybe the "new deprecations" are not needed as well.

OTOH, if we intend to add this to a 4.x release, I would assume leave everything in as deprecated? @erikbosch do you have an idea? Maybe we even documented our approach here? :D

spec/Powertrain/TractionBattery.vspec
jdacoello: Thanks for the observation. I could add, no problem. However, that deprecation message indicates that one must see the `Charging.ChargingPort.IsFlapOpen` instead. In that part of the specification, the instantiation of the branch is visible. 

I agree with the fact that the instantiation is not clear in the current naming convention. I actually started recently a [discussion and proposed an option to improve the human-friendly name of VSS properties](https://github.com/COVESA/vehicle_signal_specification/issues/749). 

The problem I see is that the descriptions, comments, deprecations, are not following a strict style or format. Therefore, even if I make the suggested change here, the specification will remain to be inconsistent. I think a (semi)automated method to curate the specification would be ideal.

jdacoello:
I deleted previous deprecations as they were referring to the same properties. Please, let me know whether such a change fits into a new 4.x release, or to a 5.0 release.

I can make the corresponding modifications.

spec/Powertrain/TractionBattery.vspec
jdacoello: I updated the deprecation note to make it the new instantiation explicit.

jdacoello:
@erikbosch @SebastianSchildt , the PR was updated according to what was discussed yesterday. Could you please review it one more time? Thanks!

spec/Powertrain/TractionBattery.vspec
```diff
@@ -236,6 +236,64 @@ Charging:
   type: branch
   description: Properties related to battery charging.
 
+Charging.ChargingPort:
+  type: branch
+  instances:
+    - ["FrontLeft", "FrontMiddle", "FrontRight", "RearLeft", "RearMiddle", "RearRight", "AnyPosition"]
```
erikbosch, line 242: The instances used here are somewhat different/reduced compared to previous set of allowed values. From my perspective not a problem but good to highlight when reviewing. The new values however correspond quite well with https://developer.android.com/reference/android/car/PortLocationType but there they use unspecified "front" instead of "FrontMiddle", and "Unknown" instead of "AnyPosition"

```
allowed: ['FRONT_LEFT', 'FRONT_MIDDLE', 'FRONT_RIGHT',
            'REAR_LEFT', 'REAR_MIDDLE', 'REAR_RIGHT',
            'LEFT_FRONT', 'LEFT_MIDDLE', 'LEFT_REAR',
            'RIGHT_FRONT', 'RIGHT_MIDDLE', 'RIGHT_REAR']
  description: Location of the charge port(s).
               First part indicates side of vehicle, second part relative position on that side.
               If supported, the list in this attribute shall have the same length as Charging.ChargePortType,
               and use same the relative order.
  comment: Example - If this signal is [LEFT_FRONT, RIGHT_FRONT] and Charging.ChargePortType is
           [IEC_TYPE_2_AC, GBT_AC] that means that there is Mennekes port on the left side of the vehicle near
           the front, and a GB/T AC port on the right side, near the front.
```

jdacoello: I would say that the `allowed_values` are more normative in comparison to `instances`. 
The instances are in the end arbitrary and often overwritten with overlays. 
The main idea is to have an agreement on the branch that can be instantiated (e.g., `Charging.ChargingPort`) than in the particular occurrences (e.g., `FRONT_LEFT)`. 

I must admit that I am not a big fan of default instances as part of the main core specification. In my view, the core of the specification should be the general classes (e.g., `Window`, `Door`, `ChargingPort`) and their properties (`position`, `isOpen`, etc.). With that in place, one might further specify either more particular subclasses of them or instances.

@erikbosch, do you want me to update that list of instances in the PR?

erikbosch: I do no really have any strong opinion, but why not base it on the terms in https://developer.android.com/reference/android/car/PortLocationType as default. (AnyPosition we could keep)

i.e. 

```
- ["Front", "FrontLeft", "FrontRight", "Rear",  "RearLeft", "RearRight", "AnyPosition"]

# OR 

- ["Front", "FrontLeft", "FrontRight", "Rear",  "RearLeft", "RearRight", "Unknown"]


```

erikbosch:
MoM:

- JD: Believe the name of instances will not be so specific, will likely be changed with overlay. Can keep instances in VSS to a minimum. 
- Merge!

adobekan:
MoM: Merge.

PR #758, opened 2024-07-23
Fixed inconsistent units.yaml

sschleemilch:
Found while doing a vss-tools rework with more validation logic


erikbosch (approved):
Looks good to me

PR #761, opened 2024-07-31
Add check that catalog can be represented in correct CSV file

erikbosch:
An idea from ongoing changes in vss-tools. We try to escape characters as needed in description/comment, but as of today we do not verify that our escape mechanism is sufficient for standard catalog. This PR adds a test for that purpose.


erikbosch:
MoM: Please review

adobekan (approved):
kk

PR #763, opened 2024-08-01
Add documentation for delete argument

erikbosch:
I noticed that we did not add any documentation on the `delete` argument added some months ago in vss-tools. This is an attempt for a short documentation.

_We can discuss if we also should mention it in the syntax, and not only as an overlay example._

### DIFF
--- a/docs-gen/content/rule_set/overlay.md
+++ b/docs-gen/content/rule_set/overlay.md
@@ -10,6 +10,7 @@ In order to adopt the specification we realize that certain additions and modifi
 
 The following features with the intended usage patterns are currently supported:
 1. **Adding new nodes:** By adding nodes the standard catalog can be extended with proprietary concepts.
+1. **Deleting nodes:** Deleting nodes not relevant for the vehicle.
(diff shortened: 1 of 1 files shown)


docs-gen/content/rule_set/overlay.md
```diff
@@ -114,6 +115,39 @@ Vehicle.Occupant.Row1.DriverSide.HeadPosition.Yaw:
   unit: mm
 ```
 
+### Deleting nodes
+
+It is possible to delete nodes using the `delete` atrribute.
```
sschleemilch, line 120: typo: atrribute

docs-gen/content/rule_set/overlay.md
```diff
@@ -114,6 +115,39 @@ Vehicle.Occupant.Row1.DriverSide.HeadPosition.Yaw:
...
+# Removing Window for all Door instances
+Vehicle.Cabin.Door.Window:
+  delete: true
+
+# Removing Vehicle.Speed
+Vehicle.Speed:
+  delete: true
+```
+
+When using the `delete` argument vss-tools will print information on nodes removed.
```
sschleemilch, line 140: The output changes with https://github.com/COVESA/vss-tools/pull/382

PR #769, opened 2024-08-27
adding AdaptiveDistance AdaptiveTimeGap for insurance needs in CV activity

tguild:
Drawing from Android Automotive documentation

https://source.android.com/docs/automotive/vhal/adas-properties

which has:

ADAPTIVE_CRUISE_CONTROL_TARGET_TIME_GAP
ADAPTIVE_CRUISE_CONTROL_LEAD_VEHICLE_DISTANCE

we came up with the equivalent for VSS. Similar would logically exist for Emergency Braking Assist (EBA) which presumably uses the same sensor (lidar, camera or radar. While the time gap can be calculated based on distance and vehicle speed, we feel it should be a primary signal.

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -86,6 +86,16 @@ CruiseControl.IsError:
   type: sensor
   description: Indicates if cruise control system incurred an error condition. True = Error. False = No Error.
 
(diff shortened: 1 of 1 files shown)


tguild:
Removing as I had a whitespace isssue, have appropriate pre-commit checks now and also changed how we want to propose representing these signals.

PR #770, opened 2024-08-27
adding ObstacleDetection and Adaptive Cruise control signals

tguild:
adding ObstacleDetection:ForwardDistance ObstacleDetection:ForwardTimeGap CruiseControl.AdaptiveIsActive CruiseControl.AdaptiveDistanceSet

Drawing from Android Automotive documentation

https://source.android.com/docs/automotive/vhal/adas-properties

which has:

ADAPTIVE_CRUISE_CONTROL_TARGET_TIME_GAP
ADAPTIVE_CRUISE_CONTROL_LEAD_VEHICLE_DISTANCE

The Commercial Vehicle group came up with the proposed PR for VSS. Cruise Control and Emergency Braking Assist may share the sensor data if vehicle has both so we felt it appropriate to place some of this under ObjectDetection. While the time gap can be calculated based on distance and vehicle speed, we feel it should be defined and represented in VSS.

Nearby we also want to include blind spot detection, both left and right. Vehicle SAE ActiveAutonomyLevel less than 2 tend to merely indicate presence of a vehicle in blind spot, surely level 3 and greater keep track of distance. Instead of staging a proposed representation for left and right isWarning booleans and possibly distance, want to discuss on VSS call what a better future proof representation may be. There may be object (obstacle) presence and distance detection for multiple directions.

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -86,7 +86,7 @@ CruiseControl.IsError:
   type: sensor
   description: Indicates if cruise control system incurred an error condition. True = Error. False = No Error.
 
-CruiseControl.AdaptiveIsActive:
+CruiseControl.isAdaptiveActive:
   datatype: boolean
   type: actuator
   description: Indicates if adaptive cruise control system is active (i.e. actively controls speed).
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -86,6 +86,18 @@ CruiseControl.IsError:
   type: sensor
   description: Indicates if cruise control system incurred an error condition. True = Error. False = No Error.
 
+CruiseControl.AdaptiveIsActive:
+  datatype: boolean
+  type: actuator
+  description: Indicates if adaptive cruise control system is active (i.e. actively controls speed).
+               True = Active. False = Inactive.
+
+CruiseControl.AdaptiveDistanceSet:
+  datatype: float
+  type: actuator
+  unit: m
+  description: Distance in meters to keep from lead vehicle
+
 #
 # Lane Departure Detection System
 #
@@ -130,6 +142,18 @@ ObstacleDetection.IsError:
   datatype: boolean
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
+
+ObstacleDetection:ForwardDistance:
+  datatype: float
+  type: sensor
+  unit: m
+  description: Distance in meters to forward detected object
+
+ObstacleDetection:ForwardTimeGap:
+  datatype: int32
+  type: sensor
+  unit: ms
+  description: Time in milliseconds before potential impact to forward object
 
 
 #
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -86,7 +86,7 @@ CruiseControl.IsError:
   type: sensor
   description: Indicates if cruise control system incurred an error condition. True = Error. False = No Error.
 
-CruiseControl.isAdaptiveActive:
+CruiseControl.IsAdaptiveActive:
   datatype: boolean
   type: actuator
   description: Indicates if adaptive cruise control system is active (i.e. actively controls speed).
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -143,13 +143,13 @@ ObstacleDetection.IsError:
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
-ObstacleDetection:ForwardDistance:
+ObstacleDetection.ForwardDistance:
   datatype: float
   type: sensor
   unit: m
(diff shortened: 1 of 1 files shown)


spec/ADAS/ADAS.vspec
```diff
@@ -131,6 +143,18 @@ ObstacleDetection.IsError:
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
+ObstacleDetection.ForwardDistance:
+  datatype: float
+  type: sensor
+  unit: m
+  description: Distance in meters to forward detected object
```
erikbosch, line 150: - Is the idea that if `ObstacleDetection.IsWarning == False` then the value of this signal is not relevant, or do we need some special values (for example negative values) to represent that there is no obstacle present.
- How is this signal supposed to behave when reversing? - reporting obstacles behind vehicle or not reporting obstacles at all?
- Is it deliberately not specified how you calculate distances? Like if you are planning a turn (or the road turns), should it report the straight distance to the obstacle, or the distances you need to drive to hit the obstacle

spec/ADAS/ADAS.vspec
```diff
@@ -131,6 +143,18 @@ ObstacleDetection.IsError:
...
+
+ObstacleDetection.ForwardTimeGap:
+  datatype: int32
```
erikbosch, line 153: Could be uint32, or do we intend to use negative values to represent cases like no foreseeable impact.

spec/ADAS/ADAS.vspec
tguild: In the initial comment made with this PR and this one, very interested in how we can accommodate multitude of direction sensors on existing advanced (SAE AV 4+) or experimental vehicles. As the configurations will vary we cannot use simple directions (front, left, left front, etc) but instead could use degrees with 0 being directly in front of the vehicle. This compares to degrees on a compass (Vehicle.CurrentLocation.Heading in VSS). Sensor type would be of interest as we are seeing a mix.

StefanSellschopp: I assume the distance to object in front would relate to objects in the trajectory of the moving vehicle. So if there is a curve, stationary objects would not be relevant, if the trajectory is not hitting the object...

erikbosch:
MoM:

- Please review
- Ted: This is just first contributions, we have other additions in pipe, future proof design of signals needed, like objects in blind sport. Input from OEMs wanted
- Stefan: I have seen some examples with "intention", like is pedestrian likely to cross the street or not
- Daniel: Modelling goes a bit beyond on what we can discuss here, we could possibly reuse info from https://www.asam.net/standards/asam-openxontology/
- Stephen: "feature complete" or work in progress. May need to have "concept" info that explains the intention of all small steps.
-

spec/ADAS/ADAS.vspec
```diff
@@ -86,6 +86,18 @@ CruiseControl.IsError:
   type: sensor
   description: Indicates if cruise control system incurred an error condition. True = Error. False = No Error.
 
+CruiseControl.IsAdaptiveActive:
+  datatype: boolean
+  type: actuator
+  description: Indicates if adaptive cruise control system is active (i.e. actively controls speed).
+               True = Active. False = Inactive.
```
erikbosch, line 93: Do we need to clarify meaning of false? Like what should this signal show if cruise control is active but not in adaptive mode (i.e. it is in old fashioned mode were it keeps the speed until you press the brake pedal or crashes)

StefanSellschopp: to my knowledge, there is either adapative or non adaptive cruise control. But it does not change. This signal would just provide the info of what is active in the vehicle...

tguild: We discussed this further on the Commercial Vehicles' insurance focused breakout call and looked over some of the other nearby (obstacle) signals that exist in production vehicles we would like to see in VSS. Specifically we were looking at left and right blind spot detection mentioned previously and rear cross traffic detection.

ObstacleDetection as it is presently exists in VSS could handle this if there are multiple leaf nodes and well known (defined locations) similar to how multiple tires, seats etc with some generalized optional attributes for 
things like warning type, distance, timegap (desired but debatable).  Vehicle.ADAS.ObstacleDetection.Distance.Forward for example in place of ObstacleDetection.ForwardDistance as was in this PR.

WarningType can have various allowed values such as REAR_CROSS_TRAFFIC so that we know it is detecting a moving instead of stationary object.

Instead of scrapping and proposing a new PR, hope to discuss on an upcoming VSS call as there may be other thoughts on how to represent these.

tguild: We also decided in CV call that we should focus on mainstream production vehicles with their existing sensors and leave aside more complex configurations of evolving AV/ADAS for the time being. We would be interested in participating in discussion on future proofing VSS signals.

erikbosch:
MoM:

- Not to be part of 5.0
- Ted: We have identified some insurance/fleet signals, discussing how to progress
- Ted: Will create new PR replacing this one, input appreciated
- Ted: We have an insurance call where we discuss needed signals, feel free to join

PR #772, opened 2024-09-10
Submodule update and CI extension

erikbosch:
Adding support for new tools

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ all: clean mandatory_targets optional_targets
 
 # All mandatory targets that shall be built and pass on each pull request for
 # vehicle-signal-specification or vss-tools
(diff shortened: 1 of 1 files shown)


erikbosch:
Note that building ttl fails, but as it is an "optional tool" test passes anyway.  Topics to discuss:

- What to do with ttl, see https://github.com/COVESA/vss-tools/issues/403
- How to treat the new ones (apigear and samm), optional OR mandatory. I am leaning towards treating them all as "mandatory", and deliberately "move" them to "optional" or "obsolete" when/if they fail
- For the new ones - do we need any artifacts as part of release artifacts. If so how? They generate multiple files so if so zip them?

SebastianSchildt (approved):
Looks good to me and is a good evolutionary step before thinking about changing anything about Makefiled and/or Actions

PR #778, opened 2024-09-17
Remove submodule (version 2)

erikbosch:
An alternative to #768

I did some more changes which I think would make sense to introduce for the upcoming 5.0, in short building on what Mr Schildt already have done.

In short - adding a release workflow that triggers build and uploads what we want to upload. Also removing the VERSION file as I do not really see that it is needed and using generic names for build artifacts making handling easier.

An example on how this could affect release handling.

- If we want to release a v1.6rc0 from master then we trigger the Draft release workflow stating 1.6.rc0 as version.
- That will build all artifacts and upload them to a draft release
- When publishing the release source code is added and tagging performed and we get something like https://github.com/erikbosch/vehicle_signal_specification/releases/tag/v1.6rc0
- (In a real release we should update release notes before publishing)


erikbosch:
MoM:

- Presented
- Erik and Sebastian to agree on what to include for 5.0

PR #780, opened 2024-10-04
ObstacleDetection instances, attributes and Adaptive CruiseControl

tguild:
Adaptive cruise control distance setting typically seen on the dashboard user interface as increments (eg 1-5) that may be either vehicle lengths or an incremental variable that also factors in vehicle speed for safe following distance. If the latter a setting of 1 may be 10 meters at 30km/h but 30 at 90 km/h. In that case it would be more important to note the user preferred an adaptive buffer of 1 that has no fixed length but fluctuates with speed. We ask the OEM and suppliers on the VSS call to give clarity to conventions in production wrt adaptive cruise control.

In adding instances to ObstacleDetected, we want to be able to support multiple leaf nodes as multiple obstacles can be detected concurrently. Here too we want input from OEM and suppliers on the multitude of obstacle detecting sensors in mainstream vehicles. We realize more advanced autonomous vehicles will have far more lidar/camera/other sensors detecting and tracking obstacles. Degrees with zero being directly in front would provide more flexibility but trying to keep it simple. We want to avoid Center in both Y and X axis and based on feedback will likely go with "Middle" or "-" for Y.

  instances:
    - ["Front", "-", "Rear"]
    - ["Left", "Center", "Right"]
 
We also want ObjectType such as wildlife, pedestrian, bicycle, vehicle, etc but again hope someone in the community can direct us to an effort like Sensoris that should be leveraged instead of defining ourselves.

### DIFF
--- a/spec/ADAS/ADAS.vspec
+++ b/spec/ADAS/ADAS.vspec
@@ -86,6 +86,17 @@ CruiseControl.IsError:
   type: sensor
   description: Indicates if cruise control system incurred an error condition. True = Error. False = No Error.
 
+CruiseControl.IsAdaptive:
+  datatype: boolean
+  type: actuator
+  description: Indicates if cruise control system is adaptive (i.e. actively controls speed).
+
+CruiseControl.AdaptiveDistanceSet:
+  datatype: float
+  type: actuator
+  unit: m
+  description: Distance in meters to keep from lead vehicle
+  
 #
 # Lane Departure Detection System
 #
@@ -113,6 +124,9 @@ LaneDepartureDetection.IsError:
 #
 ObstacleDetection:
   type: branch
+  instances:
+    - ["Front", "Rear"]
+    - ["Left", "Center", "Right"]
   description: Signals form Obstacle Sensor System.
 
 ObstacleDetection.IsEnabled:
@@ -131,6 +145,28 @@ ObstacleDetection.IsError:
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
+ObstacleDetection.Distance:
+  datatype: float
+  type: sensor
+  unit: m
+  description: Distance in meters to detected object
+
+ObstacleDetection.TimeGap:
+  datatype: int32
+  type: sensor
+  unit: ms
+  description: Time in milliseconds before potential impact object
+
+ObstacleDetection.WarningType:
+  datatype: string
+  type: sensor  
+  allowed: [
+    'UNDEFINED',      # Object detection warning not further categorized
+    'CROSS_TRAFFIC',  # Detected object is on an indirect trajectory that may intersect
+    'BLIND_SPOT',     # Detected object may not be in driver line of sight, including from mirrors
+    ]
+  description: Indicates the type of obstacle warning detected as some track not only the presence of an obstacle but potential intercepting trajectory or other characteristics.
+  comment: Undefined obstacle warning type would merely alert of presence of obstacle and may measure distance.
 
 #
 # Antilock Braking System


spec/ADAS/ADAS.vspec
```diff
@@ -131,6 +145,28 @@ ObstacleDetection.IsError:
...
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
+ObstacleDetection.Distance:
+  datatype: float
+  type: sensor
+  unit: m
+  description: Distance in meters to detected object
+
+ObstacleDetection.TimeGap:
+  datatype: int32
```
erikbosch, line 155: Has negative number a meaning, or would uint32 suffice?
Has this one and `Distance` only significance if `ObstacleDetection.IsWarning` is True, or what values should be used if there is no obstacle?

spec/ADAS/ADAS.vspec
```diff
@@ -113,6 +124,9 @@ LaneDepartureDetection.IsError:
 #
 ObstacleDetection:
   type: branch
+  instances:
+    - ["Front", "Rear"]
```
erikbosch, line 128: Do we need the instances also for the existing signals like `IsEnabled`, `IsWarning` and so on? If not we could split to two branches, one called `ObstacleDetection` (without instances) and one called `Obstacles` (with instances)

spec/ADAS/ADAS.vspec
tguild: Unsigned is fine but wanted decimal instead of rounding to full integer.

spec/ADAS/ADAS.vspec
tguild: Each ObstacleDetection sensor could throw a separate warning, be disabled independently etc,

tguild: Previous conversation before considering instances approach

https://github.com/COVESA/vehicle_signal_specification/pull/770

erikbosch:
MoM:

- Please review
- Ted: For adaptive cruise control you typically have some settings. 
- Ted: Setting in meters proposed for now, or would a symblic setting in  "steps" (`1-2-3-...`) be better for OEM  needs?
- Ted: Instance concept proposed (front left, front center, and so on) , does it fit OEM needs?
- Stefan: we are not aware of any existing baseline 
- Richard: Some vehicles have car-lengths, likely OEM specific. Cars with automatic parking has more sensors so will likely differ vehicle to vehicle.
- Erik: For wipers we have both symbolic steps (e.g. to control frequency when use automatic mode) and absolute (Hz)
- Daniel: ASAM has an extensive model for similar things. https://www.asam.net/fileadmin/News/Brochures/ASAM_SIM-Guide_Online.pdf

tguild:
Hej @erikbosch I took a look at the wiper example and only see frequency in cycles per minute, not an alternative with a stepped setting (eg slow, medium and fast - potentially represented as integer 1, 2, 3).

tguild:
Thanks @erikbosch I was looking at WiperSystem.vspec which has Frequency instead of Body.vspec. 

I do not see any mechanism in YAML that suggests an either or situation.  It seems both are defined and the underlying vehicle can support either or both Frequency or Mode (optionally plus Intensity for vehicles equipped with rain sensor).

spec/ADAS/ADAS.vspec
```diff
@@ -86,6 +86,22 @@ CruiseControl.IsError:
...
+  type: actuator
+  description: Indicates if cruise control system is adaptive (i.e. actively controls speed).
+
+CruiseControl.AdaptiveDistanceSet:
+  datatype: float
+  type: actuator
+  unit: m
+  description: Distance in meters to keep from lead vehicle
+
+CruiseControl.AdaptiveDistanceSet:
```
erikbosch, line 100: Duplicate name?

erikbosch:
DCO check is missing, a sign off is needed, see https://covesa.global/about-covesa/#contribute-code.

Instructions on how to rebase and add signoff at https://github.com/COVESA/vehicle_signal_specification/pull/780/checks?check_run_id=33207647757

spec/ADAS/ADAS.vspec
```diff
@@ -86,6 +86,22 @@ CruiseControl.IsError:
...
+  type: actuator
+  description: Indicates if cruise control system is adaptive (i.e. actively controls speed).
+
+CruiseControl.AdaptiveDistanceSet:
+  datatype: float
+  type: actuator
+  unit: m
+  description: Distance in meters to keep from lead vehicle
+
+CruiseControl.AdaptiveIntvervalSet:
```
erikbosch, line 100: Typo `Intverval`

spec/ADAS/ADAS.vspec
```diff
@@ -131,6 +150,28 @@ ObstacleDetection.IsError:
   type: sensor
   description: Indicates if obstacle sensor system incurred an error condition. True = Error. False = No Error.
 
+ObstacleDetection.Distance:
+  datatype: float
+  type: sensor
+  unit: m
+  description: Distance in meters to detected object
```
erikbosch, line 157: Can never be negative, or?

And distance is distance from the specified "zone". I.e. an object just in front of the front number plate may be reported with distance==0.01 meter by Front.Center, but with distance==0.80 meter by Front.Right?

erikbosch:
MoM:

- OK to merge when review comments from Erik is addressed

spec/ADAS/ADAS.vspec
tguild: Distance from sensor and would be in zonal instance eg Front Right.

Wording this detail tricky, let's hash out in comments, slack, call or other. 

Description: Distance in meters from detected object to perimeter, physical boundary of vehicle at the identified zone (eg Front Right).. Cannot be a negative number, expected value is  positive float (unsigned).

erikbosch:
Merged by #793

PR #785, opened 2024-10-15
Add signals for Motion Management

erikbosch:
This is a proposal to add a number of signals for vehicle motion management, for instance making sure that vehicle stability is maintained during braking and when turning. As this is somewhat complex area it is also proposed to add a page in documentation to describe possible usage.

I will invite subject matter experts to the VSS meetings October 22nd and 29th so that they can give background to the change and answer questions.

### DIFF
--- a/docs-gen/content/_index.md
+++ b/docs-gen/content/_index.md
@@ -14,5 +14,6 @@ about what the specification is all about, what's in and out and how to quickly
 * [Governance](/vehicle_signal_specification/governance/) </br> Introduction to VSS governance model and change log.
 * [Rule Set](/vehicle_signal_specification/rule_set/) </br> Get familiar with the rule set behind the
 specification, the structure, the format and concepts.
+* [VSS Catalog](/vehicle_signal_specification/catalog/) </br> Information on content and design decisions related to the VSS catalog.
 * [License](/vehicle_signal_specification/license/) </br> Information on the VSS license.
 * [Ecosystem](/vehicle_signal_specification/ecosystem/) How to use the specification? Check out the existing tools and standards to transform or make use of VSS.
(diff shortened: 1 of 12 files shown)


SebastianSchildt:
You mean invite them October? I assume not September NEXT year?

erikbosch:
> You mean invite them October? I assume not September NEXT year?

Oh dear, is it already October?

SebastianSchildt:
I was wondering, with these "more specialized" VSS areas, should we give easy options (i.e. via overlay) to ONLY transform this branch, or maybe even formulate this as a self contained tree (and thus potential overlay).

Don't get me wrong, I think it is very cool to extend VSS in this direction, and instead of blocking this with another half year discussion, let's rather check and merge it. I am just thinking that the criteria for "generally useful" when deciding what to add later are different from the VehicleMotion Branch compared to the whole tree (or let's say infotainment), so maybe it is the time to have the moduels/profiles/composable stndnard model discussion again.

erikbosch:
MoM:

- Stefan: Important to have `AsIs` and `ShouldBe`
- Mattias: There are where relevant
- Erik: Potential merge decision next week, if no open points/reviews/remarks/discussions at next meeting 5th we may take a merge decision.
- Erik: Continue review, please add discussion topics to the Pull Request so Mattias/Markus can read/prepare and if possible join next meeting.
- Ulf: Thinks it is reasonable for some signals to have both current and target, when reasonable. "Current" shall be without suffix, target shall have "Target" suffix.

erikbosch:
MoM:

- Please review, to be merged next week unless there are any outstanding discussions/objections

erikbosch:
MoM: Merge

PR #786, opened 2024-10-28
Add RegExp pattern for VehicleIdentification.VIN property.

Kostadin-Ivanov:
Add Regular Expression pattern for the VehicleIdentification.VIN property.

This PR is dependent on [COVESA VSS-Tools PR: 426](https://github.com/COVESA/vss-tools/pull/426)

The idea is to provide a Regular Expression validation for the which is compliant with the mentioned in the property **description** - [ISO 3779](https://cdn.standards.iteh.ai/samples/52200/7d8a69aee84c4ad28231053f49f4966e/ISO-3779-2009.pdf).

Provided RegExp - pattern is tested on: https://regex101.com/, with a test string of: 1CPH423GAP71E2745 and validates VIN strings as per quoted [ISO 3779](https://cdn.standards.iteh.ai/samples/52200/7d8a69aee84c4ad28231053f49f4966e/ISO-3779-2009.pdf) on below rules:

    1. VIN length of 17 chars where:
    1.1 First 3 characters are for the VMI: 1CP
    1.2 Next 6 characters are for the VDS: H423GA
    1.3 Last 8 characters are for the VIS: P71E2745
    2. The last 4 characters of the VIS part should always be numbers
    3. Letters: I, O and Q shall not be used

**NOTE:** This update should be delivered AFTER the above mentioned  [COVESA VSS-Tools PR: 426](https://github.com/COVESA/vss-tools/pull/426) gets approved and released. Otherwise it will raise the following ERROR / WARNING:

```
WARNING: Unknown extra attribute: 'Vehicle.VehicleIdentification.VIN':'pattern'
```

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -55,6 +55,7 @@ VehicleIdentification.VIN:
   datatype: string
   type: attribute
   description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
+  pattern: ^([0-9A-HJ-NPR-Z]{3})([0-9A-HJ-NPR-Z]{6})([0-9A-HJ-NPR-Z]{4}[0-9]{4})$
 
 VehicleIdentification.WMI:
   datatype: string


spec/Vehicle/Vehicle.vspec
```diff
@@ -55,6 +55,7 @@ VehicleIdentification.VIN:
   datatype: string
   type: attribute
   description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
+  pattern: ^([0-9A-HJ-NPR-Z]{3})([0-9A-HJ-NPR-Z]{6})([0-9A-HJ-NPR-Z]{4}[0-9]{4})$
```
erikbosch, line 58: I like the idea. Lets discuss the general idea of supporting/allowing a `pattern`keyword at the VSS meeting today.

If we agree that it is a good idea for cases like this then we need to look into details:
* When should it be allowed, I assume only for `string` and `string[]` signals and struct properties.
* Which syntax is supported - I think it would be good if we could specify something specific like that the pattern should comply with [POSIX Basic Regular Expressions ](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03). At least for the standard catalog there is a value in selecting a standard/subset which is widely supported in different libraries so that it easily can be verified in VISS (Go), Kuksa (Rust/Python) and vss-tools (Python) 

With those questions answered it would be easy to document this. One could also when discussing this think about #771. For VIN it is unlikely that we will have more than one signal with VIN-syntax, but for other use cases one could think of that we want to reuse the pattern, and if we introduce/refactor format-handling then one could in addition consider allowing pattern also for formats, for instance to give  regexp that complies with `iso8601` timestamps.

Kostadin-Ivanov: Thank you Erik.
I will add recommended validations in the related VSS-Tools PR, where @sschleemilch also advised to provide similar validation of the field in the VSS model.

erikbosch:
MoM:

- Ulf: Creating a dependency might give problems if standard changes. A bit doubtful on putting it in the tree, could be put in documentation
- Stefan: Validation proposal is good, 
- Daniel: Better to provide function that do the check, can be standardized for some types.
- Erik: With format we can reuse
- Erik: Can iso8601 be described using a regexp, or too many variants
- Erik: Please review

Kostadin-Ivanov:
> MoM:
> 
>     * Ulf: Creating a dependency might give problems if standard changes. A bit doubtful on putting it in the tree, could be put in documentation
> 
>     * Stefan: Validation proposal is good,
> 
>     * Daniel: Better to provide function that do the check, can be standardized for some types.
> 
>     * Erik: With format we can reuse
> 
>     * Erik: Can iso8601 be described using a regexp, or too many variants
> 
>     * Erik: Please review

Hello @erikbosch , I tried to check online about using Regular Expressions for datetime strings and found below two discussions on StackOverflow:

1. [Regex and ISO8601 formatted DateTime [duplicate]](https://stackoverflow.com/questions/12756159/regex-and-iso8601-formatted-datetime)
2.  [Regex validate correct ISO8601 date string with time](https://stackoverflow.com/questions/28020805/regex-validate-correct-iso8601-date-string-with-time)

The first one has some good examples and the **accepted** answer in the second one provides very nice breakdown diagram of the corresponding RegExp.

With regards to the first comment, from Ulf, I'd say that we could just provide some general, in the case of iso8601 (date standard), regular expression and if user needs to update it or have some more strict one, they can always use the [VSS Overlays](https://covesa.github.io/vehicle_signal_specification/rule_set/overlay/index.html) to customize it.

Same can be said about the VIN number, although I doubt that there would be too many variants of VIN strings, but this is just my guess here.

erikbosch:
MoM:

- Ok to merge when vss-tools updated and test passes

PR #787, opened 2024-10-31
Add ConsumptionSinceLastRefuel

CoffeeMoneyMakers:
Added fuel consumption from current refuel to next refuel.


erikbosch:
MoM:

- Please review
- Stephen:  Do we have any guidelines on when/how to add units into standard unit catalog?
- Erik: We have previously been quite open to accept units that could be useful for the industry.
- Stephen: We should make sure that it has a general value.

PR #793, opened 2024-12-10
Pr780

erikbosch:
Based on #780 with minor proposed updates


erikbosch:
Reviewed in slack

image

PR #796, opened 2024-12-13
Clarify VSS expectations on read/write

erikbosch:
I ended up in a discussion of what "doing a read" in VSS means. In short, as of today that we say that a read returns actual value. This PR suggests to change the wording somewhat, so that a read unless otherwise specified shall return the actual value.

I.e. it is not forbidden by VSS (project) for implementations to have a read-function that returns the desired value instead, but then that must be clearly specified in the corresponding implementation specification.

This shall not impact VISS/VISSR.


SebastianSchildt (approved):
Looking good to me. It is not "strictly" VSS (i.e. a user might doe "whatever" or in a Non-API system this does not even apply), but there where words to this effect before and maybe now it is more clear what the best - or at least common - practice is.

It is also not contradicting what VISS or KUKSA is doing, just to name two "VSS users" I am aware off.

erikbosch:
MoM:

- Please review
- Daniel will review, possibly suggest improvements
-

erikbosch:
MoM: Merge

Issue #797, opened 2024-12-17
Modify the vspec to guarantee uniqueness for the pattern {inmmediate_parent_name}.{branch_name}


jdacoello:
Sub issue derived from #790

PR #799, opened 2024-12-17
refactor: Unique names for the pattern {parent_banch_name}.{branch_name}

jdacoello:
Addresses the issue #797 .

### DIFF
--- a/spec/Cabin/Seat.vspec
+++ b/spec/Cabin/Seat.vspec
@@ -29,6 +29,7 @@ IsOccupied:
   description: Does the seat have a passenger in it.
 
 Occupant:
+  deprecation: v5.0 - use data from Vehicle.Occupant.
   type: branch
   description: Occupant data.
 
@@ -53,6 +54,15 @@ SeatBeltHeight:
 #include include/ItemHeatingCooling.vspec
 
 Massage:
+  datatype: uint8
+  type: actuator
+  min: 0
+  max: 100
+  unit: percent
+  description: Seat massage level. 0 = off. 100 = max massage.
+  deprecation: v5.0 - refactored to Seat.MassageLevel
+
+MassageLevel:
   datatype: uint8
   type: actuator
   min: 0
@@ -105,8 +115,18 @@ Backrest.Recline:
 Backrest.Lumbar:
   type: branch
   description: Adjustable lumbar support mechanisms in seats allow the user to change the seat back shape.
+  deprecation: v5.0 - nested properties moved to Seat.Backrest.<Property>.
 
 Backrest.Lumbar.Support:
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Lumbar support (in/out position). 0 = Innermost position. 100 = Outermost position.
+  deprecation: v5.0 - refactored to Seat.Backrest.LumbarSupport
+
+Backrest.LumbarSupport:
   datatype: float
   type: actuator
   unit: percent
@@ -121,12 +141,32 @@ Backrest.Lumbar.Height:
   unit: mm
   description: Height of lumbar support. Position is relative within available movable range of the lumbar support.
                0 = Lowermost position supported.
+  deprecation: v5.0 - refactored to Seat.Backrest.LumbarHeight
+
+Backrest.LumbarHeight:
+  datatype: uint8
+  type: actuator
+  min: 0
+  unit: mm
+  description: Height of lumbar support. Position is relative within available movable range of the lumbar support.
+               0 = Lowermost position supported.
 
 Backrest.SideBolster:
   type: branch
   description: Backrest side bolster (lumbar side support) settings.
+  deprecation: v5.0 - nested properties moved to Seat.Backrest.<Property>.
 
 Backrest.SideBolster.Support:
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Side bolster support. 0 = Minimum support (widest side bolster setting).
+               100 = Maximum support.
+  deprecation: v5.0 - refactored to Seat.Backrest.SideBolsterSupport
+
+Backrest.SideBolsterSupport:
   datatype: float
   type: actuator
   unit: percent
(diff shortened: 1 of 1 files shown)


spec/Cabin/Seat.vspec
```diff
@@ -182,147 +213,300 @@ Airbag.IsDeployed:
...
+  type: actuator
+  description: Is switch to increase seating length engaged.
 
 Switch.Seating.IsBackwardEngaged:
   datatype: boolean
   type: actuator
   description: Is switch to decrease seating length engaged (SingleSeat.Seating.Length).
+  deprecation: v5.0 - refactored to Seat.Seating.IsBackwardSwitchEngaged
+
+Switch.IsBackwardSwitchEngaged:
```
erikbosch, line 433: Shall be `Seating.IsBackwardSwitchEngaged`?

spec/Cabin/Seat.vspec
```diff
@@ -182,147 +213,300 @@ Airbag.IsDeployed:
...
   description: Switches for SingleSeat.Massage.
+  deprecation: v5.0 - nested properties moved to Seat.<Property>
 
 Switch.Massage.IsIncreaseEngaged:
   datatype: boolean
   type: actuator
   description: Increase massage level switch engaged (SingleSeat.Massage).
+  deprecation: v5.0 - refactored to Seat.IsIncreaseMassageSwitchEngaged
+
+Seat.IsIncreaseMassageSwitchEngaged:
```
erikbosch, line 498: Remove `Seat.` here and on next I assume

spec/Cabin/Seat.vspec
```diff
@@ -182,147 +213,300 @@ Airbag.IsDeployed:
...
 Switch.Massage.IsDecreaseEngaged:
   datatype: boolean
   type: actuator
   description: Decrease massage level switch engaged (SingleSeat.Massage).
+  deprecation: v5.0 - refactored to Seat.IsDecreaseMassageSwitchEngaged
+
+Seat.IsDecreaseMassageSwitchEngaged:
+  datatype: boolean
+  type: actuator
+  description: Decrease massage level switch engaged.
```
erikbosch, line 512: Add a new line (return) here, otherwise we likely will get a pre-commit error

spec/Cabin/Seat.vspec
jdacoello: You are right. I missed that one. I will push the change.

erikbosch:
I took a quick look. In general I think the approach of removing the "switch" sub-branch makes sense.

A side-note for switches. In VSS we typically have for each seat attribute (e.g. headrest tilt) have 3 data, position (absolute value in mm/degrees or relative in percent) and two "switches" for more/less. In [Android VHAL](https://source.android.com/docs/automotive/vhal/system-properties) they have a partially different representation with two data, position and "speed", the advantage is that you can represent how fast you want the seat to move.

 Operation | Android | VSS
------------ |----------|-----
Move to a specific position | Set `SEAT_BACKREST_ANGLE_1_POS` |  Set `Backrest.Recline`
Move forward | Set `SEAT_BACKREST_ANGLE_2_MOVE` to a positive value |  Set `Backrest.IsReclineForwardSwitchEngaged` to True
Move backward | Set `SEAT_BACKREST_ANGLE_2_MOVE` to a negative value |  Set `Backrest.IsReclineBackwardSwitchEngaged` to True

spec/Cabin/Seat.vspec
jdacoello: True. I suggest that if we introduce sub branches, they must be as explicit as possible. For example, the branch name `SeatMassage` would be preferred over `Massage`. Why? Because such explicit names still hold a meaningful label if they are seen alone. I suggest that the maintenance of the model from now on should focus on two lists:
* List of Features of Interest (FoI).
* List of Properties (or other FoIs) that belong to that FoI.

So, the modeling is simplified as
```mermaid 

graph LR

    FoI_A --> Property_1
    FoI_A --> Property_2
    FoI_A --> Property_N
    FoI_A --> FoI_B
    
    classDef c1 fill:#9370DB,stroke:#333;
    
    class FoI_A,FoI_B c1;
```

The hierarchy of nested branches is still there. The only difference is that, in this way, the concepts make sense and can exist independently from where you want to place it in the tree.

jdacoello: I checked the spec again. We actually have a `Massage` property directly linked to `Seat`:
```yaml
Massage:
  datatype: uint8
  type: actuator
  min: 0
  max: 100
  unit: percent
  description: Seat massage level. 0 = off. 100 = max massage.
```
 
The challenge I see is that we have multiple classifications, whereas the tree has one dimension only.

- Seat instances: Vehicle can have multiple instances of `Seat`.
- Seat parts: Each `Seat` has `Headrest`, `Backrest`, and `Seating`.
- Switches in each part
- Massage function (possibly also multiple, one for each seat part).

It seems that `Massage` is modelled right now as if the function would always be centrally controlled. Here are two options I see:

1. Do not include a sub branch `Massage` and attach massage properties directly to the `Seat` branch.
```mermaid
graph LR
    Seat --> MassageProperty_A
    Seat --> MassageProperty_B
    Seat --> MassageProperty_N
    Seat --> Headrest
    Seat --> Backrest
    Seat --> Seating

    classDef c1 fill:#9370DB,stroke:#333;
    
    class Seat,Headrest,Backrest,Seating c1;
```
2. Add a massage sub branch to each seat part as `HeadrestMassage`, `BackrestMassage`, `SeatingMassage`. This option will also imply to deprecate the existing `Seat.Massage` and include that property to each seat part as `HeadrestMassage.MassageIntensity`, `BackrestMassage.MassageIntensity`, `SeatingMassage.MassageIntensity`.
```mermaid
graph LR
    Seat --> Headrest
    Seat --> Backrest
    Seat --> Seating

    Headrest --> HeadrestMassageProperty_A
    Headrest --> HeadrestMassageProperty_B
    Headrest --> HeadrestMassageProperty_N

    Backrest --> BackrestMassageProperty_A
    Backrest --> BackrestMassageProperty_B
    Backrest --> BackrestMassageProperty_N

    Seating --> SeatingMassageProperty_A
    Seating --> SeatingMassageProperty_B
    Seating --> SeatingMassageProperty_N

    classDef c1 fill:#9370DB,stroke:#333;
    
    class Seat,Headrest,Backrest,Seating c1;
```

Which option does sound better @erikbosch ?

erikbosch: For now I think keeping it simple makes sense. I.e. not divide it into different components. Maybe you could check with other COVESA representatives from your organization - I think they discussed Massage as part of the capability discussion, that it is an area that may vary between OEMs and maybe not as simple to represent as different intensity values for different sub-components. @paulboyes  - maybe you can summarize something from the capability discussions in this area.

jdacoello: So, for now, I will go for option 1.

jdacoello:
> A side-note for switches... you can represent how fast you want the seat to move

@erikbosch Interesting that you mention this feature. Specifying custom arguments for the properties is natively supported by the `GraphQL Schema Language`. 

```graphql
type Seat {
  isReclineForwardSwitchEngaged(mechanismSpeed: SeatReclineSpeed = MEDIUM): Boolean
}

enum SeatReclineSpeed {
  SLOW
  MEDIUM
  FAST
}
```

You can see how things are still connected, even when they are maintained separately. In current VSS, we would need two different data points (or an struct) to convey the same meaning.

spec/Cabin/Seat.vspec
paulboyes: Regarding Massage in the Capabilities discussion... Massage is considered a competitive feature so only the minimum  discussed as a group:

Massage Feature Available - 

- yes/no

Massage Status - 

- on/off

Seat Massage Type

- Off
- Roll
- Knead
- Anti Fatigue
- Pulse
- Stretch
- Lumbar
- Shoulder
- Wave
- Relief
- Glide

When everyone is back in January, we should go through the capabilities and align with VSS

erikbosch (approved):
Looks good to me, lets discuss tomorrow if we shall merge it

erikbosch:
MoM:

- Daniel: With this pattern we guarantee unique names. Not needed in private domains.
- Ulf: Think the full path should be the unique name. We should not require unique names for segments.
- Erik: We could extend vss-tools to enforce name checks for the standard catalog.
- Ulf to review until next meeting
- Paul to discuss with Daniel offline
- Stephen: Possible relation to capability work - as both touch seats.
- Possible merge decision next weeking

UlfBj:
I think the idea of having unique node names as a general rule (the name of the separate node, not the path which always is unique) makes the model more complicated, which is not a good direction.

Nodes representing an “object” of the same type should in my view have the same node name.

Examples:
Vehicle.Powertrain.CombustionEngine.EngineCoolant
Vehicle.Powertrain.ElectricMotor.EngineCoolant
The uniqueness is given by the path, which also provides information about the deployment context in the vehicle.

However, if nodes with the same name do not represent the same  logical type then they should have different names.

Examples:
Vehicle.Body.Lights.Brake
Vehicle.Chassis.Brake
These two Brake branch nodes above have different children definitions.

 

The model with nodes that represent the same logical object having the same name has the drawback that the definition of these do not have one unique source. This is a problem that is solved in the CVIS project where we develop multiple trees for Trucks, Trailers, etc., that share common objects. The common objects are defined separately from the tree definitions, see links below. 

[CVIS objects](https://github.com/COVESA/commercial-vehicle-information-specifications/tree/main/spec/objects
)
[CVIS trees](https://github.com/COVESA/commercial-vehicle-information-specifications/tree/main/spec/trees
)
This model is compatible with the input to VSS-tools (which are used in the CVIS project).

Combining this model with the rule that node names that do not represent the same logical type would in my view provide an improved architecture/design.

erikbosch:
I think one need to separate unique names for "classes" and "instances". Taking `Lumbar` as example we have today two different "classes", one maintaining characteristics of the Lumbar, the other maintaining the switches that can be used to control the lumbar. So if I understand it correctly, they fall into the category where Ulf thinks changing name makes sense, in this PR it is refactored by removing both instances of the Lumbar branch/class.

```
Backrest.Lumbar:
  type: branch
  description: Adjustable lumbar support mechanisms in seats allow the user to change the seat back shape.

Switch.Backrest.Lumbar:
  type: branch
  description: Switches for SingleSeat.Backrest.Lumbar.
```
The other part concern names for "instances". In VSS we have so far used two mechanisms for re-use. In most cases we have (potentially) unique branch name but reuse content, like in the example below where we reuse the contents of `SingleConfigurableLight.vspec` but add it into the `AmbientLight` branch.

```
AmbientLight:
  type: branch
  instances:
    - Row[1,2]
    - ["DriverSide","PassengerSide"]
  description: Decorative coloured light inside the cabin, usually mounted on the door, ceiling, etc.
#include SingleConfigurableLight.vspec AmbientLight
```

In some cases we do however have the branch-definition in the included file, like in this example

`#include TractionBattery/BatteryConditioning.vspec`

My view:

- I do not see that requiring unique `parent.leaf` in VSS standard catalog would give us any significant limitations
- We could add it as a "strict" check in vss-tools. I.e. enforce it for standard VSS-catalog but allow non-unique names when not using "strict". Or control it my a separate argument, to enforce uniqueness in VSS std catalog, but not in CVIS, for example
- The handling of `#include` may anyhow need to be improved (for some tools). Today it act like a macro, and expanded early in the analysis process, which is not optimal if you want to represent a VSS catalog as a model.
- The changes in the PR are quite good anyway, I do not really like the old "switch" branch

Proposed discussions point in next VSS meeting.

- Do we agree enforcing unique `parent.branch` for std VSS catalog (or all VSS models) is a good step forward?
- Regardless of that, do we think the proposed changes in this PR makes VSS better?

jdacoello:
@UlfBj 
> I think the idea of having unique node names as a general rule (the name of the separate node, not the path which always is unique) makes the model more complicated, which is not a good direction.

In contrast to what you mention, using the whole concatenated path as the ID is the actual problem. That is what makes changes, modifications, deprecations, etc. so painful. 
Considering the same example you listed:
- `Vehicle.Powertrain.CombustionEngine.EngineCoolant`
- `Vehicle.Powertrain.ElectricMotor.EngineCoolant`
> Nodes representing an “object” of the same type should in my view have the same node name.

Here, they both have the same `EngineCoolant` name. Which is fine if they convey the same meaning. But, if they refer to the same thing, why do we have two unique IDs? Here, there must be just one!

The proposal is to remove that dependency, such that we can simply maintain `EngineCoolant` decoupled from the path.
Then, arbitrary paths would be still possible to attach.

jdacoello:
@erikbosch 
You mention `parent.leaf` combination. So I think it is important that we can uniquely identify both: `parent` and `leaf`. 
`parent` is the branch (i.e., feature, object, or entity), whereas `leaf` is the property you want to read or write.

An ideal case, with the example of `Vehicle.Powertrain.CombustionEngine.EngineCoolant.Temperature`,would be something like:

All the elements should have unique IDs. like:
* `feat:Vehicle` (entity)
* `feat:Powertrain` (entity)
* `feat:CombustionEngine` (entity)
* `feat:EngineCoolant` (entity)
* `prop:Temperature` (property)

Here, `ns` is a namespace as a prefix for a compact representation. Where:
* `PREFIX feat: https://www.covesa.org/vspec/feature/`
* `PREFIX prop: https://www.covesa.org/vspec/property/`

Namespace can also be unique. Concepts of public interest will be allocated in the covesa namespace. Likewise, if a company needs their own namespace, the approach can follow the same construction.

Note: In the case of uniqueness for `parent`, at the moment, it is easier to address the pattern `{parent_branch_name}.{branch_name}` first.

UlfBj:
>> Considering the same example you listed:
Vehicle.Powertrain.CombustionEngine.EngineCoolant
Vehicle.Powertrain.ElectricMotor.EngineCoolant
Nodes representing an “object” of the same type should in my view have the same node name.
Here, they both have the same EngineCoolant name. Which is fine if they convey the same meaning. But, if they refer to the same thing, why do we have two unique IDs? Here, there must be just one!

These are two different signals so they must have unique IDs=paths.

Unique parent.leaf node name combinations are Ok, unique node names would as I described be a problem.

erikbosch:
MoM:

- Ulf: Having unique names for parent+child combinations is OK. Change OK if it helps Daniel.
- Decision: Merge

jdacoello:
> the rule should be unique parent.child node name combinations.

That's correct. The combination is and must remain unique.

jdacoello:
Now that this change has been merged, we need to add a test that checks for this rule (at least when strict mode is used). @erikbosch Where is the best place to do that (vspec or vss-tools)?

erikbosch:
Vss-tools for the check. Possibly extending the method below.

https://github.com/COVESA/vss-tools/blob/770a7a39aec669d1f89d3ca0a75dab0a804503fc/src/vss_tools/tree.py#L268

Then we should better add the rule as a style guide in vss documentation as well.

I have limited time to work on generic VSS-improvments right now so would be great if you or @sschleemilch would have time.

jdacoello:
I will try to implement it then.

BTW, it seems that while this fix was being discussed, the concept `Vehicle.MotionManagement.Brake.Axle.Wheel` was introduced to the spec. Which is again creating the undesired duplicated names:
```
                                            
fqn                                                  
Vehicle.Chassis.Axle.Wheel                 Axle.Wheel
Vehicle.MotionManagement.Brake.Axle.Wheel  Axle.Wheel
```

I will create another issue and suggest the fix for it.

PR #800, opened 2024-12-18
D3 statistics

Giaccomole:
New content for introduction page.

+Interactive plots from Observable
+Static PNG's (few unused fallbacks in images)

Please do simulate in your local machine if you want to inspect the front page, or shoot me a message for specific notebook links.

**The last two charts with Donut Chart in Releases and Hierarchical Bars are probably unnecessary. Feedback appreciated**

Screenshot 2024-12-18 at 16 14 41


docs/images/fav.png
erikbosch, line 1: Do we need the `docs` images here. Isn't the `docs` folder just the location for generated documentation. Possible add `docs` to `.gitignore`?

docs-gen/config.toml
```diff
@@ -25,3 +25,8 @@ weight = 12
 # Generation of JSON index to allow search
 [outputs]
 home = ['html', 'rss', 'search', 'searchpage']
+
+# So that raw HTML content is enabled in Hugo project
+
+[markup.goldmark.renderer]
+unsafe = true
```
erikbosch, line 32: The pre-commit error is likely because lack of new line after last line

erikbosch:
I like the idea, great work. I published generated documentation at https://erikbosch.github.io/vehicle_signal_specification/introduction/index.html if someone wants to look at it without building locally.

Some thoughts:

- We should possibly exclude VSS 1.0 from the statistics as it was too much a prototype so numbers are not that relevant
- 
image

The circle diagram shows v1.0 as default, if we want to have it possibly latest release should be default if possible. Also numbers seems to differ compared to the diagram above for some releases. We should possibly also add a note on how the numbers are calculated - like is it numbers before/after instance expansion (like multiple seats). It could actually be an idea to present both.

image


Concerning where to have the sources for the images/grahics, I think it would be great if we could have it locally here in this repo and for example build them as part of https://github.com/COVESA/vehicle_signal_specification/blob/master/.github/workflows/docgen.yml

erikbosch:
Replaced by #801

PR #801, opened 2025-01-20
D3 now bypasses Observable Notebook framework

Giaccomole:
Now, the 3 interactive plots are being generated via HTML Shortcodes with Javascript, directly in the codebase. This enables to edit and deploy without depending on observable.

I made the tweaks that @erikbosch asked in the latest pull request (D3-Statistics). 

I find that the static plots were redundant, but I also succeeded to make them generated directly from the codebase. 

Here are the screenshots:

Screenshot 2025-01-20 at 11 13 00
Screenshot 2025-01-20 at 11 12 47
Screenshot 2025-01-20 at 11 12 36


erikbosch:
Great!

So this means this PR replaces #800, right?

I tested locally, seems to work as expected. What I would like to see is a short instruction on what to do when preparing a release, like which things needs to be changed when we release 5.1 or 6.0, either in the docs README file or in the wiki. Like do we need to regenerate the *png (or where they obsolete anyway?) and if so how.

As you do not have edit access to the wiki I assume adding a section to the README is the easiest solution.

https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/README.md
https://github.com/COVESA/vehicle_signal_specification/wiki/Release-Instructions-and-Checklist

Giaccomole:
Indeed, this replaces #800. 

I can put this info into README as you said too, but to summarize it is like this:

1. I exported static .csv and .json files from the VSS with custom python scripts. The files are included in static/data in docs. This is the interim solution, I will change this flow so the data can be updated in each version update. Until then I can include the scripts, when I refine them. I will update the README too.

2. The website now has 4 less graphics than #800. The hierarchical bar chart and 3 static *png files.

The 3 *pngs below contain the same information as the interactive plots: 
Screenshot 2025-01-20 at 14 11 43

I already wrote shortcodes for these images to be generated from the data too, but they are redundant and they might look dirtier when VSS is updated(widths and proportions etc.). So I deprecated them.

Hierarchical bar chart is the last graphic in #800, I needed to work on it way more than others to carry over the d3-notebook into a html shortcode, but I don’t know if it’s worth the effort.

SebastianSchildt:
I like what i see 👍  If it does not break our website we should merge it. I have not actually checked whether it breaks or has the potential to break something.....

erikbosch:
MoM:

- Merge decision
- Fix remaining topics, if any, afterwards

PR #808, opened 2025-02-05
D3 bypass observable - stats corrected

Giaccomole:
Corrected a mistake in stats. See the file.


docs-gen/layouts/shortcodes/piechart.html
```diff
@@ -110,59 +109,60 @@
...
+    selectorContainer.appendChild(radioForm);
+    console.log("Dataset selector created successfully");
   }).catch(error => {
     console.error("Error loading CSV data:", error);
     document.getElementById("donut").innerText = "Error loading data. Please check the console for details.";
   });
 
   console.log("Script execution complete");
-</script>
+</script>
```
erikbosch, line 168: I think this the reason that pre-commit complains, that it as of today expects a line break at end of file

docs-gen/static/data/piechart.csv
```diff
@@ -0,0 +1,5 @@
+Type,V2,V3,V4,V5,V6
```
erikbosch, line 1: Maybe we should discuss a bit what we want to use as headings. I am thinking if it would be better to specify exact version like v4.2 rather than v4 (even if we have a v4 tag in git, which should point to latest v4.x release). I assume V6 refers to "latest master", but right now it is not dynamically calculated right? I.e. it refers to master at some time in the past?

(And nothing that needs to be addressed now - I am more thinking on how we want to handle it in the future. Like using a "tag" `master-2025-02-06`, `commit-sha-id` or something similar to show origin if it is manually created, or just "master" if it is generated on the fly on merge)

erikbosch:
MoM:

- Please review

docs-gen/static/data/piechart.csv
Giaccomole: +I assume V6 refers to "latest master", it refers to master at some time in the past?
 - Yes, it refers to the point in time where the datafile behind the graph was created. The .csv is not dynamically created yet, but the graph now renders new radio buttons for new versions (like V7 in the future) relying on the .csv
 
 I built an automated logic on this .csv as of last week, but it waits for some review. Here is the simplified code:

    if 'Vehicle.VersionVSS.Major' in row['Signal'] and row['Default'] > 5:
            major_version = int(row['Default'])

    if (major_version is not None):
        
        type_counts 
        column_name = f'V{major_version}'
        version_cols = [col for col in latest.columns if col.startswith('V')]
        for i in range(len(version_cols) - 1):
            v1 = int(version_cols[i][1])
            v2 = int(version_cols[i+1][1])
            if v2 != v1 + 1:
                print(f"MISSING VERSION: V{v1+1}")

So retrospectively for the first 5 versions, I just collected the VX.0 from the documentation, stays as a static template in the code. But starting with V6 this logic counts the types dynamically and concatonates. But keeps the heading as the major version number, even if the minor version changes. I could change this to specify the minor version on the chart. But it could result in something like:

V2 V3 V4 V5 V6.12 V7.3 V8.1

Giaccomole: PS: I built new exporters to vss-tools and modified the docgen.yml in VSS GitHub Actions to generate this in each commit.

https://github.com/Giaccomole/vehicle_signal_specification/tree/Stats-git-actions
https://github.com/Giaccomole/vss-tools-statistics-py-scripts/tree/stat-scripts

.github/workflows/docgen.yml
```diff
@@ -16,6 +16,27 @@ jobs:
         with:
           submodules: true
 
+      - name: Clone vss-tools-statistics-py-scripts repository
+        run: |
+          git clone -b stat-scripts https://github.com/Giaccomole/vss-tools-statistics-py-scripts
```
sschleemilch, line 21: Should just use `pip install` on the vss-tools version we want to use here. No uv required

.github/workflows/docgen.yml
```diff
@@ -16,6 +16,27 @@ jobs:
...
+          cd vss-tools-statistics-py-scripts
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          uv sync
+          uv run vspec stats piechart -s ../spec/VehicleSignalSpecification.vspec -o ../docs-gen/static/data/piechart.csv
+          uv run vspec stats sankey -s ../spec/VehicleSignalSpecification.vspec -o ../docs-gen/static/data/sankey.csv
+          uv run vspec stats radial -s ../spec/VehicleSignalSpecification.vspec -o ../docs-gen/static/data/radial.json
+          cd ..
```
sschleemilch, line 28: No need to cd into and out of it when using `pip install`

docs-gen/static/data/piechart.csv
sschleemilch, line 1: Not sure we even need historic data? If one is interested in historic data, just checkout the tag and build the static site

erikbosch:
Replaced by #819

PR #811, opened 2025-02-12
Missing "s" in vehicle_signal_specification/docs-gen/content/rule_set…

Herrmann-Alexander:
Found that there was an "s" missing on the github.io page while reading the rule set.


erikbosch (approved):
LGTM

PR #820, opened 2025-03-31
fix: fixed units for most recent vss-tools compatability

sschleemilch:


erikbosch:
Noticed the failing build. Maybe it is worth trying specifying `ubuntu-22.04` as builder in this old branch, if the error depends on that ubuntu-latest (i.e. now ubuntu-24.04) does not support the workflow as specified today.

erikbosch (commented):
LGTM.

But this is for the 4.X branch and is just a "copy" of #758, right? I.e. reflecting what is already on master.

erikbosch:
@sschleemilch - I merged and tagged it v4.2.1 (and moved the v4 tag). Please double-check that the tagging seems to be correct from your perspective.

PR #821, opened 2025-04-24
Update with Seating Capability Data

paulboyes:
Updating with Seating Capability Data

### DIFF
--- a/spec/Cabin/Seat.vspec
+++ b/spec/Cabin/Seat.vspec
@@ -31,7 +31,7 @@ IsOccupied:
 OccupancyStatus:
   datatype: string
   type: sensor
-  allowed: ['UNKNOW', 'OCCUPIED', 'EMPTY']
+  allowed: ['UNKNOWN', 'OCCUPIED', 'EMPTY']
   description: Occupancy status of the seat.
 
 
@@ -91,8 +91,8 @@ MassageStatus:
 SupportedMassageTypes:
   datatype: string[]
   type: attribute
-  allowed:  ['ROLL','KNEAD','ANTI_FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
   description:  Type of massage.
+  comment: OEMs may define a list of identifiers for supported massage types using the allowed keyword
 
 MassageType:
   datatype: string
--- a/spec/Cabin/Seat.vspec
+++ b/spec/Cabin/Seat.vspec
@@ -97,9 +97,8 @@ SupportedMassageTypes:
 MassageType:
   datatype: string
   type: actuator
-  allowed:  ['ROLL','KNEAD','ANTI_FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
   description: Type of massage active.  
-  comment:  Must match the SupportedMassageTypes.  
+  comment: OEMs may define a list of identifiers for supported massage types using the allowed keyword 
 
 Position:
   datatype: uint16
--- a/spec/Cabin/Seat.vspec
+++ b/spec/Cabin/Seat.vspec
@@ -91,7 +91,7 @@ MassageStatus:
 SupportedMassageTypes:
   datatype: string[]
   type: attribute
-  ALLOWED:  ['ROLL','KNEAD','ANTI FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
+  allowed:  ['ROLL','KNEAD','ANTI FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
   description:  Type of massage.
 
 MassageType:
--- a/spec/Cabin/Seat.vspec
+++ b/spec/Cabin/Seat.vspec
@@ -28,6 +28,13 @@ IsOccupied:
   type: sensor
   description: Does the seat have a passenger in it.
 
+OccupancyStatus:
+  datatype: string
+  type: sensor
+  allowed: ['Unknown', 'Occupied', 'Empty']
+  description: Occupancy status of the seat.
+
+
 Occupant:
   deprecation: v5.0 - use data from Vehicle.Occupant.
   type: branch
@@ -53,6 +60,11 @@ SeatBeltHeight:
 
 #include include/ItemHeatingCooling.vspec
 
+HasMassageCapability:
+  datatype: boolean
+  type: attribute
+  description: Is the belt engaged.
+
 Massage:
   datatype: uint8
   type: actuator
@@ -70,6 +82,25 @@ MassageLevel:
   unit: percent
   description: Seat massage level. 0 = off. 100 = max massage.
 
+MassageStatus:
+  datatype: string
+  type: sensor
+  allowed: ['ON', 'OFF']
+  description: Massage status.  
+
+SupportedMassageTypes:
+  datatype: string[]
+  type: attribute
+  allowed:  ['Off','Roll','Knead','Anti Fatigue','Pulse','Stretch','Lumbar','Shoulder','Wave','Relief','Glide']
+  description:  Type of massage.
+
+MassageType:
+  datatype: string
+  type: actuator
+  allowed:  ['Off','Roll','Knead','Anti Fatigue','Pulse','Stretch','Lumbar','Shoulder','Wave','Relief','Glide']
+  description: Type of massage active.  
+  comment:  Must match the SupportedMassageTypes.  
+
 Position:
   datatype: uint16
   type: actuator
@@ -77,6 +108,7 @@ Position:
   unit: mm
   description: Seat position on vehicle x-axis. Position is relative to the frontmost position supported by the seat.
                0 = Frontmost position supported.
+
 
 Height:
   datatype: uint16
@@ -101,6 +133,8 @@ Backrest:
   type: branch
   description: Describes signals related to the backrest of the seat.
 
+#include include/ItemHeatingCooling.vspec Backrest
+
 Backrest.Recline:
   datatype: float
   type: actuator
@@ -111,6 +145,8 @@ Backrest.Recline:
   comment:     Seat z-axis depends on seat tilt. This means that movement of backrest due to seat tilting will not
                affect Backrest.Recline as long as the angle between Seating and Backrest are constant.
                Absolute recline relative to vehicle z-axis can be calculated as Tilt + Backrest.Recline.
+
+
 
 Backrest.Lumbar:
   type: branch
@@ -133,6 +169,30 @@ Backrest.LumbarSupport:
   min: 0
   max: 100
   description: Lumbar support (in/out position). 0 = Innermost position. 100 = Outermost position.
+
+Backrest.LumbarSupportTop:
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Lumbar top support (in/out position). 0 = Innermost position. 100 = Outermost position.
+
+Backrest.LumbarSupportMid:
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Lumbar mid support (in/out position). 0 = Innermost position. 100 = Outermost position.
+
+Backrest.LumbarSupportBottom:
+  datatype: float
+  type: actuator
+  unit: percent
+  min: 0
+  max: 100
+  description: Lumbar bottom support (in/out position). 0 = Innermost position. 100 = Outermost position.
 
 Backrest.Lumbar.Height:
   datatype: uint8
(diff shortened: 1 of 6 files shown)


spec/Cabin/Seat.vspec
```diff
@@ -28,6 +28,13 @@ IsOccupied:
   type: sensor
   description: Does the seat have a passenger in it.
 
+OccupancyStatus:
+  datatype: string
+  type: sensor
+  allowed: ['Unknown', 'Occupied', 'Empty']
```
erikbosch, line 34: Historically we mostly used "enum style identifiers" for allowed, like:

`allowed: ['UP', 'MIDDLE', 'DOWN']`


i.e. upper case and if needed underscore

spec/Cabin/Seat.vspec
```diff
@@ -53,6 +60,11 @@ SeatBeltHeight:
 
 #include include/ItemHeatingCooling.vspec
 
+HasMassageCapability:
+  datatype: boolean
+  type: attribute
+  description: Is the belt engaged.
```
erikbosch, line 66: Copy/paste error?

spec/Cabin/Seat.vspec
paulboyes: Got it.  I will update

erikbosch: We already have IsOccupied above, so why do we need this one?

Historically we have avoid "Unknown" in VSS signals, assuming that the API has a mechanism to say that it does not have any value for Seat.IsOccupied. Is the intention of this one just to handle the case that the vehicle cannot say whether there is a passenger (=Human?) in the seat?

spec/Cabin/Seat.vspec
```diff
@@ -28,6 +28,13 @@ IsOccupied:
   type: sensor
   description: Does the seat have a passenger in it.
 
+OccupancyStatus:
+  datatype: string
+  type: sensor
+  allowed: ['Unknown', 'Occupied', 'Empty']
+  description: Occupancy status of the seat.
```
erikbosch, line 35: And here I believe we talk about a human? Or is a set occupied also if there is a jacket, a bag, a dog or a gorilla in the seat. Or si that up to the OEM to decide?

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+MassageStatus:
+  datatype: string
+  type: sensor
+  allowed: ['ON', 'OFF']
+  description: Massage status.  
+
+SupportedMassageTypes:
+  datatype: string[]
+  type: attribute
+  allowed:  ['Off','Roll','Knead','Anti Fatigue','Pulse','Stretch','Lumbar','Shoulder','Wave','Relief','Glide']
```
erikbosch, line 94: At least for me it would be good to have a short description of each term. Or is it up to OEM to define the meaning of them?

Also - no need for OFF here?

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+  description:  Type of massage.
+
+MassageType:
+  datatype: string
+  type: actuator
+  allowed:  ['Off','Roll','Knead','Anti Fatigue','Pulse','Stretch','Lumbar','Shoulder','Wave','Relief','Glide']
```
erikbosch, line 100: Is the idea that if someone turns the massage off they should both:

- Set MassageStatus to OFF
- Set MassageType to OFF

I.e. in practice it should never be so that only one of them are off?

spec/Cabin/Seat.vspec
paulboyes: Yes. The reason I added OccupancyStatus is because of the UNKNOWN status.  The vehicle does not know if the seat is occupied.

erikbosch: Could be worth bringing it up to discussion again, as it might have a general interest.

Historically, for "stupid simple sensor based signals" like `IsOpen` there has only been True/False. Similar, for `Vehicle.Speed` there is no value to indicate that speed is unknown. Assumption is that in an API if the vehicle by some reason does not know the status it would do something like:

- If doing "read" you would get "Not found" or "Null"
- If subscribing you would either get "Null" when value become unknown, or just stop getting updates, i.e. the old value would expire on client side

So that is one aspect - do we want to keep that approach?

The other aspect is that the underlying sensor maybe no longer can be considered as "stupid simple sensor" but rather is more intelligent, and that `Unknown` then rather is `Inconclusive`. I.e. the system is working, but it can not really say whether there is a person in the seat or if it just a bag.

spec/include/ItemFanSpeed.vspec
```diff
@@ -0,0 +1,13 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+FanSpeed:
+  datatype: string
+  type: actuator
+  allowed: ['AUTO', 'Level1', 'Level2','Level3','Level4','Level5']
```
erikbosch, line 12: Suggest upper case, like `LEVEL_1`or `LEVEL1`. Also add a newline to the end of the file.

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+MassageStatus:
+  datatype: string
+  type: sensor
+  allowed: ['ON', 'OFF']
+  description: Massage status.  
+
+SupportedMassageTypes:
+  datatype: string[]
+  type: attribute
+  allowed:  ['ROLL','KNEAD','ANTI FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
```
erikbosch, line 94: One error reported by CI concerns the `vss2ddsidl` exporter. In short it cannot handle whitespace in identifiers, like `ANTI FATIGUE`. It is valid VSS though, so one could see it as a defect in the exporter as VSS allowed values NOR are the same as enums. Anyhow, I suggest just changing the name to `ANTI_FATIGUE` to follow classic enum syntax.

```
Successfully installed cyclonedds-0.10.5
/home/runner/work/vehicle_signal_specification/vehicle_signal_specification/vss.idl:3828:50: syntax error
/home/runner/work/vehicle_signal_specification/vehicle_signal_specification/vss.idl:3852: fatal error: File write error
    };
```

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+  description:  Type of massage.
+
+MassageType:
+  datatype: string
+  type: actuator
+  allowed:  ['ROLL','KNEAD','ANTI FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
```
erikbosch, line 100: Add underscore here as well

spec/Cabin/Seat.vspec
erikbosch: MoM:

- Melissa: Keep definition as OEM specific
- Sunil: Signal interesting for Toyota
- Erik: Please compare with IsOccupied above, do not like us to have both
- Melissa: prefer that there is an unknown value
- Sebastian: In VSS we often do not modify "undefined/unknown"

**Daniel:**

You can model both,

1) If the seat is occupied 
2) the list of things that are detected to be on the seat by whatever mean

I don’t see a problem with that.

**Stefan:**

ISO 7000 https://www.iso.org/obp/ui#iso:grs:7000:2960

in car monitoring is mandatory in Europe (and can be done by camera) [Euro NCAP requirements | VDA](https://www.vda.de/en/topics/automotive-industry/standardization-and-technical-standards/euro-ncap-anforderungen)

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+MassageStatus:
+  datatype: string
+  type: sensor
+  allowed: ['ON', 'OFF']
+  description: Massage status.  
+
+SupportedMassageTypes:
+  datatype: string[]
+  type: attribute
+  allowed:  ['ROLL','KNEAD','ANTI_FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
```
erikbosch, line 94: MoM:

- M: Do we think everyone would have the same pattern? Or ask the vehicle which pattern it has?
- E: An alternative would be to not standardize the list, but leave it to OEMs
- S: We had descriptions from technical and marketing perspective, and they differ, so complex to align

erikbosch:
MoM:

- Please review, to be discussed in later meetings

spec/Cabin/Seat.vspec
erikbosch: MoM:

- E: We need to decide if we want to have the list of "allowed" value in the standard catalogue.
- D: We cannot model which  capabilities a vehicle has with current VSS.
- D: The terms actuator/sensor is misleading

erikbosch: As it seems from the discussion that the list will not be "standardized" I suggest that we remove the `allowed` list and instead add a comment, possibly like:

`comment: OEMs may define a list of identifiers for supported massage types using the allowed keyword`

erikbosch: MoM:

- Melissa: Agrees to remove
- Melissa: Do we need to formalize display name? With human readable name

spec/Cabin/Seat.vspec
erikbosch: MoM:

- Decision: OK to keep "unknown"

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
   unit: percent
   description: Seat massage level. 0 = off. 100 = max massage.
 
+MassageStatus:
```
sschleemilch, line 85: `Massage.Status`?

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+  datatype: string
+  type: sensor
+  allowed: ['ON', 'OFF']
+  description: Massage status.  
+
+SupportedMassageTypes:
```
sschleemilch, line 91: `Massage.Types` maybe

spec/Cabin/Seat.vspec
```diff
@@ -70,6 +82,25 @@ MassageLevel:
...
+  datatype: string[]
+  type: attribute
+  allowed:  ['ROLL','KNEAD','ANTI_FATIGUE','PULSE','STRETCH','LUMBAR','SHOULDER','WAVE','RELIEF','GLIDE']
+  description:  Type of massage.
+
+MassageType:
```
sschleemilch, line 97: `Massage.Type`?

spec/include/ItemFanSpeed.vspec
```diff
@@ -0,0 +1,14 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+FanSpeed:
+  datatype: string
+  type: actuator
+  allowed: ['AUTO', 'LEVEL1', 'LEVEL2','LEVEL3','LEVEL4','LEVEL5']
```
sschleemilch, line 12: I feel like a better fit would be 0-100 (percentage) or a number for the RPM for a fan speed. What do you think?

erikbosch:
- MoM:
  - Paul to review latest comments
  - Paul to check naming concept

slawr (commented):
Hi Paul, these commits do not have a signed-off in the commit message. You could add that with a forced push before they are merged.

paulboyes:
> Hi Paul, these commits do not have a signed-off in the commit message. You could add that with a forced push before they are merged.

signed-off  added.

spec/Cabin/Seat.vspec
erikbosch: `Massage` a few lines up is a signal, so right now we cannot create a new branch `Massage`. Also in the discussion where @jdacoello has been involved we have tended to go more to having multiple signals on the same level rather than having them grouped in branches, but a topic to align on the meeting.

spec/Cabin/Seat.vspec
```diff
@@ -28,6 +28,13 @@ IsOccupied:
   type: sensor
   description: Does the seat have a passenger in it.
 
+OccupancyStatus:
```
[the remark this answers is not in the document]
erikbosch: As I have understood the discussions the unknown field is intended ti handle the "inconclusive" state, like indicating that the system monitoring occupancy is working but could not make any conclusion for this seat.

erikbosch:
Created a partially reworked version in #838 based on the ideas from @jdacoello and @sschleemilch. Wants to discuss it with @paulboyes though.

Issue #822, opened 2025-05-05
DistanceTraveled granularity


sophokles73:
The *Vehicle.DistanceTraveled* is currently measured in kilometers. Is there a particular reason why it is not measured in meters? The Fleet Management Standard requires the total distance traveled to be tracked at meter granularity. J1939 includes SPN917 (Total Vehicle Distance (High Resolution)) for this purpose and it would be nice if we were able to map this into VSS without losing precision.

erikbosch:
The signal has been there for a long time and I assume that `km` was selected as that is the typical unit used when describing a vehicle in a used car ad or when displaying the value to the driver. It is an VSS float value (IEEE 754-2008 binary32 floating point number) so if using the default VSS type it might not be able to keep meter precision if the vehicle has traveled a long distance.

```
TraveledDistance:
  datatype: float
  type: sensor
  unit: km
  description: Odometer reading, total distance traveled during the lifetime of the vehicle.
```

We have in general historically been positive to changing datatype/units when needed. One option would be to change to meters and change datatype to uint32 (or uint64 if we think uint32 does not suffice). Another option would be to change datatype to double, that would give sufficient significant digits to support meter precision regardless of whether using meter or kilometer as unit.

erikbosch:
MoM:

- presented at weekly meeting.

SebastianSchildt:
To see what precision you have for larger values with float/double this is helpful: https://float.exposed/

sophokles73:
Having played around a little with the link provided by @SebastianSchildt I guess that I would be in favor of changing the unit to meters and using a uint32 instead of using a float/double.
A uint32 would support distances up to 4.29 million kilometers. That should be sufficient even for commercial vehicles which tend to be on the road longer than the average passenger car, shouldn't it?

erikbosch:
Feel free to open a PR, then people have to say yes/no.

erikbosch:
MoM:

- Brought up by Ted, increased granularity needed for particular signals
- There are two proposed changes, traveled distance (#824) and #827 (engine speed)
-

erikbosch:
Unit has been changed to meters, closing this

PR #824, opened 2025-05-09
Change granularity of TraveledDistance data entry

sophokles73:
The TraveledDistance and its accompanying data entries (TraveledDistanceSinceStart, TripMeterReading) have been defined using kilometer granularity.

However, use cases in the commercial vehicle domain often require a granularity down to meters. The existing definition using the float data type theoretically also supports specifying distances with meter granularity, but does so only with the usual shortcomings of encoding float values in computer systems (rounding issues etc)

Changing the unit to meters and the data type to uint32 would allow the usage of the data entries in the use cases mentioned above, i.e. without the need to create an overlay for adapting them, and would also remove the possibility to have negative values, which do not seem to be consistent with semantics of these data entries.

This addresses #822


sschleemilch:
Ok for me although of course a breaking change. Still I would be fine with it changing internal dependencies or define an overlay to get back to km

erikbosch (approved):
Ok for me. Should be considered as backward incompatible change and not to be included in any 5.X release.

erikbosch:
MoM:

- Please review
- Melissa: Wants to check internally

Melmrr (commented):
from Melissa (Ford) - 
I checked with some folks internally.
Currently we allow a Value with a single decimal and then units to be stored / passed in the vehicle,  Example 23.1, kilometers.

They expressed concerns about switching to meters in terms of required In Vehicle Software change, memory needed to store, and possible CAN bus bandwidth. 

They also wanted to know what/if the SAE states a standard for this, in,[ J2976_202205 - Speedometer and Odometer Systems for On-Highway Vehicles, ](https://www.sae.org/standards/content/j2976_202205/)

I haven't gotten a chance to access and check myself yet.

erikbosch:
MoM:

- Daniel: Import to make a distinction, what do we mean with TraveledDistance. Change datatype is just a convention, we should not change to keep it stable. Do not see it as big advantage to change unit, we can get other requests for changes if we acept it.
- Paul: Are we using km
- D: A solution could be to agree to dimension, rather than unit. Do we need to be more prescriptive to quantity, but descriptive when it concerns unit.
- E: For APIs like Kuksa & VISS you expect to get value in default unit
- D: Better to explicit in quantity kind rather than unit
- E: We have unit/quantity files, we can specify recommendations how to use a different unit. Better to refer to something else.
- E: Conclusion seems to be not to accept PR, but rather give recommendations on how to handle it in other ways or how to extend VSS
- D: A quick fix would be to specify "quantity" (e.g. length) rather than unit, that in tooling has mapping to an enum
- D: 3 action points
- 1. Check how we can reuse existing standards, including conversion, D can investigate
- 2. Using enum in tools - possibly let Sebastian Schleemilch
- 3. Include in documentation a how to section on how this solved. Like how can I use this property If am working with a different unit.

adobekan:
For BMW this would be a breaking change.

erikbosch:
MoM:

- S: Related to CV work?
- T: CV group looking at overlay or separate signal branch
- E: Possibly wait on what happens in CV group.

erikbosch:
MoM:

- Topic brought up by Ted
- T: We really want this change
- E: You just mention two types of signals (travelled distance and engine speed), so ok for me to change them if we have a common view that they will be "better" after the change
- E: Proposal - lets discuss this PR once more next week, to see if we have an ok for merge. Please comment in PR if you have a strong objection.

PR #838, opened 2025-07-24
PR821 alternative solution

erikbosch:
This is a partially reworked version of #821 -for discussion.

It incorporate the idea of having a SeatMassage Branch where the following signals are put.

```
vss.csv:1123:Vehicle.Cabin.Seat.Row2.PassengerSide.SeatMassage,branch,,,,,,Massage related information for the seat.,,,
vss.csv:1124:Vehicle.Cabin.Seat.Row2.PassengerSide.SeatMassage.IsAvailable,attribute,boolean,,,,,True if the seat have the massage capability,,,
vss.csv:1125:Vehicle.Cabin.Seat.Row2.PassengerSide.SeatMassage.Level,actuator,uint8,,percent,0,100,Seat massage level. 0 = off. 100 = max massage.,,,
vss.csv:1126:Vehicle.Cabin.Seat.Row2.PassengerSide.SeatMassage.Status,sensor,string,,,,,Massage status.,,"['ON', 'OFF']",
vss.csv:1127:Vehicle.Cabin.Seat.Row2.PassengerSide.SeatMassage.SupportedTypes,attribute,string[],,,,,Type of massage.,OEMs may define a list of identifiers for supported massage types using the allowed keyword,,
vss.csv:1128:Vehicle.Cabin.Seat.Row2.PassengerSide.SeatMassage.TypeActive,actuator,string,,,,,Type of massage active.,OEMs may define a list of identifiers for supported massage types using the allowed keyword,,
```

I have also just renamed/removed old signals which makes sense if we think should this should go to the next major only. So as of now this is a backward incompatible change, but we could handle it as backward compatible as well but does not help much if the next planned release after 5.1 anyway will be 6.0 where we will remove deprecated stuff.

I have merged all changes/commits from Paul

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ String types can now be restricted by a RegExp pattern.
 Previously the documentation stated that default values could not be given for struct types.
 That is now supported.
 
+### Signal Updates
+
+* Signals for seats refacored (backward incompatible change)
+
 ## VSS 5.0
 
 ### Signal Updates
--- a/spec/Cabin/Seat.vspec
+++ b/spec/Cabin/Seat.vspec
@@ -23,10 +23,11 @@
 # z-axis is the axis going up/down on the vehicle
 #
 
-IsOccupied:
-  datatype: boolean
+OccupancyStatus:
+  datatype: string
   type: sensor
-  description: Does the seat have a passenger in it.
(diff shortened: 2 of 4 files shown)


erikbosch:
MoM:

- Erik presented this PR, same as Puals with some minor update
- Melissa: Will take a look at it. Will give comment.
- Martin: Looks fine 
- Paul: I am good with it
- OK to merge if no remarks from Melissa

PR #839, opened 2025-07-24
Extend tire temperature signals

### LINKED ISSUE #836: Differentiation between TPMS and TMS sensors value
Hello VSS Team !

When looking at the signal `Vehicle.Chassis.Axle.Row1.Wheel.Left.Tire.Temperature`, one could ask about the nature of the temperature: are we measuring the temperature of the rubber, or the air inside the tire.

With my team, we deal with cars equipped with TPMS or TMS and that would be a nice addition to differentiate if the temperature returned by the sensors is the temperatore of  the air or the rubber.

What we are proposing is:

- Proposition n°1:
    - We replace the field `Temperature` with `TireTemperature` that would describe the temperature of the rubber and add a new field `AirTemperature` that would describe the temperature of the air inside the tire.

```
Tire.TireTemperature:
  datatype: float
  type: sensor
  unit: celsius
  description: Rubber temperature of the tire in Celsius.

Tire.AirTemperature:
  datatype: float
  type: sensor
  unit: celsius
  description: Air temperature inside the tire in Celsius.
```

- Proposition n° 2:
    - We keep the field `Temperature`, just change the description and add the new field `AirTemperature`

```
Tire.Temperature:
  datatype: float
  type: sensor
  unit: celsius
  description: Rubber temperature of the tire in Celsius.

Tire.AirTemperature:
  datatype: float
  type: sensor
  unit: celsius
  description: Air temperature inside the tire in Celsius.
```

As I understood, the guidelines would point out more towards the second proposition. Lemme what you think about this change
Cheers,

Jérémy

erikbosch:
At the VSS meeting this week we said that it could be an idea to assist @JMorceaux  with a possible PR for the proposed changes, to show a possible implementation.

In this PR I propose to create two new signals and deprecate the old one. Deprecated signals might be removed in major releases but we can keep them longer if we see that as relevant. That would give users time to change to the new signals.

If I am thinking if `Tire.RubberTemperature` might be a better name than `Tire.TireTemperature` as the latter name might be somewhat ambiguous even if the description states that it concern rubber temperature, but I am open to use  `Tire.TireTemperature` if that is preferred

Fixes #836

### DIFF
--- a/spec/Chassis/Wheel.vspec
+++ b/spec/Chassis/Wheel.vspec
@@ -50,7 +50,6 @@ Tire.Pressure:
   type: sensor
   unit: kPa
   description: Tire pressure in kilo-Pascal.
-
 Tire.IsPressureLow:
   datatype: boolean
   type: sensor
@@ -61,7 +60,19 @@ Tire.Temperature:
   type: sensor
   unit: celsius
   description: Tire temperature in Celsius.
+  deprecation: v6.0 - use RubberTemperature or AirTemperature instead.
 
+Tire.RubberTemperature:
+  datatype: float
+  type: sensor
+  unit: celsius
+  description: Rubber temperature of the tire in Celsius.
+
+Tire.AirTemperature:
+  datatype: float
+  type: sensor
+  unit: celsius
+  description: Air temperature inside the tire in Celsius.
 
 #
 # Wheel signals


SebastianSchildt:
MoM
 - Ok to merge

PR #840, opened 2025-07-29
Update TorqueElectricMinimum max value to an integer

yf6:
0.0 makes kuksa angry.


erikbosch (approved):
LGTM

As datatype is an int-type having max/min/default in float syntax is at least ugly, and could possibly be considered as a syntax error. I will add an issue to vss-tools

PR #842, opened 2025-07-30
fix: TorqueElectricMinimum max -> int

sschleemilch:
As a result of fixing https://github.com/COVESA/vss-tools/issues/460, need to fix the model here as well


erikbosch (approved):
LGTM. Thanks!

PR #849, opened 2025-09-19
changes unit of Battery Capacities from kWh to Ah TractionBattery.vspec

DavidMatzek:
# About
hWh is not a Battery Capacity. Since Batteries can have different Voltages kWh is not a good unit. 
kWh tells somethin about Power which is P = U * I .

## Use Case
We want to use NetCapacity to fulfill EU Battery Pass regulations. Unit of Ah is expected for remaining Capacity. 

Signed-off-by: David Matzek<david.dm.matzek@bmw.de>

### DIFF
--- a/spec/Powertrain/TractionBattery.vspec
+++ b/spec/Powertrain/TractionBattery.vspec
@@ -110,13 +110,13 @@ CellVoltage.CellVoltages:
 GrossCapacity:
   datatype: uint16
   type: attribute
-  unit: kWh
+  unit: Ah
   description: Gross capacity of the battery.
 
 NetCapacity:
   datatype: uint16
   type: sensor
-  unit: kWh
+  unit: Ah
   description: Total net capacity of the battery considering aging.
 
 StateOfHealth:


DavidMatzek:
Signed-off-by: Author Name <authoremail@example.com> was missing in commit.

erikbosch:
A bit fun that the [EU Battery_Passport_Content_Guidance](https://thebatterypass.eu/assets/images/content-guidance/pdf/2023_Battery_Passport_Content_Guidance.pdf) isn't consistent, as it typically use Ah for Capacity, except when they mention the 2 kWh limit:

_According to Article 10, all “LMT batteries, rechargeable industrial batteries with a capacity greater
than 2 kWh, and electric vehicle batteries shall be accompanied by a document containing values for
the electrochemical performance and durability parameters”, already by 18 August 2024_

Anyhow that document is superseded by DIN DKE SPEC 99100:2025-02 and there it seems more consistent, so should we possibly use that as reference? If so 4 good names would be

- RatedCapacity (6.7.2.2, Ah, Static, Decimal)
- RemainingCapacity (6.7.2.3, Ah, Dynamic, Decimal)
- CertifiedUsableBatteryEnergy (6.7.2.5, kWh, Static, Decimal)
- RemainingUsableBatteryEnergy (6.7.2.6, kWh, Dynamic, Decimal)

One idea, if we think the names above are good is to create 4 new signals based on the information above (using float to support decimals), and mark the existing ones as deprecated and to be removed in a later release. Then we can handle the change as a "minor change" and if there are users of the existing signals they get some time to shift to the new signals. It is always good to have a reference to a standard, as that prevents that the discussion comes up again.

PR #865, opened 2025-11-05
Add Extended Range Electric Vehicle (EREV) signal definitions.

ChoiSeokHeum:
Add Extended Range Electric Vehicle (EREV) signal definitions.


erikbosch (approved):
LGTM

Issue #859, opened 2025-10-17
Questions from FMS


erikbosch:
After you have left our meeting on 15.10.2025, we raised some more questions concerning the integration of FMS in VSS.
I think you could give us answers from COVESA point of view.
Here are our questions:
1.	 In the COVESA standard catalog VIN is defined as following:
```

VehicleIdentification.VIN:
  datatype: string
  type: attribute
  description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
  pattern: ^([0-9A-HJ-NPR-Z]{3})([0-9A-HJ-NPR-Z]{6})([0-9A-HJ-NPR-Z]{4}[0-9]{4})$

```
The pattern attribute is removed by the vspec, should we upgrade the vspec tool to the latest version or is it something also that is wrong?
2.	We have discussed these two styles to express enums: 
Like strings:
```
SupportedAutonomyLevel:
  datatype: string
  type: attribute
  allowed: [
    'SAE_0', # No Driving Automation
    'SAE_1', # Driver Assistance
    'SAE_2', # Partial Driving Automation
    'SAE_3', # Conditional Driving Automation
    'SAE_4', # High Driving Automation
    'SAE_5'  # Full Driving Automation
  ]
  description: Indicates the highest level of driving automation according to the SAE J3016 taxonomy the vehicle is capable of.
```
Or like uints:

```
Vehicle.Cabin.Tachograph.Drivers.Driver2.WorkingState:
  allowed:
  - 0
  - 1
  - 2
  - 3
  comment: ¤FMS¤ Corresponds SAE J1939 signal PGN 0xFE6C(TCO1)/SPN 1613 in ACEA FMS-Standard description version 5.0 chapter 
1.1.8
  datatype: uint8
  description: State of work of the driver 2. 0 = Rest, 1 = Driver available, 2 = Work, 3 = Drive
  type: sensor

```
Which is the preferable one?
3.	We have discussed the situation with different fuel systems. Truck hybrids can have gaseous fuel system together with diesel one. To structure up these two fuel systems, do we need a separate branch?
For example, `Vehicle.Powertrain.FuelSystemGaseous.TotalConsumptionGaseous` or keep it in the `Vehicle.Powertrain.TotalConsumptionGaseous` ?
4.	The same for retarder:
`Vehicle.Chassis.Brake.Retarder.ActualRetarderPercentTorque` or `Vehicle.Chassis.Brake.ActualRetarderPercentTorque`?

[...]

PR #866, opened 2025-11-05
Add fuel economy metrics

InsangJang-ivis:
Add fuel economy metrics.

### DIFF
--- a/spec/Powertrain/FuelSystem.vspec
+++ b/spec/Powertrain/FuelSystem.vspec
@@ -123,3 +123,42 @@ IsFuelPortFlapOpen:
   datatype: boolean
   type: actuator
   description: Status of the fuel port flap(s). True if at least one is open.
+
+InstantantFuelEconomy:
+  datatype: float
+  type: sensor
+  unit: km/l
+  description: Real-time instantaneous fuel economy calculated over a short time window.
+  comment: This value represents the current fuel efficiency at any given moment during vehicle operation.
+           It is calculated based on the distance traveled and fuel consumed over a brief sampling period (typically around 500ms).
+           The calculation compares the vehicle state before and after the sampling window to determine instantaneous efficiency.
+           This metric refreshes frequently and can fluctuate significantly based on driving conditions, acceleration, and terrain.
+           Unlike average fuel economy measurements, this provides immediate feedback on current driving efficiency.
+
+CumulativeFuelEconomy:
+  datatype: float
+  type: sensor
+  unit: km/l
+  description: Cumulative average fuel economy calculated from vehicle start or last user reset.
+  comment: This represents the accumulated fuel economy over the entire period since the last reset.
+           Unlike trip-based measurements, this provides a long-term average of fuel efficiency.
+
+DriveFuelEconomy:
+  datatype: float
+  type: sensor
+  unit: km/l
+  description: Average fuel economy for the current drive cycle, calculated from engine start to current time.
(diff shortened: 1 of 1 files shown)


spec/Powertrain/FuelSystem.vspec
```diff
@@ -123,3 +123,42 @@ IsFuelPortFlapOpen:
   datatype: boolean
   type: actuator
   description: Status of the fuel port flap(s). True if at least one is open.
+
+InstantaneousFuelEconomy:
```
sschleemilch: I would vote for grouping it in an `Economy` branch:

```yaml
Economy:
  type:  branch
  description: ...

Economy.Live:
  ...

Economy.Cumulative:
  ...

Economy.Drive:
  ...

Economy.SinceRefuel:
  ...
```

spec/Powertrain/FuelSystem.vspec
```diff
@@ -123,3 +123,42 @@ IsFuelPortFlapOpen:
...
+  type: sensor
+  unit: km/l
+  description: Real-time instantaneous fuel economy calculated over a short time window.
+  comment: This value represents the current fuel efficiency at any given moment during vehicle operation.
+           It is calculated based on the distance traveled and fuel consumed over a brief sampling period (typically around 500ms).
+           The calculation compares the vehicle state before and after the sampling window to determine instantaneous efficiency.
+           This metric refreshes frequently and can fluctuate significantly based on driving conditions, acceleration, and terrain.
+           Unlike average fuel economy measurements, this provides immediate feedback on current driving efficiency.
+
+CumulativeFuelEconomy:
```
erikbosch: A bit similar idea as `AverageConsumption` but I like this name better so lets keep it

spec/Powertrain/FuelSystem.vspec
erikbosch: MoM:

- Ted: From FMS /Trucks - you may have multiple fuelsystems, do we want cumulative fueleconomy or separated by fuel tank.
- Erik: Think we should treat the "multi tank", "multi fuel" topic as a separate topic, PR/proposal welcome, related to #859
- T: For ev there is a factor for cumulative, new unit recently added
- E: No need for subbranch
- **Decision: Change `InstantaneousFuelEconomy` to `InstantantFuelEconomy`**
- **Then ok to merge**

erikbosch (changes requested):
Name change needed for one signal

InsangJang-ivis:
> Name change needed for one signal

Changed the signal name from InstantaneousFuelEconomy to InstantantFuelEconomy.

spec/Powertrain/FuelSystem.vspec
```diff
@@ -123,3 +123,42 @@ IsFuelPortFlapOpen:
...
+  datatype: float
+  type: sensor
+  unit: km/l
+  description: Cumulative average fuel economy calculated from vehicle start or last user reset.
```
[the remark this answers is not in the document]
InsangJang-ivis: It should not be automatically reset at 'engine start'. The cumulative fuel economy can only be reset when the vehicle battery is discharged or reset. Generally, during the vehicle's lifetime, the vehicle battery is not discharged or reset, and the battery can be considered to remain intact. Here, 'vehicle start' refers to a state where the battery is maintained without being discharged or reset.

erikbosch (approved):
OK to merge. As minor and previously discussed I will include this one to VSS 6.0 branch as well

PR #867, opened 2025-11-06
Add fahrenheit to unit

ivis-Hyojun:
Add "fahrenheit" to temperature section of units.yaml

### DIFF
--- a/spec/units.yaml
+++ b/spec/units.yaml
@@ -81,6 +81,12 @@ celsius:
   quantity: temperature
   allowed-datatypes: ['numeric']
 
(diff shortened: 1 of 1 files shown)


ivis-Hyojun:
According to #868, the key introduced in this PR, “fahrenheit,” should ideally be updated to “Fahrenheit” for consistency. However, since this decision is still under discussion, I’ll hold off on making that change for now.

ivis-Hyojun:
Capitalization has been updated. 
Thank you for reviewing. 
@erikbosch, @SebastianSchildt

PR #885, opened 2026-03-09
FOR DISCUSSION: Datatype syntax and representation

erikbosch:
We currently have ongoing discussion to add out first datatype (struct) to the catalog tree, as well as to support enums (see #873) and constants. I created this as a proposal on how we potentially could add structs/enums/consts to the catalog, building on our existing struct support. The changes in this PR does not work as of today.

Rough idea:

- Have a types subtree
- In that one refer to definition
- Follow rules from structs on how you can reference elems

Intention with this one is not to discuss the specific structs/enum/consts but rather the used syntax,

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -14,6 +14,16 @@
 # VSS Versioning information
 #
 
+ExampleTimestamp:
+  datatype: VehicleDataTypes.Timestamp
+  type: sensor
+  description: Just an example.
+
+ExampleEnum:
+  datatype: VehicleDataTypes.MyExampleEnum
+  type: sensor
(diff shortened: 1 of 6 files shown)


UlfBj:
I think this is the right way to go. I have the following comments to the proposal:
* According to the [VSS struct documentation](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types_struct/index.html#general-idea-and-basic-semantics) the tree root node shall have the name Types. I think we should stick with that.
* The struct/enum/const definitions should be grouped on separate branches having the names Struct/Enum/Const.

erikbosch:
Closing this one now as timestruct has been added and no pressure that have enums in typetree

PR #891, opened 2026-03-25
Enum syntax documentation

erikbosch:
Related to #873 and https://github.com/COVESA/vss-tools/pull/495. A few topics to discuss:

* Shall it be allowed to "reuse" the same symbolic name twice, like "A:1, B:1, C:2". Maybe better to say no as behavior may be come a bit undefined
* Shall we enforce a name pattern for symbols. Always or just for std catalog. Like that "'123': 456" shall not be allowed as it cannot be used as an enum identifier in many target languages.

### DIFF
--- a/docs-gen/content/rule_set/data_entry/enum.md
+++ b/docs-gen/content/rule_set/data_entry/enum.md
@@ -1,0 +1,82 @@
+---
+title: "Enums"
+date: 2019-08-04T12:37:12+02:00
+weight: 45
+---
+
+
+
+## Specifying a signal as an enum
+
+VSS has introduced `enum` as an alternative to [allowed values](/vehicle_signal_specification/rule_set/data_entry/allowed/).
+In VSS allowed values have typically been used to restrict values for string signals, and most exporters and downstream projects have treated it as a string value.
+With enum the base type is an integer type and exporters and downstream projects are free to decide on how they want to use the symbolic names,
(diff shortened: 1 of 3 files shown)


erikbosch:
MoM:

- Revcew this one and https://github.com/COVESA/vss-tools/pull/495 until end of week. Ok to merge if no remarks

docs-gen/content/rule_set/data_entry/enum.md
```diff
@@ -0,0 +1,71 @@
...
+VSS has introduced `enum` as an alternative to [allowed values](/vehicle_signal_specification/rule_set/data_entry/allowed/).
+In VSS allowed values have typically been used to restrict values for string signals, and most exporters and downstream projects have treated it as a string value.
+With enum the base type is an integer type and exporters and downstream projects are free to decide on how they want to use the symbolic names,
+like if they want to expose the symbolic names in APIs or generated assets, or if they just ignore the symbolic names.
+
+It is however expected that VSS implementations has mechanisms to assure that only values matching the specified enums can be used for the related signals.
+
+## Basic syntax
+
+Enum delcarations can be used for `actuator`, `sensor`, `attribute` and for structs in `property`.
```
JMorceaux: I spotted only 1 typo here, just in case you have the courage to modify your PR ^^'

```suggestion
Enum declarations can be used for `actuator`, `sensor`, `attribute` and for structs in `property`.
```

JMorceaux:
Just read this PR. All seems LGTM (apart from 1 small typo)

PR #894, opened 2026-04-13
Feat/safety 877 v2

vje013:
Contributes to #877 (Connected Safety signals). Adds three derived/smart signals to a new `Vehicle.Safety` branch per @erikbosch's guidance.

Signals added:

- `Vehicle.Safety.IsFire` — boolean sensor. Fire detected in or on the vehicle. Implementation specific. May derive from temperature sensors, battery thermal runaway detection, or dedicated fire/smoke systems.
- `Vehicle.Safety.IsSubmersed` — boolean sensor. Vehicle partially or fully submersed in water. Implementation specific. May derive from pressure sensors, water contact sensors, or door seal sensors.
- `Vehicle.Safety.RolloverDetected` — string sensor enum. Values: `NONE`, `RISK`, `DETECTED`, `UNKNOWN`. Covers both active rollover events and elevated rollover risk per Erik's suggestion. OEM specific criteria. May derive from `Vehicle.AngularVelocity`, `Vehicle.Roll`, and `Vehicle.Pitch`.

Why these signals:

- Connected Safety group identified fire, submersion, and vehicle orientation as unsupported in current VSS
- First responders need these at dispatch time. A 911 operator needs "car is on fire" not raw temperature readings
- All three are derived/smart values computed from underlying sensor data. Fits the `Vehicle.Safety` branch pattern Erik proposed

Placement:

- New `spec/Safety/Safety.vspec` file
- `Vehicle.Safety` branch included in root spec after `Vehicle.ADAS` (pre-crash) and before `Vehicle.Chassis` (structural). Seems like pre-crash prevention and post-crash detection should sit adjacent.

Companion PR:

- Absolute orientation signals (Roll, Pitch) will follow in a separate PR under `Vehicle` near the existing `AngularVelocity` block at line 302 of `Vehicle.vspec`. These are the raw inputs that `RolloverDetected` derives from.

Signed-off-by: Vladimir Edouard <vje013@gmail.com>

PR #897, opened 2026-04-14
feat(vehicle): add Vehicle.Orientation branch with absolute roll, pitch, yaw

vje013:
Contributes to #877 (Connected Safety signals). Adds absolute orientation
signals to complement the existing Vehicle.AngularVelocity (rate) signals,
per guidance from @erikbosch.

Signals added:

- Vehicle.Orientation.Roll — float sensor, degrees. Vehicle tilt around
  X (longitudinal). Positive = right side down.
- Vehicle.Orientation.Pitch — float sensor, degrees. Vehicle tilt around
  Y (lateral). Positive = nose up.
- Vehicle.Orientation.Yaw — float sensor, degrees. Vehicle rotation around
  Z (vertical) relative to geographic north. IMU-derived complement to
  GNSS-based CurrentLocation.Heading.

Rationale:

- AngularVelocity measures rate of rotation (degrees/s) but not absolute
  position. A vehicle upside down in a ditch has zero angular velocity.
  Dispatchers need to know current orientation, not rotation rate.
- Same ISO 8855 axis convention and branch structure as AngularVelocity.
- Useful beyond crash response: off-road, trailer loading, parking assist,
  ADAS tilt compensation.
- Provides raw inputs for Vehicle.Safety.Rollover (PR #894).

Placement:

- Added to spec/Vehicle/Vehicle.vspec after AngularVelocity and before
  RoofLoad, per Erik's guidance to place absolute orientation values
  near existing angular velocity signals.

Companion to PR #894 (Vehicle.Safety branch).

### DIFF
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -332,6 +332,8 @@ Orientation.Roll:
   datatype: float
   type: sensor
   unit: degrees
+  min: -180
+  max: 180
   description: Vehicle rotation (tilt) around X (longitudinal). Positive values = right side down.
   comment: >
     Reference is the vehicle at rest on a level surface.
@@ -342,6 +344,8 @@ Orientation.Pitch:
   datatype: float
   type: sensor
   unit: degrees
+  min: -180
+  max: 180
   description: Vehicle rotation (tilt) around Y (lateral). Positive values = nose up.
   comment: >
     Reference is the vehicle at rest on a level surface.
@@ -352,6 +356,8 @@ Orientation.Yaw:
   datatype: float
   type: sensor
   unit: degrees
+  min: 0
+  max: 360
   description: Vehicle rotation around Z (vertical) relative to geographic north.
   comment: >
     0 = North, 90 = East, 180 = South, 270 = West.
--- a/spec/Vehicle/Vehicle.vspec
+++ b/spec/Vehicle/Vehicle.vspec
@@ -322,6 +322,43 @@ AngularVelocity.Yaw:
   description: Vehicle rotation rate along Z (vertical).
 
 #
+# Spatial Orientation
+#
+Orientation:
+  type: branch
+  description: Spatial orientation. Axis definitions according to ISO 8855.
+
+Orientation.Roll:
+  datatype: float
+  type: sensor
+  unit: degrees
+  description: Vehicle rotation (tilt) around X (longitudinal). Positive values = right side down.
+  comment: >
+    Reference is the vehicle at rest on a level surface.
+    Useful for rollover detection and off-road applications.
+    Related to VSS issue #877 (Connected Safety signals).
+
+Orientation.Pitch:
+  datatype: float
+  type: sensor
+  unit: degrees
+  description: Vehicle rotation (tilt) around Y (lateral). Positive values = nose up.
+  comment: >
+    Reference is the vehicle at rest on a level surface.
+    Useful for incline detection and trailer loading applications.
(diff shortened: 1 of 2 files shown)


erikbosch:
MoM:

- Please review

spec/Vehicle/Vehicle.vspec
```diff
@@ -322,6 +322,43 @@ AngularVelocity.Yaw:
...
+#
+Orientation:
+  type: branch
+  description: Spatial orientation. Axis definitions according to ISO 8855.
+
+Orientation.Roll:
+  datatype: float
+  type: sensor
+  unit: degrees
+  description: Vehicle rotation (tilt) around X (longitudinal). Positive values = right side down.
```
erikbosch, line 337: For clarification we should possibly add min/max for each, -180 to 180 here I assume. I.e. we do not allow representation outside a full circle

spec/Vehicle/Vehicle.vspec
```diff
@@ -322,6 +322,43 @@ AngularVelocity.Yaw:
...
+  comment: >
+    Reference is the vehicle at rest on a level surface.
+    Useful for rollover detection and off-road applications.
+    Related to VSS issue #877 (Connected Safety signals).
+
+Orientation.Pitch:
+  datatype: float
+  type: sensor
+  unit: degrees
+  description: Vehicle rotation (tilt) around Y (lateral). Positive values = nose up.
```
erikbosch, line 349: min/max -180 to 180 here as well?

spec/Vehicle/Vehicle.vspec
```diff
@@ -322,6 +322,43 @@ AngularVelocity.Yaw:
...
+    Useful for incline detection and trailer loading applications.
+    Related to VSS issue #877 (Connected Safety signals).
+
+Orientation.Yaw:
+  datatype: float
+  type: sensor
+  unit: degrees
+  description: Vehicle rotation around Z (vertical) relative to geographic north.
+  comment: >
+    0 = North, 90 = East, 180 = South, 270 = West.
```
erikbosch, line 363: And 0 to 360 here?

spec/Vehicle/Vehicle.vspec
vje013: @erikbosch thanks! Was initially going to put to -90 to 90 for pitch, but I understand why -180 to 180 is stronger here.

spec/Vehicle/Vehicle.vspec
vje013: @erikbosch thank you! Edited.

spec/Vehicle/Vehicle.vspec
vje013: @erikbosch thanks, fixed. I can see this now matches with CurrentLocation.Heading

erikbosch:
MoM:

- Ok to merge after maintainer approval
- S: Double check that axes are well defined according to standard

vje013:
> MoM:
> 
> * Ok to merge after maintainer approval
> * S: Double check that axes are well defined according to standard

@erikbosch @sschleemilch  From what I can tell:

1. the Roll definition is ISO 8855 compliant. 
2. Pitch matches the positive direction but uses a wider range than the standard's -90 to +90 to cover crash scenarios. 
3. Yaw intentionally uses compass bearing convention (matching the existing CurrentLocation.Heading) != ISO 8855's East-referenced yaw, since the primary consumers are dispatch and navigation systems that think in compass bearings.

erikbosch (approved):

Issue #828, opened 2025-06-03
Reuse existing standard vocabulary for units and their conversions


jdacoello:
>
- D: 3 action points
- 1. Check how we can reuse existing standards, including conversion, D can investigate

_Originally posted by @erikbosch in https://github.com/COVESA/vehicle_signal_specification/issues/824#issuecomment-2912755968_

Enhance the VSS modeling approach by importing or reusing established standards to avoid re doing it from scratch.

* Would this better fit the `vss-tools` instead?

jdacoello:
Alternatives:
- https://www.qudt.org  --> Most complete option with conversion information
- https://ucum.org
- http://www.ontology-of-units-of-measure.org

jdacoello:
Example on how we could adapt VSS to reuse existing QUDT units.

> Let's say we want to reuse the kilometre `km` from the [reference standard QUDT](https://qudt.org/vocab/unit/KiloM).


## Option 1: Add identifier of QUDT definition to VSS units file

```yaml
km:
  definition: Length measured in kilometers
  unit: kilometer
  quantity: length
  allowed-datatypes: ['numeric']
  unitRef: http://qudt.org/vocab/unit/KiloM  # <--- This is the pointer to the resource
```
This file can also follow the `YAML-LD` approach to write compact pointers by using prefixes.
```yaml
@context:
  qudtUnit: "http://qudt.org/vocab/unit/"
  unitRef: "@id"
km:
  definition: Length measured in kilometers
  unit: kilometer
  quantity: length
  allowed-datatypes: ['numeric']
  unitRef: "qudtUnit:KiloM"  # <--- This is the pointer to the resource
```

## Option 2: Use the URI of the unit directly
```yaml
km: http://qudt.org/vocab/unit/KiloM  # <--- This is the pointer to the resource
```

PR #900, opened 2026-04-28
feat(units): Add QUDT unit and quantity kind correspondances

jdacoello:
Adds external reference to unit model [QUDT](https://qudt.org/) as explained in #828 

For example: 

the unit kilometer `km` points to its correspondence https://qudt.org/vocab/unit/KiloM which has lots of extra metadata like translated labels, and conversion formulas.

### DIFF
--- a/spec/units.yaml
+++ b/spec/units.yaml
@@ -8,26 +8,36 @@ mm:
   unit: millimeter
   quantity: length
   allowed-datatypes: ['numeric']
+  qudt-unit: http://qudt.org/vocab/unit/MilliM
+  qudt-quantity-kind: https://qudt.org/vocab/quantitykind/Length
 cm:
   definition: Length measured in centimeters
   unit: centimeter
   quantity: length
   allowed-datatypes: ['numeric']
+  qudt-unit: http://qudt.org/vocab/unit/CentiM
+  qudt-quantity-kind: https://qudt.org/vocab/quantitykind/Length
 m:
   definition: Length measured in meters
   unit: meter
   quantity: length
   allowed-datatypes: ['numeric']
+  qudt-unit: http://qudt.org/vocab/unit/M
+  qudt-quantity-kind: https://qudt.org/vocab/quantitykind/Length
 km:
   definition: Length measured in kilometers
   unit: kilometer
   quantity: length
   allowed-datatypes: ['numeric']
+  qudt-unit: http://qudt.org/vocab/unit/KiloM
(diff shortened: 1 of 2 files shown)


spec/units.yaml
```diff
@@ -8,26 +8,36 @@ mm:
   unit: millimeter
   quantity: length
   allowed-datatypes: ['numeric']
+  qudt-unit: http://qudt.org/vocab/unit/MilliM
+  qudt-quantity-kind: https://qudt.org/vocab/quantitykind/Length
```
erikbosch: I like it 

From a tooling perspective it seems to be a bit superfluous to have both `qudt-quantity-kind` and `quantity` here. Maybe make `quantity`optional if there is a qudt reference?

erikbosch:
MoM:

- D: Same unit can be used for multiple quantities
- D: Makes sense to list at least one, most important to have unit reference
- E: Please review/discussion

spec/units.yaml
erikbosch: @sschleemilch  @jdacoello - adding the qudt references can be done without changing vss-tools. But what do you think of changing validation logic so that only one of `quantity` and `qudt-quantity-kind` are required. I.e. `quantity` needed only if `qudt-quantity-kind` is not defined. That way we could limit our file https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/quantities.yaml to only include "custom" quantities.

erikbosch:
@jdacoello - could you update the markdown at https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/rule_set/data_entry/data_units.md#unit-file-syntax so that it mentions the two new fields as optional fields. Possibly also give an example.

spec/units.yaml
sschleemilch: Fine for me. I can have a look.

Btw I would make it a bit more nested:

```yaml
qudt:
  unit: ...
  kind: ...
```

erikbosch:
MoM:

- D to sync with @sschleemilch and update PR

spec/units.yaml
jdacoello: Added indentation as suggested

erikbosch:
MoM:

- Erik to a new look
- D: Do you want to add checks to vSS-tools?
- E: I can create a ticket and see if @sschleemilch have time to add checks
- Ok to merge after maintainer approval

erikbosch (approved):

sschleemilch:
See https://github.com/COVESA/vss-tools/pull/534

PR #903, opened 2026-05-13
Prepare makefile for avro

erikbosch:
This is a follow up for https://github.com/COVESA/vss-tools/pull/520

What we can discuss is release artifacts - if it (as of today) makes sense to include avro as a release artifact (see https://github.com/COVESA/vehicle_signal_specification/blob/master/.github/workflows/buildcheck.yml#L47), if so we may need to think more about arguments like namespace. For now assuming that it does not make sense to add it as a release artifact. Will add a separate ticket to discuss release artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,13 @@
 # Makefile to generate specifications
 #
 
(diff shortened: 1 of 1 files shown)


erikbosch:
MoM:

- Agreed at meeting the avro as of today is not useful as release artifact

PR #906, opened 2026-05-18
feat(safety): add VisibilityImpairment + RoadIcingState derived-assessment signals

aki1770-del:
## Summary

Adds two derived-assessment signals to `Vehicle.Safety.*`:

- `Vehicle.Safety.Whiteout` — assessment of visibility-loss risk in heavy snow / blowing snow / fog conditions
- `Vehicle.Safety.RoadIcingState` — assessment of icing-related loss-of-control risk, distinct from observed surface state

Both signals follow the `Rollover` precedent (#894) in the `Vehicle.Safety` branch: 4-value enum `[UNKNOWN, NONE, RISK, DETECTED]`, derived from sensor-fusion inputs, primary consumers including emergency dispatch and ADAS safety systems.

## Motivation

Extends the `Vehicle.Safety` branch (#894 by @vje013) per the architectural direction articulated by @erikbosch in #877:

> "a `Vehicle.Safety` branch with derived/smart values like `IsFire`, `IsSubmersed` or similar would look nice"

These two signals address driver-life-impact scenarios in snow / icing conditions — vocabulary that allows safety systems and driver-warning systems to act on derived risk rather than only observed surface state.

Both signals are suitable as Semantic Reasoning inputs within the COVESA Central Data Service Playground (CDSP), aligning with the integration direction raised in #877 by @slawr.

## Boundary against existing signals

- **`Vehicle.Exterior.RoadSurfaceCondition`** (#892, merged 2026-05-05) reports observed surface state (DRY/WET/SNOW/ICE/SLUSH/...). `RoadIcingState` reports a derived risk assessment that is bidirectionally distinct: `RoadIcingState.RISK` can be elevated even when `RoadSurfaceCondition` reports `DRY` (black-ice, freezing rain onset), and `RoadIcingState.NONE` can hold even when `RoadSurfaceCondition` reports `SNOW` (low-speed packed-snow with adequate winter-tire grip). The two signals coexist non-redundantly: RoadSurfaceCondition is one input among several to RoadIcingState's risk assessment.
- **`Vehicle.Safety.Rollover`** (#894, merged 2026-05-05) provides the precedent enum shape and comment pattern that both new signals follow.

## VSS issue link

Related to VSS issue #877 (Connected Safety signals).

## Spec changes

- `spec/Safety/Safety.vspec`: +2 signals (Whiteout + RoadIcingState)

## Testing

- `make csv --strict` passes locally (91 VSpecs loaded; new signals appear in output)
- `make json` passes locally
[...]

PR #907, opened 2026-05-20
feat(chassis): add Tire.WinterStatus fitted winter-tire capability attribute

aki1770-del:
## Summary

Adds `Tire.WinterStatus`, a per-wheel attribute reporting the winter (cold-weather / snow) capability class of the fitted tire, under the existing `Vehicle.Chassis.Axle.Row[N].Wheel.{Left,Right}.Tire` branch.

Allowed values: `UNKNOWN` (default), `NON_WINTER`, `WINTER`.

## Motivation

The fitted tire's winter capability is the primary determinant of available traction on snow and ice. Exposing it as a standard signal lets driver-warning and navigation systems adapt snow / mountain-pass advisories to the grip the vehicle actually has, rather than assuming it. Per-wheel placement also surfaces mismatched or partially-fitted winter-tire sets.

This is the second of a two-signal Connected-Safety set related to #877; the first (Vehicle.Safety derived risk signals) is in #906.

## Boundary against existing signals

`Tire.WinterStatus` is fitted-equipment class (an attribute, changes only on tire swap). It is distinct from live-condition sensors such as `Vehicle.Safety.RoadIcingState` (#906) and `Vehicle.ADAS.ESC.RoadFriction`, which report derived/measured live traction. WinterStatus reports installed capability; those report current conditions.

## Spec changes

`spec/Chassis/Wheel.vspec`: one new attribute `Tire.WinterStatus`.

## Testing

`make csv`, `make json`, `make jsonschema`, and `pre-commit` all pass (strict).

---
@erikbosch @vje013 @paulboyes — would appreciate your review when convenient. Companion to #906.

### DIFF
--- a/spec/Chassis/Wheel.vspec
+++ b/spec/Chassis/Wheel.vspec
@@ -74,6 +74,40 @@ Tire.AirTemperature:
   unit: Celsius
   description: Air temperature inside the tire in Celsius.
 
+Tire.WinterStatus:
+  datatype: string
+  type: attribute
+  allowed: [
+    'UNKNOWN',
+    'NON_WINTER',
+    'WINTER'
+  ]
+  default: 'UNKNOWN'
+  description: Winter (cold-weather / snow) capability class of the tire currently fitted to this wheel.
+  comment: >
+    Describes the cold-weather grip capability of the fitted tire, which is the
+    primary determinant of available traction on snow and ice. Modelled as an
+    attribute because it is a property of the mounted equipment that changes only
+    when tires are swapped, not a continuously varying measurement.
+    WINTER indicates a dedicated winter / snow tire (e.g. carrying the
+    three-peak-mountain-snowflake 3PMSF marking), designed for cold-weather rubber
+    compounds and tread patterns that retain grip below approximately 7 degrees
+    Celsius and on snow / ice.
+    NON_WINTER covers summer and all-season tires alike: neither is a dedicated
+    winter tire, and from a snow-traction standpoint both leave the vehicle with
+    materially reduced cold-weather grip. The summer-vs-all-season distinction is
+    a longevity / comfort property that does not change the snow-safety answer and
+    is therefore intentionally not split here.
+    UNKNOWN shall be used when the fitted-tire winter capability cannot be
+    determined (e.g. no tire-identification data available, or input confidence
+    below threshold).
+    Primary consumers: driver-warning systems and navigation systems that adapt
+    snow / mountain-pass routing and advisory thresholds to the grip the vehicle
+    actually has. Combined with per-wheel placement, allows detection of mismatched
+    or partially-fitted winter tire sets.
+    Criteria for each state may be OEM specific.
+    Related to VSS issue #877 (Connected Safety signals).
+
 #
 # Wheel signals
 #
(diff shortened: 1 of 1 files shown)


vje013:
@aki1770-del thanks for the review request

Placement in Chassis/Wheel is correct. This is fitted equipment 
class, not a derived safety assessment. Attribute type is right.

UNKNOWN as default is consistent. Boundary against RoadIcingState 
and RoadFriction is clean.

One question for the group: is WINTER/NON_WINTER sufficient 
granularity? European regulations distinguish alpine-rated (3PMSF) 
from M+S. Studded tires and chains are also distinct capability 
classes. Fine to start simple but worth flagging for reviewer input.

LGTM from an architecture perspective.

spec/Chassis/Wheel.vspec
```diff
@@ -74,6 +74,40 @@ Tire.AirTemperature:
...
+    a longevity / comfort property that does not change the snow-safety answer and
+    is therefore intentionally not split here.
+    UNKNOWN shall be used when the fitted-tire winter capability cannot be
+    determined (e.g. no tire-identification data available, or input confidence
+    below threshold).
+    Primary consumers: driver-warning systems and navigation systems that adapt
+    snow / mountain-pass routing and advisory thresholds to the grip the vehicle
+    actually has. Combined with per-wheel placement, allows detection of mismatched
+    or partially-fitted winter tire sets.
+    Criteria for each state may be OEM specific.
```
erikbosch: Or even market/location specific? 

Like an OEM/Garage could use one criteria in Sweden (to comply with Swedish regulations) and maybe a different criteria in other countries

https://www.transportstyrelsen.se/en/road/vehicles/winter-tyres/

aki1770-del:
Thanks @vje013 — good question, and you're right that 3PMSF, M+S, studded, and chains are all genuinely distinct capability classes, so it's worth being explicit about why I'd start with three values.

The one boundary that's load-bearing for the safety/navigation use case is binary: does the fitted tire have *certified* cold-weather/snow grip or not. That's the 3PMSF line, which is what `WINTER` is anchored to in the comment. M+S is self-declared with no traction test, so for worst-case safety it really sits on the NON_WINTER side rather than as a middle tier — splitting it out risks a consumer reading "M+S = winter-capable" and under-warning the driver. Studded is an ice-specific *modifier* of a winter tire and is regionally/seasonally constrained, and chains are transient add-on equipment that change without a tire swap — which doesn't fit an attribute modelled on swap-cadence.

So my preference is to start with `UNKNOWN / NON_WINTER / WINTER` and keep the safe reading explicit. The string enum extends additively, so if there's reviewer demand we can add a finer class later (e.g. a studded refinement of WINTER) without breaking existing consumers, since `WINTER` stays the certified-grip floor.

Very open to the group's view on this — happy to expand if reviewers feel a 3PMSF/M+S split or a studded class earns its keep.

spec/Chassis/Wheel.vspec
aki1770-del: Good point @erikbosch, and thanks for the Swedish reference — agreed, the criteria are market- and jurisdiction-specific, not only OEM-specific: a garage in Sweden would classify against the national winter-tyre rules while another market applies different criteria. I've widened that line to "OEM-, market-, or jurisdiction-specific (e.g. national winter-tire regulations differ)." The intent is for the signal to report the fitted-tire capability class while leaving the exact classification criteria to the OEM/market — consistent with how VSS leaves criteria implementation-specific elsewhere. Happy to refine the wording further.

erikbosch:
MoM: Please review

erikbosch (approved):
LGTM

erikbosch:
No further comments received, ok to merge

PR #914, opened 2026-05-29
docs(overlay): document the new: intent directive

SoundMatt:
## Summary

Companion documentation for COVESA/vss-tools#527, which introduces the opt-in `new:` attribute for overlay files.

- `new: false` — asserts the node already exists in the base spec (modify intent); errors on typos and upstream removals
- `new: true` — asserts the node does not exist in the base spec (create intent); errors on future naming collisions

New section added between "Node content in Overlays" and "Deleting nodes" in `docs-gen/content/extensions/overlay.md`.

## Test plan

- [ ] Rendered locally with Hugo or reviewed as raw Markdown
- [ ] Links checked — no internal links added, examples are self-contained YAML snippets

### DIFF
--- a/docs-gen/content/extensions/overlay.md
+++ b/docs-gen/content/extensions/overlay.md
@@ -133,6 +133,41 @@ If a glob matches no nodes, a warning is logged and processing continues success
 
 Glob keys can be combined with `delete:` in the usual way.
 
+### Declaring overlay intent with `overlay:`
+
+By default, overlays are permissive: a node that exists in the base spec is modified, and a node that does not exist is created as a new one.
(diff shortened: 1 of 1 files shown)


erikbosch:
@SoundMatt - minor conflict after I merged the other PR, could you resolve

SoundMatt:
Updated to match the naming decision from the maintainer meeting: the attribute has been renamed from `new: true/false` to `overlay: add/edit`.

- `overlay: edit` (was `new: false`) — asserts the node exists; errors if absent
- `overlay: add` (was `new: true`) — asserts the node is new; errors if already present

The companion vss-tools#527 PR has been updated with the same rename.

PR #915, opened 2026-05-29
docs(overlay): document glob wildcard keys

SoundMatt:
## Summary

Companion documentation for COVESA/vss-tools#530, which adds glob wildcard support to overlay files.

- Adds "Wildcard keys" to the feature list at the top of the page
- New section between "Node content in Overlays" and "Deleting nodes" covering:
  - How `*` patterns are expanded against the accumulated spec before merging
  - Multi-segment matching behaviour (each `*` matches across dots)
  - Deterministic expansion order (follows base spec iteration order)
  - Non-fatal no-match behaviour (warning logged, processing continues)
  - Examples for both multi-segment patterns and forward-compatible overlays

## Test plan

- [ ] Reviewed as raw Markdown — no internal links added, examples are self-contained YAML snippets


erikbosch (approved):
LGTM, but new so suggest leaving it open for a while to receive more comments

erikbosch:
MoM: Ok to merge after maintainer approval

PR #917, opened 2026-05-31
Add initial AUTOSAR to VSS mapping reference

nrb2022:
Title: Proposal: AUTOSAR ↔ VSS Signal Mapping Reference

Many automotive organizations use AUTOSAR naming conventions internally while adopting VSS as a common semantic vehicle data model.

I would like to contribute a reference mapping document between commonly used AUTOSAR signals and corresponding VSS signals.

The intent is to help:

* SDV architects
* Middleware developers
* Signal database engineers
* AUTOSAR migration projects

Initial mappings include:

* Vehicle.Speed
* Vehicle.AverageSpeed
* Vehicle.CurrentLocation.Latitude
* Vehicle.CurrentLocation.Longitude
* Vehicle.Powertrain.TractionBattery.StateOfCharge.Current
* Vehicle.MotionManagement.Steering.SteeringWheel.Angle

Before expanding the mapping set, I would appreciate guidance on:

1. Whether such a reference document would be valuable to the project.
2. Preferred location within the repository.
3. Expected format and maintenance approach.

### DIFF
--- a/contributions/autosar_vss_mapping.md
+++ b/contributions/autosar_vss_mapping.md
@@ -1,0 +1,12 @@
+| AUTOSAR Signal | Candidate VSS Signal |
+|----------------|----------------------|
+| VehSpd | Vehicle.Speed |
+| AvgVehSpd | Vehicle.AverageSpeed |
+| Latitude | Vehicle.CurrentLocation.Latitude |
+| Longitude | Vehicle.CurrentLocation.Longitude |
+| Odometer | Vehicle.Odometer |
+| SOC | Vehicle.Powertrain.TractionBattery.StateOfCharge.Current |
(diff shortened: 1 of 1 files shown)


erikbosch:
MoM:

- Presented at meeting by Erik
- Stephen: Covered by AUTOSAR- VSS  mapping, now part of AUTOSAR standard https://www.autosar.org/fileadmin/standards/R24-11/AP/AUTOSAR_AP_TR_VSSRepresentation.pdf
- Ulf: Mapping file should not be in VSS, rather on top
- E: To be further discussed

erikbosch:
MoM:

- Stephen: We can loop in AUTOSAR guys
- E: Can try to reach out
- Ulf: Not a good idea to include it here, should better be defined by AUTOSAR
- E: Agree on that it is not well suited for this repo, could possibly be a separate covesa repo if we have active maintainers

PR #920, opened 2026-06-08
feat: add schemas/vspec-node.schema.json — machine-readable vspec node format

SoundMatt:
## Summary

- Adds `schemas/vspec-node.schema.json` — a [JSON Schema 2020-12](https://json-schema.org/draft/2020-12/schema) document describing every field a vspec YAML node may carry.
- Updates `README.md` with a one-paragraph note pointing to the schema.

## Motivation

The vspec YAML node format is currently documented only by example and by the Python toolchain's parser. Third-party tooling authors must reverse-engineer the field set and silently drop unknown fields when the spec adds new ones. A JSON Schema document gives them a stable, versioned contract to validate against.

The `additionalProperties: false` constraint is the key part — it makes the schema **normative** rather than just descriptive. Any field added to the spec in the future without updating the schema becomes a visible, reviewable diff rather than a silent addition.

## Schema coverage

| Field | Type | Notes |
|---|---|---|
| `type` | enum | `branch`, `sensor`, `actuator`, `attribute` (required) |
| `datatype` | string | required for leaf nodes |
| `description` | string | |
| `unit` | string | should match `spec/units.yaml` |
| `min` / `max` | any | numeric range |
| `default` | any | type must match datatype |
| `comment` | string | informative only |
| `deprecation` | string | non-empty = deprecated |
| `instances` | string or string[] | instance expansion |
| `allowed` | array | enumerated allowed values |
| `pattern` | string | regex constraint on value |

## Test plan

- [ ] `jsonschema` validates each of the sample vspec files against this schema
- [ ] All fields actually used in `spec/` are represented
- [ ] `additionalProperties: false` catches unknown fields

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ You can find current call coordinates and dates in [our wiki](https://github.com
 
 For detailed information see our [contribution guide](CONTRIBUTING.md)!
 
+### Machine-readable node schema
+
+`schemas/vspec-node.schema.json` is a [JSON Schema 2020-12](https://json-schema.org/draft/2020-12/schema) document that describes every field a vspec YAML node may carry. Third-party tooling authors can validate against it so that unknown or misspelled fields are caught at parse time, and any future field additions to the spec are automatically a visible, reviewable diff to the schema. The `additionalProperties: false` constraint is intentional — it makes the schema normative rather than just descriptive.
+
 ### VSS version and release handling
 
 Both VSS (this repository) and [VSS-tools](https://github.com/COVESA/vss-tools) use a [PEP](https://peps.python.org/pep-0440/)
--- a/schemas/vspec-node.schema.json
+++ b/schemas/vspec-node.schema.json
@@ -1,0 +1,61 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://covesa.github.io/vehicle_signal_specification/schemas/vspec-node.schema.json",
+  "title": "VSS vspec node",
+  "description": "A single node in a VSS vspec YAML file.",
+  "type": "object",
+  "required": ["type"],
+  "properties": {
+    "type": {
+      "type": "string",
+      "enum": ["branch", "sensor", "actuator", "attribute"],
+      "description": "Node kind. branch = non-leaf container; sensor/actuator/attribute = leaf."
+    },
+    "datatype": {
+      "type": "string",
+      "description": "Value type for leaf nodes, e.g. float, uint8, string, boolean. Required for sensor/actuator/attribute; not used on branch."
+    },
+    "description": {
+      "type": "string",
+      "description": "Human-readable description of the signal or branch."
+    },
+    "unit": {
+      "type": "string",
+      "description": "Unit of measure. Should be a value from the VSS unit catalogue (spec/units.yaml)."
+    },
+    "min": {
+      "description": "Minimum valid value (numeric). Applies to leaf nodes only."
+    },
+    "max": {
+      "description": "Maximum valid value (numeric). Applies to leaf nodes only."
+    },
+    "default": {
+      "description": "Default value for the signal at startup. Type must match datatype."
+    },
+    "comment": {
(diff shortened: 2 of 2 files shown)


schemas/vspec-node.schema.json
```diff
@@ -0,0 +1,61 @@
...
+    "description": {
+      "type": "string",
+      "description": "Human-readable description of the signal or branch."
+    },
+    "unit": {
+      "type": "string",
+      "description": "Unit of measure. Should be a value from the VSS unit catalogue (spec/units.yaml)."
+    },
+    "min": {
+      "description": "Minimum valid value (numeric). Applies to leaf nodes only."
```
sschleemilch, line 27: type `number`?

schemas/vspec-node.schema.json
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "max": {
+      "description": "Maximum valid value (numeric). Applies to leaf nodes only."
```
sschleemilch, line 30: number type

schemas/vspec-node.schema.json
```diff
@@ -0,0 +1,61 @@
...
+    "instances": {
+      "description": "Instance expansion expression (string or array of strings) for parameterised branches.",
+      "oneOf": [
+        { "type": "string" },
+        { "type": "array", "items": { "type": "string" }, "minItems": 1 }
+      ]
+    },
+    "allowed": {
+      "type": "array",
+      "description": "Enumerated allowed values for the signal. Items should match the datatype.",
```
sschleemilch, line 52:
```suggestion
      "description": "Enumerated allowed values for the signal. Items must match the datatype.",
```

schemas/vspec-node.schema.json
```diff
@@ -0,0 +1,61 @@
...
+      "minItems": 1
+    },
+    "pattern": {
+      "type": "string",
+      "description": "Regular expression that the signal value must match."
+    }
```
sschleemilch, line 58: `enum` is missing

sschleemilch (commented):
I believe `vss-tools` could (and should) generate it since we are using pydantic and is then the single source of truth aligned.

Basically `VSSData.model_json_schema()` and `VSSDataDatatype.model_json_schema()`.
Sure, it also contains internal things and probably needs to be tuned a bit

erikbosch:
Thanks for the PR.

README.md
```diff
@@ -29,6 +29,10 @@ You can find current call coordinates and dates in [our wiki](https://github.com
 
 For detailed information see our [contribution guide](CONTRIBUTING.md)!
 
+### Machine-readable node schema
+
+`schemas/vspec-node.schema.json` is a [JSON Schema 2020-12](https://json-schema.org/draft/2020-12/schema) document that describes every field a vspec YAML node may carry. Third-party tooling authors can validate against it so that unknown or misspelled fields are caught at parse time, and any future field additions to the spec are automatically a visible, reviewable diff to the schema. The `additionalProperties: false` constraint is intentional — it makes the schema normative rather than just descriptive.
```
erikbosch, line 34: I do not know how easy it would be to create a minimal example on how this could be used, that we potentially could execute as part of CI to verify that std catalog and the schema are in sync.  I am thinking of something

- `vspec export yaml` or `vspec export json`
- And then some command that checks that it complies with schema
Have you done any checks like that to verify that the schema works with the current std catalog?

erikbosch:
MoM:

- Presented at meeting, please review

erikbosch:
> I believe `vss-tools` could (and should) generate it since we are using pydantic and is then the single source of truth aligned.
> 
> Basically `VSSData.model_json_schema()` and `VSSDataDatatype.model_json_schema()`. Sure, it also contains internal things and probably needs to be tuned a bit

Any thoughts @sschleemilch on how easy/difficult it would be to have it generated? One can discuss if this should be a vss-tools artifact or placed here in this repo. If/when generated vss-tools might be a better place. If generated one could also just generate it on the fly like we do with other release artifacts.

PR #929, opened 2026-07-27
feat: documenting `includes` vss-tools feature

sschleemilch:
Companion PR for vss-tools change: https://github.com/COVESA/vss-tools/pull/546


erikbosch:
Thanks!

Things to discuss in the VSS meeting

- Check that no-one objects the new syntax
- Agree on preferred variant for std catalog
  - Possibly request new syntax for new additions, and by time replace old with new so we can deprecate old syntax, and possibly even give error in strict mode for old syntax
- Agree on long term plan (Support both or remove support for old construct)

erikbosch:
MoM:


-     Ticket presented
-     Schildt: We shall better stick to one of the options in the std catalog. Makes life easier if someone wants to implement their own tooling

erikbosch:
@sschleemilch - some conflicts reported

Proposal for meeting August 25th:

- Lets merge this one and https://github.com/COVESA/vss-tools/pull/546, include them in upcoming 6.1
- For now stick to old syntax in std catalog
- We may switch to new syntax if someone does the work to convert so we only get new syntax, but that would be a separate PR

PR #931, opened 2026-07-28
Add air quality signals for cabin and exterior air

### LINKED ISSUE #930: Air quality: VSS has no cabin or ambient air composition signals, and no concentration units
VSS currently has no way to express air quality. I went looking for a signal to publish a cabin particulate reading to and could not find one, so I checked properly across master, v6.0 and v5.1.

There are no air quality signals anywhere in the catalog. No particulate matter, no CO2, no VOC, no NO2, no ozone. The only CO2 in the spec is `Vehicle.EmissionsCO2`, which is a tailpipe g/km attribute, plus the diesel particulate filter temperatures under Powertrain. Neither describes the air in or around the cabin.

The gap is structural as well as missing leaves: `units.yaml` has no concentration units at all. No ppm, no ppb, no ug/m^3, and `quantities.yaml` has no mass-density quantity. So a proposal here cannot only add signals, it has to touch the shared unit infrastructure too.

I could not find any open issue or PR proposing this.

Why it seems worth having:

- Cabin particulate sensors ship in production cars today, and several OEMs display a PM2.5 reading or drive an air purifier from it. The hardware exists, the signal standard does not, so every implementation is proprietary.
- Automatic recirculation is the clearest use case, and it needs an outside reading as well as an inside one, which is why I think this belongs on both Cabin and Exterior.
- Cabin filter monitoring falls out of the same signal.

I have a change prepared: seven measurements (PM1, PM2.5, PM10, CO2, NO2, Ozone, TVOC) on both `Cabin.AirQuality` and `Exterior.AirQuality` through a shared include, plus the three units and the one quantity that are missing. It validates with vss-tools from master in strict mode and adds 14 leaves, and I have it running on a QNX signal service against a real I2C sensor, so the shape has been through something other than review.

Happy to open the PR, or to rework the scope first if the branch placement or the unit choices are wrong. I would rather find that out here than in a PR.

ethanrossauto:
Closes #930

Adds seven air quality measurements to both the cabin and the exterior, plus the concentration units the catalog is missing.

```
Vehicle.Cabin.AirQuality.{PM1, PM25, PM10, CO2, NO2, Ozone, TVOC}
Vehicle.Exterior.AirQuality.{PM1, PM25, PM10, CO2, NO2, Ozone, TVOC}
units:      ppm, ppb, ug/m^3        quantities: mass-per-volume
```

All `type: sensor`, `datatype: float`, `min: 0`, no max. Defined once in `spec/include/AirComposition.vspec` and included from both branches with a prefix, the same way `SingleHVACStation.vspec` and `include/MovableItem.vspec` are reused today, so the two branches cannot drift apart.

**Validation.** `vspec export json --strict` passes on the patched tree using vss-tools installed from git master, which is what CI uses. Leaf count goes from 1367 to 1381, which is the 14 new signals and nothing else. Released vss-tools 6.0.0 from PyPI fails on unmodified master for unrelated reasons, so master is the version to test with.

**Some decisions worth flagging, since I expect discussion on them:**

`PM25` rather than `PM2.5` or `PM2_5`, because no signal name in the catalog currently contains a dot or an underscore. The description spells out the 2.5 micrometer cutoff.

Not under `Cabin.HVAC`, because HVAC in VSS models the climate actuation system. Air composition is a property of the cabin that exists whether or not HVAC acts on it.

New units are the widest part of this change. `ppm` and `ppb` go under the existing `relation` quantity alongside `percent`. `ug/m^3` needs a new `mass-per-volume` quantity, named after the existing `mass-per-time` and `mass-per-distance`. QUDT references included, all resolvable.

No AQI. Air quality indices are jurisdiction specific and derivable from the raw values, so they seemed like the wrong thing to standardize.

CO2 is the one I am least sure about. Many low cost parts report estimated eCO2 derived from a VOC channel rather than true NDIR CO2, and the two behave differently, so the signal carries a comment saying so. If reviewers would rather split it into `CO2` and `CO2Estimated`, that seems reasonable to me.

TVOC is in ppb, which is what sensors emit, though the reference methods report total VOC as a mass concentration. If the standards aligned unit is preferred, switching it to `ug/m^3` is a one line change since that unit is already in this PR.

The exterior branch is the weaker half of this proposal and I would rather say so than have it found. Exterior CO2 in particular sits near 420 ppm everywhere and has no obvious use. It is there because the shared include keeps the branches identical by construction. If you would rather trim exterior to the signals with a real use case, that is a small change and I have no objection.

**Where this came from.** I was bringing an air quality sensor into QNX and wanted to publish it as VSS signals on the QNX signal service, which is COVESA VSS based. There was nothing to publish to. The implementation runs on real hardware and is public, if it is useful context: https://github.com/ethanrossauto/qnx-air-quality-sensor

### DIFF
--- a/spec/Cabin/Cabin.vspec
+++ b/spec/Cabin/Cabin.vspec
@@ -30,6 +30,14 @@ HVAC:
 #include HVAC.vspec HVAC
 
 ##
+# Air quality signals
+##
+AirQuality:
+  type: branch
+  description: Signals describing the composition of the air inside the vehicle cabin.
+#include ../include/AirComposition.vspec AirQuality
+
+##
 # Infotainment
 ##
 Infotainment:
--- a/spec/Vehicle/Exterior.vspec
+++ b/spec/Vehicle/Exterior.vspec
@@ -223,3 +223,8 @@ WindDirection:
     (0 = from North, 90 = from East, 180 = from South, 270 = from West).
     Only meaningful when WindSpeed is above a minimum threshold; value when
     WindSpeed is near 0 is implementation specific.
+
+AirQuality:
+  type: branch
+  description: Signals describing the composition of the ambient air outside the vehicle.
+#include ../include/AirComposition.vspec AirQuality
--- a/spec/include/AirComposition.vspec
+++ b/spec/include/AirComposition.vspec
@@ -1,0 +1,76 @@
+# Copyright (c) 2026 Contributors to COVESA
+#
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+#
+# Air composition (air quality) signals.
+#
+# This file is intended to be included with a branch prefix so that the same
+# set of measurements can be reported for cabin (interior) air and for
+# ambient (exterior) air, e.g.:
+#
+#   #include ../include/AirComposition.vspec AirQuality
+#
+# All signals are concentrations of pollutants relevant to occupant health
+# and to HVAC automation (recirculation control, cabin filter monitoring,
+# air purifier control).
+#
+
+PM1:
+  datatype: float
+  type: sensor
+  unit: ug/m^3
+  min: 0
+  description: Mass concentration of particulate matter with aerodynamic diameter
+               of 1 micrometer or less (PM1.0).
+
+PM25:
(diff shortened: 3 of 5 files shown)


erikbosch:
Thanks for the PR. Took a quick look and see no major problems. We will follow the regular VSS process - present the PR at the next weekly VSS meeting, and then wait a week or more to see if anyone has any objections. Due to vacation period in Europe things goes a bit slower than usual.

sschleemilch (approved):

PR #928, opened 2026-07-20
FMS 05 11.11.2025 signals added

sergejsaibel:
All the CAN FMS 05 11.11.2025 signals are added.

PR #932, opened 2026-08-03
Acea fms05_11_11_2025 signals

sergejsaibel:
This pull request comes from the ACEA onboard API working group and adds signals from the ACEA "FMS-Standard description
Version 05 11.11.2025" specification document that can be found at https://www.fms-standard.com/ .
These new-added signals together with some already existing VSS catalog signals fully cover the ACEA FMS interface version 05 11.11.2025. However, to not create any compatibility issues, some signal definitions are different from the ACEA ones that can be found in the pull request https://github.com/COVESA/vehicle_signal_specification/pull/928 for the [acea-namespace](https://github.com/COVESA/vehicle_signal_specification/tree/acea-namespace) branch.
Please see the differences in the file attached, where ACEA VSS tree is a source and the COVESA VSS tree is a target.
[ComparisonAceaSource_CovesaTarget.txt](https://github.com/user-attachments/files/30656127/ComparisonAceaSource_CovesaTarget.txt)


tguild:
We need to resolve the instances discussion as part of this PR for axels, @SebastianSchildt has a good suggestion that I believe aligns with JASPAR. ACEA shares instances handling interest and following JASPAR suggestions.

Should we factor in concept of potentially multiple fuel tanks in VSS or have that only be in ACEA namespace branch?

I defer to others on bus variant files and how they impact vss-tools or if they should be in ACEA namespace.

Door should be under Cabin not Body.

TractionBattery.ChargingState perhaps should be TractionBattery.Charging.ChargingState or TractionBattery.Charging.State, the description says "beside status 0 and 15" but the enum only defines values 0–3 and 10–13

CruiseControl.IsSwitchedOn is semantically similar to CruiseControl.IsEnabled, how does it differ?

Tachograph VehicleSpeed seems redundant to our other concepts of Speed and should be discussed more.

Retarder 125% should be explained in comment, it exists in J1939 as an error.

Bellows instead of Bellow?

sergejsaibel:
> We need to resolve the instances discussion as part of this PR for axels, @SebastianSchildt has a good suggestion that I believe aligns with JASPAR. ACEA shares instances handling interest and following JASPAR suggestions.

I believe it would be good to ask @SebastianSchildt to present his suggestion on one of the ACEA Onboard API meetings.

sergejsaibel:
> Door should be under Cabin not Body.

Let's discuss that, because the Swedish "Kaross" translates to "Body". A "Cabin" usually used talking about trucks.

sergejsaibel:
> CruiseControl.IsSwitchedOn is semantically similar to CruiseControl.IsEnabled, how does it differ?

We were not sure if these two are similar during our discussions in ACEA. IsSwitchedOn means that the physical button or other kind of control is set into ON position. Is it the same as IsEnabled?

tguild:
> > CruiseControl.IsSwitchedOn is semantically similar to CruiseControl.IsEnabled, how does it differ?
> 
> We were not sure if these two are similar during our discussions in ACEA. IsSwitchedOn means that the physical button or other kind of control is set into ON position. Is it the same as IsEnabled?

That is my understanding.

spec/Chassis/Chassis.vspec
```diff
@@ -124,12 +131,18 @@ Axle.Torque:
 #
 # Wheels on Axles
 #
-Axle.Wheel:
-  instances: ["Left","Right"]
+#ATTENTION! The number of wheels on each axle below is just an example and should be adjusted to the actual vehicle configuration!
+Axle.Row1.Wheel:
+  instances: ["Pos7", "Pos9"]
```
erikbosch, line 136: MoM:

- Sergej: Left and Right not enough for commercial vehicles. There is a "standard naming convention" for commercial vehicles. This example is 2 wheels on front axle and 4 wheels on rear axle.
- Sergej: ok to not have exactly this std configuration on master branch
- Ulf: Expansion shall be part as post processing
- Ulf: CVII has a complete solution to it. Pre-processor do some work before VSS tool. Use overlay mechanism. Replaces instance information.
- Erik: So this maybe goes in the direction of having a separate instance file.
- S: COnfiguration is OEM specific. But good to have a default value
- Ulf: Default values like today is good, as of today we have for passenger cars
- E: We could potentially have a "default truck profile" like https://github.com/COVESA/vehicle_signal_specification/blob/master/overlays/profiles/motorbike.vspec 
- E: Can show an example based on this PR.
- S: Shall we use this 15 wheel standard?
- E: Is there a (public) standard that use the 15 wheel std. Maybe @abdelkadersellami  knows. Pos8 is centre wheel?

abdelkadersellami: Both ISO 11992-2 and SAE J1939-71 uses one byte parameter to encode Axle/Wheel identification: 
Screenshot 2026-08-19 085426

In ISO standard:
- Axles are numbered 1, 2, 3... from the front of the direction of the travel.
- Wheel numbering starts at the vehicle centerline.
- Centre wheel: 8.
- Left side: 7 nearest the centreline, then 6, 5 outward.
- Right side: 9 nearest the centreline, then 10, 11 outward.
image

In SAE standard:
- Axles are numbered 1, 2, 3... 15 from the front to the back while facing forward.
- Tyres on each axle are counted continuously across the vehicle, including dual tyre (1..15, left to right while facing forward)

| Physical position| ISO 11992-2|SAE J1939|
| -------- | ------- | ------- |
| Axle 1, left tyre| 0x17|0x00|
| Axle 1, right tyre| 0x19|0x01|
|Axle 2, left outer|	0x26|	0x10|
|Axle 2, left inner|	0x27|	0x11|
|Axle 2, right inner|	0x29|	0x12|
|Axle 2, right outer|	0x2A|	0x13|

spec/Chassis/Chassis.vspec
```diff
@@ -85,6 +85,13 @@ Axle.SteeringAngle:
   description: Single track two-axle model steering angle.
                Angle according to ISO 8855. Positive = degrees to the left. Negative = degrees to the right.
   comment: Single track two-axle model steering angle refers to the angle that a centrally mounted wheel would have.
+
+Axle.Weight:
+  datatype: uint16
+  type: sensor
+  unit: kg
+  description: The weight of the axle in kg.
```
erikbosch, line 93: I am not a truck guy, but do we need to clarify with a comment what we mean here. I first thought we meant the weight of the axle as a spare part. But this weight as if you would put a scale under the axle. I.e. for a lifted axle (not touching the ground) the value shall be 0, right?

Possibly something like `comment: Weight bearing on the roadway for all wheels connected to the given axle`. Taken from https://en.wikipedia.org/wiki/Axle_load

erikbosch:
I see two types of changes in this PR. One is changes that makes sense also in std catalog, and can be added without affecting backward compatibility. The other concerns adaptions for commercial vehicles that we likely do not want in std catalog, at least not as long as our std setup in std catalog is a passenger car. In #940 I present some ideas on how we can put acea adaptions in an overlay. If we go that way we can consider having multiple acea profiles, if we want to differentiate between with/without bus door signals.

spec/ADAS/ADAS.vspec
```diff
@@ -103,6 +103,26 @@ CruiseControl.AdaptiveIntervalSet:
   type: actuator
   description: Follow distance setting, commonly 1-5 with 1 being closest.
 
+CruiseControl.IsSwitchedOn:
```
erikbosch, line 106: We have a similar signal in `CruiseControl.IsActive` and I think we need to clarify the differences, if any. If we want same logic for both signals we do not need both

Signal | IsActive | IsSwitchedOn
---------|----------|-------------
"Normal" cruise control active | True | True
Driver overrides by pressing accelerator pedal | ? | True
Driver overrides by pressing brake pedal | ? | True

spec/ADAS/ADAS.vspec
```diff
@@ -103,6 +103,26 @@ CruiseControl.AdaptiveIntervalSet:
...
+  type: sensor
+  enum:
+    'OFF_DISABLED': 0
+    'HOLD': 1
+    'ACCELERATE': 2
+    'DECELERATE': 3
+    'RESUME': 4
+    'SET': 5
+    'ACCEL_OVERRIDE': 6
+  description: State of the cruise control system.
```
erikbosch, line 124: Do we have a description of the states? Possibly add as comment.

spec/Body/Accessibility.vspec
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2016 Contributors to COVESA
```
erikbosch, line 1: Use 2026 for new files

spec/Body/DoorsBusOnly.vspec
```diff
@@ -0,0 +1,54 @@
...
+  datatype: uint8
+  enum:
+    'AT_LEAST_ONE_DOOR_OPEN': 0
+    'CLOSING_LAST_DOOR': 1
+    'ALL_DOORS_CLOSED': 2
+  type: sensor
+  description: Signal indicating the actual position of the doors.
+
+# Per door signals, instantiated for each door
+Doors.IsLocked:
```
erikbosch, line 41: Does this imply that the door is closed as well or should you better check IsOpen as well?

spec/Cabin/Cabin.vspec
```diff
@@ -197,3 +189,24 @@ Convertible.Status:
...
 #include include/PowerOptimize.vspec
+
+##
+# Driver compartment signals
+##
+DriverCompartment:
+  type: branch
+  description: Driver compartment signals
+
+DriverCompartment.InteriorTemperature:
```
erikbosch, line 200: This quite similar to `Vehicle.Cabin.HVAC.AmbientAirTemperature`. Why are both needed?

spec/Cabin/Tachograph.vspec
```diff
@@ -0,0 +1,88 @@
...
+
+#
+# All Tachograph signals
+#
+
+#---Drivers branch
+Drivers:
+  type: branch
+  instances:
+    - Driver[1,2]
```
erikbosch, line 17: Add a comment on why two drivers are specified. Like refer to the standard that supports two drivers. I also assume the intention is to support scenarios where two drivers alternate, not American fire trucks with two drivers at the same time, one at the front and one at the end

spec/Cabin/Tachograph.vspec
```diff
@@ -0,0 +1,88 @@
...
+  description: Driver signals
+
+Drivers.WorkingState:
+  datatype: uint8
+  enum:
+    'REST': 0
+    'DRIVER_AVAILABLE': 1
+    'WORK': 2
```
erikbosch, line 25: Add a comment on what the states mean.

spec/Cabin/Tachograph.vspec
```diff
@@ -0,0 +1,88 @@
...
+Drivers.TimeRelatedState:
+  datatype: uint8
+  enum:
+    'NORMAL': 0
+    'STATE_15_MIN_BEF_4_5_H': 1
+    'STATE_4_5_H_REACHED': 2
+    'STATE_15_MIN_BEF_9_H': 3
+    'STATE_9_H_REACHED': 4
+    'STATE_15_MIN_BEF_16_H': 5
+    'STATE_16_H_REACHED': 6
```
erikbosch, line 39: Add a comment with some context. As I understand the  4.5 hours and 9 hour limits are EU-specific and the 16 hour limit US-specific.

But they refer to different measurements, like a driver can theoretically at the same time exceed the 4.5 h continuous driving limit and the 9 hour total limit. Is it then undefined which state to show or are there some order to be used?

spec/Cabin/Tachograph.vspec
```diff
@@ -0,0 +1,88 @@
...
+IsOverspeed:
+  datatype: boolean
+  type: sensor
+  description: Indicates whether the vehicle is exceeding the legal speed limit set in the tachograph.
+
+VehicleSpeed:
+  datatype: float
+  type: sensor
+  unit: km/h
+  description: Speed of the vehicle registered by the tachograph.
```
erikbosch, line 62: Are negative values supported, e.g. when reversing

spec/Chassis/Retarder.vspec
```diff
@@ -0,0 +1,32 @@
...
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+RetarderTorqueMode:
+  datatype: uint8
+  type: sensor
+  min: 0
+  max: 15
```
erikbosch, line 13: VSS min/max is inclusive. Can value 15 be used, not mentioned in description

erikbosch: Are definition/interpretation and number of modes available OEM specific? I.e. Torque Mode 7 may mean something different for different brands?

spec/Chassis/Suspension.vspec
```diff
@@ -0,0 +1,25 @@
...
+
+#--- Chassis.Suspension branch
+Suspension:
+  type: branch
+  description: Vehicle suspension system data.
+
+Suspension.Axles:
+  type: branch
+  instances:
+    - ["Front","Rear"]
```
erikbosch, line 17: Would it be an idea to put this in `Vehicle.Chassis.Axle.Suspension` to take advantage of the same instantiation concerning number of axles.

spec/Powertrain/AlternatorBusOnly.vspec
```diff
@@ -0,0 +1,22 @@
...
+
+#
+# Bus alternator signals
+#
+
+Alternators:
+  type: branch
+  instances:
+    - Alternator[1,4]
+  description: Bus alternators.
```
erikbosch, line 17: We do not have any signal for 12/24V alternator as of today. Would we like to have it also for passenger vehicles, but leave the instantiation as a acea bus extension?

spec/Powertrain/FuelSystem.vspec
```diff
@@ -27,6 +27,46 @@ SupportedFuel:
                with additional suffix for octane (RON) where relevant.
   comment: RON 95 is sometimes referred to as Super, RON 98 as Super Plus.
 
+# SHOULD be synced with the supported fuel types above!!!
```
erikbosch, line 30: We should better have a description of the items - or alternatively reference a document where they are defined.

Like when shall NATURAL_GAS be used instead of LNG/CNG?`, and what is the difference between BI_GAS and HYB_GAS?

Needs to be discussed if/how this can be mapped to SupportedFuelTypes. There are some differences as this signal includes combinations and electricity, while the other only concern combustion fuels. Differences can also be discussed on detailed levels, like is E85 and ETH the same thing.

spec/Powertrain/FuelSystem.vspec
```diff
@@ -54,6 +94,20 @@ RelativeLevel:
   max: 100
   description: Level in fuel tank as percent of capacity. 0 = empty. 100 = full.
 
+FuelTypes:
+  type: branch
+  instances:
+    - Fuel[1,2]
+  description: Vehicle fuel type data.
+
+FuelTypes.RelativeLevel:
```
erikbosch, line 103: Or is it the whole (or parts)  FuelSystem branch that better shall be instantiated by adding a tank branch. Like having separate instances for TankCapacity

spec/Powertrain/TractionBattery.vspec
```diff
@@ -157,6 +157,19 @@ StateOfCharge.CurrentEnergy:
...
+  type: sensor
+  enum:
+    'IDLE': 0
+    'CHARGING': 1
+    'STANDBY': 2
+    'INITIALISING': 10
+    'PRE_CHARGE': 11
+    'TERMINATING': 12
+    'BATTERY_FAILURE': 13
+  description: Overall State of the DC charging system. The information is reported by the EVSE (Electric Vehicle Supply Equipment) beside status 0 and 15.
```
erikbosch, line 171: 15 not mentioned in enum

erikbosch: We should possibly have a comment indicating what the states mean, or a reference to some document/standard defining that. When is for example STANDBY used?

spec/Vehicle/HMI.vspec
```diff
@@ -0,0 +1,20 @@
...
+
+#
+# In-Vehicle HMI Signals
+#
+
+#---TellTales branch
+TellTales:
+  type: branch
+  instances:
+    - TellTale[1,75]
```
erikbosch, line 17: Add a comment on why 75 is selected. Likely because of some standard I suppose.

Are the indexes well defined, i.e. that it is standardised what number 33 indicates? If not, do we need some extra field in TellTales.vspec to include some (free text) identifier or is it supposed that both sender/receiver knows the mapping for this particular vehicle.

spec/Vehicle/Vehicle.vspec
```diff
@@ -582,4 +588,10 @@ CurrentLocation.GNSSReceiver.MountingPosition.Z:
                Axis definitions according to ISO 8855. Origin at center of (first) rear axle.
                Positive values = above center of rear axle. Negative values = below center of rear axle.
 
+CurrentDateTime:
+  datatype: string
+  type: sensor
+  unit: iso8601
+  description: Current date and time provided by the vehicle's internal clock.
```
erikbosch, line 595: UTC timezone, or arbitrary?

spec/Chassis/Chassis.vspec
sergejsaibel: My proposal is to use the SAE wheel numbering for all types of vehicles. If it's ok, we can define a standard configuration on master branch having 2 axles with 2 wheels on each. Other cases will be solved when we have vehicle variant handling on place.

spec/Chassis/Chassis.vspec
sergejsaibel: Good point, we can use the J1939 definition: "Total mass imposed by the tires on the road surface at the specified axle." Knowing that we at ACEA for now are using the comments to reference to J1939, I prefer to update the definition combining these two sentences in one: "Axle weight in kg: refers to the total mass imposed by the tires on the road surface at the specified axle."

spec/ADAS/ADAS.vspec
sergejsaibel: We have a similar signal in `CruiseControl.IsActive` and I think we need to clarify the differences, if any. If we want same logic for both signals we do not need both

Signal | IsActive | IsSwitchedOn | IsEnabled
---------|----------|-------------|-------------
"Normal" cruise control active | True | True | True
Driver overrides by pressing accelerator pedal | False | True| ?
Driver overrides by pressing brake pedal | False | True |  ?

As I understand it, both are False. However, there is IsEnabled too. What does that mean?

spec/ADAS/ADAS.vspec
sergejsaibel: There is nothing more specific in the J1939 standard; the SAE most likely considers the states to be self-explanatory.

spec/Body/Accessibility.vspec
sergejsaibel: Fixed!

spec/ADAS/ADAS.vspec
erikbosch: Below is the definition we have as a comment as of today.

In many vehicles there are two steps required to activate cruise control. I would assume then that after the first "button press" (or similar) it enters enable state, on second press it gets active and starts controlling accelerator and possibly brakes.

But this is just my interpretation, and especially for the driver override scenario and what happens afterwards it is quite undefined. And does not implementation vary as well - in some vehicles cruise control is deactivated if driver presses the accelerator/brake, in others it is just temporarily paused until the pedal is released, or?

```
#
# For systems that only takes action under specific circumstances (like ABS) two signals are often used:
# IsEnabled indicates that the system has been turned on, is monitoring and is ready to take action
# IsEngaged indicates that the system is currently taking action (e.g. regulating brake pressure)
#
# For systems that are taking action over an longer period of time (like Cruise Control) two signals are often used:
# IsEnabled indicates that the system is enabled and available for e.g. configuration, but it is not yet taking action.
# IsActive indicates that the system has been turned on and is continuously taking action,
# (e.g. regulating speed and brakes)
#
```

spec/ADAS/ADAS.vspec
erikbosch: I found some description in a table in the J1939DA support info document. But as you say most likely partially self-explanatory or up to OEM implementation.

spec/ADAS/ADAS.vspec
sergejsaibel: It feels like IsSwitchedOn is the same as IsEnabled. @abdelkadersellami , what do you think?

spec/ADAS/ADAS.vspec
sergejsaibel: It's interesting that "RESUME" often leads to "ACCELERATE", but the state can only have one value at a time. Will "RESUME" stay for just some milliseconds in this case? How many milliseconds? How detailed should we describe cases like this? It's an important question, otherwise the state will be implemented differently by the brands. Is there a COVESA guideline about that?

spec/Body/DoorsBusOnly.vspec
sergejsaibel: The FMS spec says: 
Lock Status: 
locked -> doors cannot be operated by the driver or a passenger 
unlocked -> door may be operated by the driver or a passenger 
Open Status: 
closed -> door is completely closed 
open -> door is not completely closed
As I understand that, a door can be locked in open position.

spec/Cabin/Cabin.vspec
sergejsaibel: The InteriorTemperature is more specific, it's where the driver is sitting while the AmbientAirTemperature is "Ambient air temperature inside the vehicle." For a bus, the difference can be quite big.

spec/Cabin/Tachograph.vspec
sergejsaibel: :) No, this is about the tachograph SAE J1939 TCO1 CAN message, and the message itself supports 2 drivers. Is it ok if I replace "All Tachograph signals" by "Tachograph signals according SAE J1939 TCO1 CAN message"?

spec/Cabin/Tachograph.vspec
sergejsaibel: J1939 says:
000b = Rest - sleeping
001b = Driver available – short break
010b = Work – loading, unloading, working in an office
011b = Drive – behind wheel 
100b = Reserved
101b = Reserved
110b = Error
111b = Not available
I'm not quite happy with the "Rest - sleeping". It's very difficult to know if the driver is really sleeping, how to implement that?

spec/Cabin/Tachograph.vspec
sergejsaibel: If the driver has reached the 4.5 hours limit, it should take a 45 minutes break. Then the driver is allowed to drive 4.5 more hours. When the driver has reached 9 hours limit, it's not allowed to drive more that day. So it starts with Normal and continue showing all the limits one by one. Should I describe the whole sequence of the states or just comment what each state means?

spec/Cabin/Tachograph.vspec
sergejsaibel: No. There is a DirectionIndicator for that.

spec/Chassis/Retarder.vspec
sergejsaibel: Good point, max should be 14 because 15 is "Not available" and will be returned by the comm. protocol. Otherwise, the full description is: 

Retarder Torque Mode: State signal which indicates which retarder torque mode is currently generating, limiting, or controlling the torque. Note that the modes are not in prioritized order. Not all modes may be relevant for a given device. Some devices may not implement all functions. Mode 0000b means “No request”: retarder torque = 0 (no braking). See Appendix D in SAE documentation Modes 0001b to 1110b indicate that there is either a torque request or the identified function is currently controlling the retarder: retarder torque may range from 0 (no braking) to the upper limit.

As I understand that, it can be OEM specific.

spec/ADAS/ADAS.vspec
erikbosch: For newer or modified signals we try when possible to state (a possible) logic, or state that details are OEM specific. I found some state info in J1939DA_202607 Support Information.pdf Jul 2026 A.7. But as you say details seems to be up to implementation. Like if ACCELERATE only shall be active when drivers wants to increase both captured and actual speed "one step", or also after resume where the captured/stored speed isn't changed.

Here we could possibly add something like. " States according to J1979, exact logic is OEM dependent"

spec/Cabin/Tachograph.vspec
erikbosch: Maybe based on some work time laws somewhere that differentiate between shorter breaks when you are on standby and longer breaks.

We could possibly just add J1939 to the description or as a comment. Might be good to have some traceability for use-cases where we want to have same values as in J1939

spec/Cabin/Tachograph.vspec
erikbosch: Nah, think we can keep it as is.

spec/Chassis/Chassis.vspec
erikbosch: I created https://github.com/COVESA/vehicle_signal_specification/pull/940 to show a possible ACEA/FMS profile. We need to agree on how to proceed - change default tree instances for all or just as a profile

spec/Powertrain/FuelSystem.vspec
erikbosch: We had an old discussion which ended up in a placeholder issue at https://github.com/COVESA/vehicle_signal_specification/issues/875. we need to discuss in general if we need to refactor for multiple fuel systems and/or multiple fuel tanks

erikbosch:
MoM:

- Erik: I see 3 discussion areas
 -  Instances for wheels/axles
 -  Fuelsystem and tank instantiation
 - Individual signals
- Sergej: Is it possible to deprecate old way, to get common vocubalary. Helps us end up with better naming.
- E: Doable -if we go that way there are two aspects to discuss (or two alternatives)
 - Keep default structure of 2 axles,  2 wheels each , just rename left and right 
 - Or also add twin wheels for rear
- S: That would be enough
- E: I see the "new" proposed wheel naming as an improvement, but would be a very visible change so needs to be well socialized
- E: Compare with Jaspar Yorito - to be discussed next week with Jaspar
- T: Supportive to change instances
- T: Like to see descriptive items. Would be good to know where it would be. Hoped Daniel would have a good description
- E: Would not be a problem to have a wheel attribute with "user friendly name", same for other instances
- E: Jaspar have thought about numeric instances for electric motors
- T: Could be by enum attributes left/right, driverside/passengerside

**Please let us know if you have any objections to replacing Left/Right with Pos7/Pos9 in COVESA standard Catalog (and change Center to Pos8 in motorbike profile). VSS std catalog should still have 2 wheels for rear axle as default.**

- S: Vehicles may have multiple tanks, often they have a common sensor but they may have different sensors
- S: Do not remember that one fuelsys has multiple tanks, or multiple fuel systems
- E: I will need to check - I have a vague memory that I saw something the potentially could support also fuels where volume typically is measured in kg.
- S: Relative levels sufficient
- S: Only sensor on one tank often, you do not see if you have filled the tank, you see it first when running.
- E: Assume there is some form of connection between them that levels out by time
- S: Need to check if we can/need to support individual sensors
- E: **I will check if I find some old notes**
- **Comments/Reqs from other welcome**

- DoorBus
- S: Does not matter if door bus is in std tree, as long as it is in vspec format
- E: Then possibly a good candidate for a profile, could be confusing to have both passenger car doors and bus doors in std catalog.

- S: Would like to get more feedback from ACEA members

**Please continue reviewing**

erikbosch:
I just thought of another aspect we better should discuss if we change wheel designation to Pos7/Pos8, and that is to review the signals in 

https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Chassis/Wheel.vspec

... and check if they need to be refactored to support twin wheels and similar. Today it contain a mix of various things:

- Brake fluid data - is there any need at all to have it wheel or "half axle" based. For motorbikes there is typically one container for front and one for back. For trucks with pneumatic brakes not relevant. 
- Padwear, Speed, Torque and similar - they are normally for "half axle"
- Tire data - they are for individual wheels

So is it maybe so that we shall split and keep one left/right branch for the "half axle" data

spec/Cabin/Tachograph.vspec
Armin-Keller: I think it means rest e.g., sleeping
This is a status which is declared at the digital tachograph by the driver (as all these states)

spec/Vehicle/Vehicle.vspec
Armin-Keller: ISO 8601 indicates the timezone with Z (or +00:00) for UTC 
Example:
UTC: 2026-09-02T10:00:00Z or 2026-09-02T10:00:00+00:00
UTC+2: 2026-09-02T12:00:00+02:00

spec/Vehicle/HMI.vspec
Armin-Keller: The index is based on the definition in FMS-Standard. For the moment we have 75 tell tales defined.
SAE refers to the definition in FMS-Standard.

spec/Powertrain/TractionBattery.vspec
Armin-Keller: 15 means "not available" Therefore it should be mentioned only status 0 and 15 should be deleted

spec/ADAS/ADAS.vspec
Armin-Keller: We have to note that this is the switch and not what the system does.
If the driver wnats to resume the switch is on "resume" until the stored speed is reached or any other state has been choosen by the driver

spec/ADAS/ADAS.vspec
Armin-Keller: main difference is that we have the state of the system in FMS - we have in FMS how the switch is set (means not an actuator).
Text in J1939:  Cruise control is switched on. It is not ensured that the engine is controlled by cruise control, as in the case of a large driver's demand the engine is controlled by the 
driver while cruise control is active (maximum selection of cruise control and driver's demand). The cruise control is set to 0 if a switch off condition occurs. 

IsEnabled means as Eric mentioned that it is possible to use the cruise control. If it is not enabled you cannot use the cruise control system by using the switch

spec/Vehicle/Vehicle.vspec
wiedmama: FMS states here 'Time base is OEM dependant' so from FMS side it is not specified. Regarding SAE definition of the PGN 65254 Time/Date Boradcast here are two SPNs 1601 LocalMinuteOffset and 1602 LocalHourOffset which are included in the same PGN that should contain offsets to UTC or else be 0xFF.

Armin-Keller: That means that it is the date/time which has been set in the vehicle and is shown to the driver. There might be a few sources for the actual date/time in a vehicle e.g., GPS, radio controlled time, tachograph, etc.. In most cases the time is local time. Therefore, I think it should be made clear what the source is and the result is according ISO8601. FMS does not state according ISO8601.
A truck driving from Spain to Portugal might not switch the time automatically.

wiedmama: FMS is not specific here. But according the underlying SAE definition it is quite clear:

- LocalHourOffset (SPN 1602) is set to 0xFF -> a unknown time standard is used. 
- LocalHourOffset is set to 0xFA -> time and date is local time and date
- LocalHourOffset is set to 0xF9 -> time and date is set as UTC and the local offset is either unknown or not supported
- LocalHourOffset between 0x66 and 0x94 (incl) -> UTC is used and local time offset is given by LocalHourOffset and LocalMinuteOffset (SPN 1601)
- else -> unknown time and date standard (could be either UTC or local time) so basically the same as 0xFF

I see no easy way how to show this in VSS  without additional information like timebase (local/utc/unkown) and time offset information (if available from source).

PS: as the signal is mentioned in the bus section, I assume that the original intent was to have a time which is shown to the passengers.

spec/Powertrain/FuelSystem.vspec
erikbosch: I found an old draft I did on adding support for mass based fuel system signals.

https://github.com/boschglobal/vehicle_signal_specification/commit/45d3650a6a032e9e963e2a4d07bd8141c4b4a768

The idea there was to separate current fuel system signals into two levels - fuelsystems and tanks. A vehicle may have multiple fuel systems. A fuelsystem has separate metrics for remaining range, fuel economy and similar. Taking DUAL_FUEL_DIESEL_CNG from the proposed signal above I assume that it does not make sense to have a shared "instant consumption" signal, especially as instant consumption likely is measure in liters for diesel and kg for CNG. So that vehicle would as I see it have two separate fuel systems.

But then a fuelsystem may have multiple tanks which potentially have different signals for capacity/level, in case they are measured separately. If they are physically connected and the vehicle computer system does not know that they are two tanks then they could better be represented as a single tank.

But I do not know what combinations that are important to support for FMS/ACEA. Taking a vehicle supporting DUAL_FUEL_DIESEL_CNG - how is for instance instant consumption supposed to be measured/reported for that one. One signal? Two signals? Unit? If we for instance see no short/mid-term need to support vehicles that use mass based fuel, then we can wait with supporting that.

```
#include VolumeFuelSystem.vspec FuelSystem
# Tank integrated by default, not added to a separate branch
#include FuelSystem/VolumeTank.vspec FuelSystem
```

spec/Chassis/Chassis.vspec
erikbosch: Based on meeting discussion I extended my draft at https://github.com/COVESA/vehicle_signal_specification/pull/940 on a possible wheel/axle refactoring. I think it would be good to treat that discussion/change separately to give more visibility and as it would be backward incompatible change to std catalog

PR #934, opened 2026-08-05
6.1 prepare - CI update

erikbosch:


erikbosch:
With this we update Python version similar to in vss-tool. Also adding the compose command

@jdacoello - if we do like in this PR and uses same upload approach for releases like in https://github.com/COVESA/vehicle_signal_specification/releases/tag/v6.0 for some other artifacts there would be a `vss_compose.tar.gz` file in the release, with contents as below. Please let me know if you think this is a good way to handle it, we could alternatively upload files individually, but then we maybe want to remove the "source" quantities/unit files on top and instead just use the "compose" variant (which is sorted alphabetically)

image

.github/workflows/buildcheck.yml
```diff
@@ -16,7 +16,7 @@ on:
 
 
 env:
-  PYTHON_VERSION: "3.10.6" # keep in sync with vss-tools
+  PYTHON_VERSION: "3.11.5" # keep in sync with vss-tools
```
sschleemilch: i think we should actually use the current stable Python instead (3.14)

erikbosch: I have no strong opinion here, As we test the "full range" in vss-tools (right now 3.11 to 3.14) using latest released version here should not be a problem. I can try that out.

erikbosch: Updated it

PR #939, opened 2026-08-21
Adding PR template

erikbosch:
This a proposal on how to handle the new COVESA contributor agreement. With a PR template you will see something like this when creating a PR

image

And in rendered view this corresponds to:

image

There are quite a lot of details where I do not know if we need to clarify corner cases. Like I contribute on behalf of a German company but I am located and employed in Sweden. I also assume that we do not want to request e-mail address in cleartext as that might lead to spams.

And if you do not like it this way we need to come up with an alternative way. Like if we do not want to have this info public we may request everyone to send a mail to Paul with the data. To be discussed in coming meetings

- And my assumption is that git commit signoff no longer is required, or? Well we still mention it at https://covesa.global/about-covesa/#contribute-code

```
For legal compliance, every contribution must carry the following information:

a) sign-off line with your real name and email (i.e., Signed-off-by: Firstname Lastname you@example.com)
b) entity name if contributing on its behalf (otherwise, it will be considered an individual contribution)
c) country from which contribution originates.
```

So do we want this info on every commit or every PR? every commit is safer if we cherry pick commits, but also more complex to use. So is PR sufficient?

## Contributor Information

* Contributor name(s): Erik Jaegervall
* Contributing Organization: Robert Bosch GmbH
* E-mail: erik dot jaegervall at bosch dot com
* Country: Sweden

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,47 @@
+<!--
+Contributions to COVESA must follow the COVESA Antitrust and Open Statements
+available on the link below and for convenience copied below
+
+https://covesa.atlassian.net/wiki/x/FhxUAg
+
+
+Contributing to COVESA Activities
+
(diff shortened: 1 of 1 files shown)


erikbosch:
MoM:

- Presented at meeting
- Jeremy may not be a problem for companies as they have good spam filters, but could be a problem for individual contributors 
- E: Should we allow an alternative, like send a mail to Paul with your github handle included
- E: Do we require it for very commit/PR, also for known contributors
- Input from @paulboyes wanted

erikbosch:
MoM:

- P: Do not have details on how COVESA wants to do it

PR #942, opened 2026-09-02
NOT FOR MERGE -6.1 preparation

erikbosch:
Just to follow up on changes

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -54,7 +54,6 @@ jobs:
           path: |
             vss.csv
             vss.fidl
-            vss.graphql
             vss.idl
             vss.json
             vss.jsonschema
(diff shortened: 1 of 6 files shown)


.github/workflows/buildcheck.yml
```diff
@@ -63,6 +62,7 @@ jobs:
             vss_compose.tar.gz
             vss_samm.tar.gz
             vss_apigear.tar.gz
+            vss_graphql.tar.gz
```
erikbosch, line 65: @jdacoello  - FYI - I had to do this change as s2dm now generates multiple files. Please check if it looks ok. I am also a bit uncertain if we want to call the artifact s2dm or graphql

.github/workflows/buildcheck.yml
```diff
@@ -63,6 +62,7 @@ jobs:
             vss_compose.tar.gz
```
erikbosch, line 62: @jdacoello - please check that this artifact has expected name and content

erikbosch:
Will be replaced by other

PR #943, opened 2026-09-07
Multi fuelsystem refactoring

erikbosch:
This is a draft for discussion. The FMS/ACEA wants to support multiple fuel tanks. Today VSS supports only one fuel tank with liquid fuel. This PR brings up an idea of supporting both multiple fuel systems and multiple tanks.

In short - If a vehicle produce different "remaining range" metrics for different (combustion engine) fuel types, then it has multiple fuel systems. If it just has multiple fuel tanks, but only one set of fuel economy metrics, those tanks belong to the same fuel system.

Things to discuss:

- Does the differentiation in this PR make sense?
- Do we need to support all those combos? No reason to overdo anything? 


docs-gen/content/catalog/_index.md
```diff
@@ -11,4 +11,5 @@ Each signal in the VSS catalog has a `description` field and they may have a `co
 
 ## VSS Catalog Concepts
 
-* [Vehicle Motion Management](/vehicle_signal_specification/catalog/vehicle_motion_management/) Information on the VSS concept for motion management.
+* [Fuel System Representation](/vehicle_signal_specification/catalog/vehicle_motion_management) Information on the VSS concept for motion management.
```
SebastianSchildt, line 14: Link and test not updated. Link title says "Fuel" but link and text still say "motion management"

erikbosch:
MoM:

- Erik presented the PR
- Please review
- Sam: For fuel codes fixed set

Issue #46, opened 2017-08-15
Transmission Drive Type  enum:  abbreviation not really clear “fwd”, “rwd”


Geraldbosch:

gunnarx:
File, and line, please

gunnarx:
I assume it's fwd = forward wheel drive, rwd = rear whell drive, awd = all wheel drive.
Should there be a comment to clarify?  Please suggest a change through PR.

klotzbenjamin:
In https://github.com/GENIVI/vehicle_signal_specification/blob/master/spec/Drivetrain/Transmission.vspec, it has been corected already:

`- DriveType:`
`  datatype: string`
`  type: attribute`
`  enum: [ "unknown", "forward wheel drive", "rear wheel drive", "all wheel drive" ]`
`  value: "unknown"`
`  description: Drive type.`

@danielwilms we can close this issue.

danielwilms:
Hi @klotzbenjamin, great. There's one open question for me: Do we have a common way to define enums? The spaces look weird.

klotzbenjamin:
> Do we have a common way to define enums? The spaces look weird.

I do not see any rule for defining enums, in this case the only rule is that the elements from the enum are of `datatype: string`.

danielwilms:
OK, I'll close this one then and open a new one for uniform enum definition.

PR #946, opened 2026-09-21
Add FOUR_WHEEL_DRIVE to Transmission.DriveType

samsullivandelgobbo:
`Transmission.DriveType` offers `ALL_WHEEL_DRIVE` and no four-wheel-drive value. Registration and type data record the two separately, so where a source says four-wheel drive there is nothing in the allowed list to carry it.

One value added. Existing values unchanged.

## What we measured

Cardog decodes VINs into vehicle identity records from Canadian and US registration, dealer and insurer feeds. 1,870,496 distinct decoded units, counted 2026-09-09. Counts are exact; shares are of that total.

| Drive type as decoded | Units | Share |
|---|---|---|
| Four-wheel drive (4x4) | 898,701 | 48.04% |
| All-wheel drive | 391,045 | 20.90% |
| 4x2 (driven axles by count, not position) | 204,358 | 10.92% |
| Front-wheel drive | 65,533 | 3.50% |
| Rear-wheel drive | 9,123 | 0.49% |
| Not decoded | 298,678 | 15.97% |

Those rows cover 99.82% of the corpus; the remaining 3,058 units carry other axle forms.

In the model catalogue, where the mix is passenger-weighted rather than fleet-weighted, the shape is different but neither value is small: FWD 134,848, RWD 72,983, AWD 71,505, 4WD 19,498, unknown 13,832, of 312,666 variants.

## Prior art

- The allowed list has held three values plus `UNKNOWN` since the first vspec file in 2016. Renamed in 2018 (#46, #64), `enum` to `allowed` in 2021, upper-cased in 2022 (#404). No issue or PR has raised four-wheel drive.
- schema.org `driveWheelConfiguration`, the vocabulary several `VehicleIdentification` names came from, carries four members including `FourWheelDrive`.
- vPIC records 4WD/4x4 and AWD as separate values.

## On the comment

The comment states that assignment follows the manufacturer's or registration authority's designation, and gives no mechanical rule. Full-time four-wheel drive and automatically engaged all-wheel drive both exist, so a mechanism-based definition would contradict the registration data it is meant to describe.

## Not in scope

- Axle designations (4x2, 6x4, 6x6). 10.92% of the units above carry that form and `DriveType` cannot express it; that belongs with the commercial and fleet work.
- Renaming `FORWARD_WHEEL_DRIVE` to `FRONT_WHEEL_DRIVE`. Breaking, so 7.0.

## Contributor information

- Contributor name: Sam Sullivan-Del Gobbo
- Contributing organization: Cardog Inc.
- E-mail: sam@cardog.io
- Country where the change was written: Canada


erikbosch:
I know that in US that it is common to differentiate between AWS and 4WD. Have not heard that much about similar "official" categorization in Europe. Does anyone know if similar differentiation officially exist in Europe?

(In Sweden you can see in public registers that a car can drive and both front and rear axle, but no information on technology used)

erikbosch:
MoM presented at meeting

